PR #1, opened 2017-12-31
Added functions for managing wireguard tunnels

mrubacky:
setup_wg_if: takes an ip address to assign to the interface, along with a peer
address and returns the interface name that was created

open_tunnel: takes a remote endpoint, a local private key file path, a
remote public key, and the wirguard interface and configures the
wireguard tunnel.

delete_tunnel: takes an interface name and deletes it

create_wg_key: saves a wireguard private key to a given path.

get_wg_pubkey: takes a wireguard private key path and returns the
corresponding public key


jtremback:
can you put those comments above the functions with a triple comment syntax `///`. That way they will be picked up by the autogenerated docs

PR #2, opened 2018-01-06
Openwrt integration

drozdziak1:
This pull request makes sure that `althea_rs` is ready for batch building in `althea-firmware` and cuts loose the dependency from OpenSSL (`hwaddr` -> `eui48` in `althea_kernel_interface`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "althea_rs"
+version = "0.1.0"
+authors = ["Stan Drozd <drozdziak1@gmail.com>"]
+
+[dependencies]
+guac-light-client = { path = "./guac-light-client" }
+rita = { path = "./rita" }
+
+[workspace]
--- a/althea_kernel_interface/src/lib.rs
+++ b/althea_kernel_interface/src/lib.rs
@@ -4,21 +4,22 @@ extern crate derive_error;
 #[macro_use]
 extern crate log;
 
-use std::str;
-extern crate hwaddr;
+extern crate eui48;
 extern crate regex;
 extern crate itertools;
 
-use itertools::join;
-use std::net::{IpAddr, SocketAddr, SocketAddrV6, Ipv6Addr};
-use hwaddr::HwAddr;
-use std::str::FromStr;
-use regex::Regex;
-use std::process::{Command, Output, ExitStatus, Stdio};
-use std::os::unix::process::ExitStatusExt;
-use std::path::Path;
 use std::fs::{File, remove_file};
 use std::io::Write;
+use std::net::{IpAddr, SocketAddr, SocketAddrV6, Ipv6Addr};
+use std::os::unix::process::ExitStatusExt;
+use std::path::Path;
+use std::process::{Command, Output, ExitStatus, Stdio};
+use std::str::FromStr;
+use std::str;
+
+use eui48::MacAddress;
+use itertools::join;
+use regex::Regex;
 
 #[derive(Debug, Error)]
 pub enum Error {
@@ -51,7 +52,7 @@ impl KernelInterface {
         (self.run_command)(args, program)
     }
 
-    fn get_neighbors_linux(&mut self) -> Result<Vec<(HwAddr, IpAddr)>, Error> {
+    fn get_neighbors_linux(&mut self) -> Result<Vec<(MacAddress, IpAddr)>, Error> {
         let output = self.run_command("ip", &["neighbor"])?;
         trace!("Got {:?} from `ip neighbor`", output);
 
@@ -61,7 +62,7 @@ impl KernelInterface {
             trace!("Regex captured {:?}", caps);
 
             vec.push((
-                caps.get(2).unwrap().as_str().parse::<HwAddr>()?,
+                MacAddress::parse_str(caps.get(2).unwrap().as_str()).unwrap(), // Ugly and inconsiderate, ditch ASAP
                 IpAddr::from_str(&caps[1])?,
             ));
         }
@@ -71,7 +72,7 @@ impl KernelInterface {
 
     fn start_flow_counter_linux(
         &mut self,
-        source_neighbor: HwAddr,
+        source_neighbor: MacAddress,
         destination: IpAddr,
     ) -> Result<(), Error> {
         self.delete_flow_counter_linux(source_neighbor, destination)?;
@@ -81,7 +82,7 @@ impl KernelInterface {
                 "-A",
                 "INPUT",
                 "-s",
-                &format!("{}", source_neighbor),
+                &format!("{}", source_neighbor.to_hex_string()),
                 "-p",
                 "IPV6",
                 "--ip6-dst",
@@ -142,14 +143,14 @@ impl KernelInterface {
 
     fn delete_flow_counter_linux(
         &mut self,
-        source_neighbor: HwAddr,
+        source_neighbor: MacAddress,
         destination: IpAddr,
     ) -> Result<(), Error> {
         self.delete_ebtables_rule(&[
             "-D",
             "INPUT",
             "-s",
-            &format!("{}", source_neighbor),
+            &format!("{}", source_neighbor.to_hex_string()),
             "-p",
             "IPV6",
             "--ip6-dst",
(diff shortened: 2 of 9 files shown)


src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(test)]
```
drozdziak1: I'm actually a little proud of that submodule purge above, it's more difficult than you think :D 


This `lib.rs` file is probably pointless, but I think we'd set up a crate here anyway at some point.

Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "althea_rs"
+version = "0.1.0"
+authors = ["Stan Drozd <drozdziak1@gmail.com>"]
```
drozdziak1: Add yourselves and whomever you deem worthy here.

althea_kernel_interface/src/lib.rs
```diff
@@ -81,7 +82,7 @@ impl KernelInterface {
                 "-A",
                 "INPUT",
                 "-s",
-                &format!("{}", source_neighbor),
+                &format!("{}", source_neighbor.to_hex_string()),
```
jtremback: I'm surprised that `format!` doesn't do this automatically with the `to_string` trait or whatever.

althea_kernel_interface/src/lib.rs
```diff
@@ -175,13 +176,15 @@ impl KernelInterface {
...
+    fn read_flow_counters_linux(&mut self) -> Result<Vec<(MacAddress, IpAddr, u64)>, Error> {
         let output = self.run_command("ebtables", &["-L", "INPUT", "--Lc"])?;
         let mut vec = Vec::new();
         let re = Regex::new(r"-s (.*) --ip6-dst (.*)/.* bcnt = (.*)").unwrap();
         for caps in re.captures_iter(&String::from_utf8(output.stdout)?) {
             vec.push((
-                caps[1].parse::<HwAddr>()?,
+                    MacAddress::parse_str(&caps[1]).unwrap_or_else(|e| {
+                        panic!("{:?}", e);
+                    }), // Ugly and inconsiderate, remove ASAP
```
jtremback: Just wondering, what's the reason here?

althea_kernel_interface/src/lib.rs
drozdziak1: @jtremback it's the MAC package's gimmick, default format is `aa-bb-cc-dd-ee-ff` when `aa:bb:cc:dd:ee:ff` is used more often and that's what `to_hex_string` formats it to

althea_kernel_interface/src/lib.rs
drozdziak1: That MAC package doesn't implement a trait for converting its own error type to standard `Err()`, this seemed quicker to do

drozdziak1: I'd suggest implementing that trait in the future. Libraries are no place for panics, it's the user who should decide how to handle the parsing error.

src/lib.rs
drozdziak1: (it makes no sense in the main PR view, click "View changes")

althea_kernel_interface/src/lib.rs
```diff
@@ -455,9 +458,9 @@ fe80::433:25ff:fe8c:e1ea dev eth0 lladdr 1a:32:06:78:05:0a STALE
 b"Bridge table: filter
 
 Bridge chain: INPUT, entries: 3, policy: ACCEPT
--p IPv6 -s 0:0:0:aa:0:2 --ip6-dst 2001::1/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff -j ACCEPT , pcnt = 1199 -- bcnt = 124696
--p IPv6 -s 0:0:0:aa:0:0 --ip6-dst 2001::3/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff -j ACCEPT , pcnt = 1187 -- bcnt = 123448
--p IPv6 -s 0:0:0:aa:0:0 --ip6-dst 2001::3/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff -j ACCEPT , pcnt = 0 -- bcnt = 0"
+-p IPv6 -s 00:00:00:aa:00:02 --ip6-dst 2001::1/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff -j ACCEPT , pcnt = 1199 -- bcnt = 124696
```
jtremback: Why did this change? I found that the Linux I was using (Ubuntu) output the MAC address in the `0:0:0:aa:0:2` format. Is OpenWRT different?

althea_kernel_interface/src/lib.rs
```diff
@@ -545,7 +548,7 @@ Bridge chain: INPUT, entries: 3, policy: ACCEPT
         };
 
         match ki.delete_flow_counter_linux(
-            "0:0:0:aa:0:2".parse::<HwAddr>().unwrap(),
+            MacAddress::parse_str("00:00:00:aa:00:02").unwrap(),
```
jtremback: Hmm, surprised that `.parse` doesn't work with this library, I think it just needs the `ToString` trait IIRC. No big deal I guess...

src/lib.rs
jtremback: Yea... not sure how that submodule happened. Probably I did it by accident somehow.

althea_kernel_interface/src/lib.rs
drozdziak1: The MAC lib couldn't understand this, I'll check tomorrow, as an alternative I could just write a regex for this.

althea_kernel_interface/src/lib.rs
drozdziak1: Sure, I'm all in for alternatives. After looking at your other comments this lib doesn't seem as good as before

jtremback:
I left some questions about some stuff, but nothing blocks me merging, so I'll merge it and you can respond to those questions.

jtremback:
Oh I see you already did lol

althea_kernel_interface/src/lib.rs
jtremback: Either we add more robust parsing to this MAC library, or we deal with the dep situation in the HwAddress lib. Not sure.

althea_kernel_interface/src/lib.rs
jtremback: The only real issue I have with this lib is the less robust parsing discussed in the comment above

althea_kernel_interface/src/lib.rs
drozdziak1: I'd patch the new lib, its parsing code is pretty straightforward.     

```rust
/// Parses a String representation from any format supported
    pub fn parse_str(s: &str) -> Result<MacAddress, ParseError> {
        let mut offset = 0; // Offset into the u8 Eui48 vector
        let mut hn: bool = false; // Have we seen the high nibble yet?
        let mut eui: Eui48 = [0; EUI48LEN];

        match s.len() {
            14 | 17 => {}  // The formats are all 12 characters with 2 or 5 delims
            _ => return Err(ParseError::InvalidLength(s.len())),
        }

        for (idx, c) in s.chars().enumerate() {
            if offset >= EUI48LEN {
                // We shouln't still be parsing
                return Err(ParseError::InvalidLength(s.len()));
            }

            match c {
                '0'...'9' | 'a'...'f' | 'A'...'F' => {
                    match hn {
                        false => {
                            // We will match '0' and run this even if the format is 0x
                            hn = true; // Parsed the high nibble
                            eui[offset] = (c.to_digit(16).unwrap() as u8) << 4;
                        }
                        true => {
                            hn = false; // Parsed the low nibble
                            eui[offset] += c.to_digit(16).unwrap() as u8;
                            offset += 1;
                        }
                    }
                }
                '-' | ':' | '.' => {}
                'x' | 'X' => {
                    match idx {
                        1 => {
                            // If idx = 1, we are possibly parsing 0x1234567890ab format
                            // Reset the offset to zero to ignore the first two characters
                            offset = 0;
                            hn = false;
                        }
                        _ => return Err(ParseError::InvalidCharacter(c, idx)),
                    }
                }
                _ => return Err(ParseError::InvalidCharacter(c, idx)),
            }
        }

        if offset == EUI48LEN {
            // A correctly parsed value is exactly 6 u8s
            Ok(MacAddress::new(eui))
        } else {
            Err(ParseError::InvalidLength(s.len())) // Something slipped through
        }
}
```

PR #3, opened 2018-01-06
Rita alpha groundwork

jtremback:
This gets Rita into the basic state that it needs to be. There is still a lot of work to be done cleaning it up, getting rid of unwraps, and actually testing it.


rita/Cargo.toml
```diff
@@ -6,11 +6,19 @@ authors = ["Jehan <jehan.tremback@gmail.com>"]
...
+num256 = { path = "../num256" }
 ip_network = "0.1"
 log = "0.3"
 docopt = "0.8.1"
 simple_logger = "0.4.0"
+rouille = "1.0.3"
+serde = "1.0.24"
+serde_derive = "1.0.24"
+serde_json = "1.0.8"
+reqwest = "0.8.2"
```
drozdziak1: This is going to cause that OpenSSL stench to return

drozdziak1: Do we need HTTP this bad?

rita/src/main.rs
```diff
@@ -21,25 +18,25 @@ use babel_monitor::Babel;
...
 
+extern crate tunnel_manager;
+use tunnel_manager::TunnelManager;
+
 extern crate num256;
 use num256::Int256;
 
-extern crate rouille;
-use rouille::{Request, Response};
+#[macro_use] extern crate rouille;
```
drozdziak1: People usually write `#[macro_use]` above `extern crate`, does this build normally?

jtremback: As far as I can tell, yes

rita/Cargo.toml
jtremback: It's an easy to use protocol that is understood by everybody already, which is why I'm using it. I agree that the libraries that implement it tend to be heavy.  We could in principle use any rpc technique.

I can take a look at how hard it is to submit a patch to our http stuff to put https behind a cfg flag.

jtremback: Sorry, I missed most of this stuff in the chat. What are the specific issues around openssl? This will help inform my pull request.

jtremback: Actually, thinking about it more. We could avoid https on node-to-node comms, but as soon as we're talking to servers on the internet we will either need to use https or reinvent the wheel.

drozdziak1: Yeah, we're probably gonna hit that brick wall anyway when we'll need to reach APIs like an automated version of the node locator map API

PR #4, opened 2018-01-07
get rid of unwrap in lib code

jtremback:


jkilpatr:
Looks good to me. The ? operator might be strange to rust newbies but they can just look it up.

rita/tunnel_manager/src/lib.rs
```diff
@@ -1,17 +1,27 @@
...
+extern crate derive_error;
+
 use std::net::IpAddr;
 
 extern crate althea_types;
 use althea_types::EthAddress;
 
 extern crate debt_keeper;
-use debt_keeper::{Identity, Key};
+use debt_keeper::Identity;
```
drozdziak1: Import grouping

PR #5, opened 2018-01-09
Remove [[workspace]] from Rita's Cargo.toml:

drozdziak1:
Cargo does not permit multiple workspace roots ([[workspace]]
declarations) which would prevent the package from building.

This commit removes the redundant workspace root.


jkilpatr:
yeha I noticed that as well. I'm not sure why it was working for Jehan. I agree on principle but it might be part of a larger org problem we need to address.

PR #10, opened 2018-01-16
Metering, Balances and Centralized Bounty Hunter

kingoflolz:


althea_kernel_interface/src/lib.rs
```diff
@@ -46,6 +47,9 @@ impl KernelInterface {
     fn run_command(&mut self, program: &str, args: &[&str]) -> Result<Output, Error> {
         let output = Command::new(program).args(args).output()?;
         trace!("Command {} {:?} returned: {:?}", program, args, output);
+        if !output.status.success() {
+            trace!("An error was returned");
```
jtremback: Why?

kingoflolz: I was grepping to see what commands failed, and there were too many different error codes that can occur lol

PR #11, opened 2018-01-17
Add travis CI builds

jkilpatr:
DebtKeeper doens't seem to build right, so we should probably clean up our crates and then get this merged.

Should we test only nightly or all three release streams of rust?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: rust
+rust:
+  - stable
+  - beta
+  - nightly
+script:
+  - cargo build --verbose --all
+  - cargo test --verbose --all


drozdziak1:
IMHO knowing more won't ever hurt us, and I'd build all of them. IIUC in travis there is an option to mark the desired kinds of failures as non-critical, maybe we could use that on some of the Rust toolchain flavors?

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: rust
+rust:
+  - stable
+  - beta
+  - nightly
+script:
+  - cargo build --verbose --all
+  - cargo test --verbose --all
```
drozdziak1: Huge +1 for this, failing test builds are easy to overlook when doing that last build before pushing and opening a PR.

jkilpatr: I actually really like travis for how simple it is, but it's limited for larger builds like the firmware. Which means we may need a firmware build server with a jenkins instance :(

PR #19, opened 2018-01-20
Optimizations to Reduce binary size

jkilpatr:
Three major changes have been performed here.

We enable link time optimization, so that all
the libraries we compile are stipped down to
exactly the functions we use.

We enable the 's' optimization flag, which
compiles with size optimization instead of
performance ones.

And we move from the Jemalloc memory allocator
to dynamically linking against the system allocator.

Total reduction in binary size is a little more
than half.


kingoflolz:
I don't think that we need to strip down the binary size of bounty hunter, as it should not be running on anything too resource constrained. I think we should also put the system allocator behind a feature flag (maybe have a feature flag just for reduced size, so future optimisations can be put behind that as well), so it can still build on stable, even if the binaries are larger.

jkilpatr:
@kingoflolz sounds good to me, where can I find the syntax for that?

kingoflolz:
See https://doc.rust-lang.org/book/first-edition/conditional-compilation.html

I think I did something similar in Rita for clippy, so you can check that out as well.

PR #21, opened 2018-01-21
Don't ignore Cargo.lock

drozdziak1:
# What this change means
The `Cargo.lock` file is usually automatically updated in tandem
with dependency changes in a project. The dependencies listed inside are
always in precise versions which fall into the requirements of the
project's crates. These precise versions are supposed to make Cargo
always download the same version of a crate even if its dependency
version in the `[dependencies]` section of a `Cargo.toml` is set to
`"*"`.

# Rationale
The Cargo manual [suggests](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) including `Cargo.lock`s in version control of binary crates as a good practice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,5 @@
 *.bak
 /target/
 **/*.rs.bk
-Cargo.lock
 .idea
 integration_test/test.db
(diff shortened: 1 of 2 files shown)


jkilpatr:
+1 a good idea so that we don't get surprised by non building crates, we will have to do cargo update every so often and test the results.

PR #22, opened 2018-01-21
Seperate build and test into different jobs

jkilpatr:
This is more readable in the output and provides more information at a glance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: rust
 cache: cargo
+env:
+  - TYPE=build
+  - TYPE=test
 rust:
   - stable
   - beta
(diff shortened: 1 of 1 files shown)


drozdziak1:
Okay, so it seems that only nightly works for now?

jkilpatr:
we're using some nightly only features

jkilpatr:
not sure if I should cut out the other two channels but we do want to know when the features we need finally land in beta and stable

kingoflolz:
Just put a commit in which makes rita build on all 3 channels, so I think we should keep building on all of them

jkilpatr:
Nice I'll land this then.

PR #29, opened 2018-01-30
Added trait PaddedHex

janus:
Implemented PaddedHex trait for Int256, method to_padded_hex which returns string
Change fmt::Display trait implemenation to return base 16
These modifications are to make way for ethereum


num256/src/lib.rs
```diff
@@ -250,6 +248,29 @@ impl CheckedMul for Int256 {
   }
 }
 
+impl fmt::Display for Int256 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", &self.to_str_radix(16))
+        //changed to base 16 for ethereum
```
kingoflolz: Is it possible to make the base16 write function a different one from fmt? Currently, fmt is used a lot for debugging, so it would be good to preserve base 10 for it.

jkilpatr:
@janus the test databases shouldn't go into the repository, because of the way git history works it's very inefficient to merge any sort of binary file, please remove them and squash these two commits.

janus:
Thanks ...I have made the changes.

PR #31, opened 2018-01-31
Introduced PaddedHex trait and removed typo

janus:
This is mainly for althea num256 to work with EVM
The method of the trait is to_padded_hex()
Removed error caused by self::to_str_radix(16)  instead of self::to_str_radix(16)


jkilpatr:
looks like good work, but rustfmt really makes the commit hard to read. Reapply your changes on a different branch and then run rustfmt as a separate commit. that will then be a separate pull request (since it's on a new dev branch)

PR #34, opened 2018-02-07
Improve integration tests, fix bugs in rita

kingoflolz:
(this PR is a superset of #30)


rita/payment_controller/src/lib.rs
```diff
@@ -87,16 +90,19 @@ impl PaymentController {
...
         }
     }
 
     fn update_bounty(&self, update: BountyUpdate) -> Result<(), Error> {
+        trace!("Sending bounty hunter update: {:?}", update);
         let bounty_url = if cfg!(not(test)) {
             format!("http://[{}]:8888/update", "2001::3".parse::<Ipv6Addr>().unwrap())
         } else {
-            String::from("http://127.0.0.1:1234") //TODO: This is mockito::SERVER_URL, but don't want to include the crate in a non-test build just for that string
+            String::from("http://127.0.0.1:1234/update") //TODO: This is mockito::SERVER_URL, but don't want to include the crate in a non-test build just for that string
```
drozdziak1: This todo most likely should've become a GitHub issue

rita/payment_controller/src/lib.rs
```diff
@@ -299,7 +322,7 @@ mod tests {
     #[test]
     fn test_multi_make_payments() {
         // mock neighbour
```
drozdziak1: It's `neighbor` or `neighbour`, let's use only one and be consistent or stupid bugs will befall us. Personally I'd go with `neighbor`

rita/src/main.rs
```diff
@@ -77,37 +77,42 @@ fn main() {
 
     trace!("Starting with Identity: {:?}", my_ident);
 
-    let (tx, rx) = mpsc::channel();
 
-    let tx1 = mpsc::Sender::clone(&tx);
+    let (debt_keeper_input, debt_keeper_output) = DebtKeeper::start(Int256::from(1), Int256::from(-100000000000000i64), Int256::from(100), 2u32);
```
drozdziak1: What makes `-100000000000000i64` special? Why not just use `Int256`'s minimum value by default?

rita/src/main.rs
```diff
@@ -77,37 +77,42 @@ fn main() {
...
-                let adjustment = DebtAdjustment {ident, amount};
-                trace!("Sent debt adjustment {:?}", &adjustment);
-                tx1.send(adjustment).unwrap();
+            for (from, amount) in debts {
+                let update = DebtKeeperMsg::TrafficUpdate { from, amount };
+                let adjustment = DebtKeeperMsg::SendUpdate { from };
+                debt_keeper_input1.send(update).unwrap();
+                debt_keeper_input1.send(adjustment).unwrap();
             }
+            payment_controller_input1 .send(PaymentControllerMsg::Update);
```
drozdziak1: A likely unwanted space after the `1`

rita/src/main.rs
```diff
@@ -77,37 +77,42 @@ fn main() {
...
 
-    let tx1 = mpsc::Sender::clone(&tx);
+    let (debt_keeper_input, debt_keeper_output) = DebtKeeper::start(Int256::from(1), Int256::from(-100000000000000i64), Int256::from(100), 2u32);
+
+    let payment_controller_input = PaymentController::start(
+        &my_ident,
+        Arc::new(Mutex::new(debt_keeper_input.clone()))
+    );
+
+    let payment_controller_input1 = payment_controller_input.clone();
```
drozdziak1: Why is this variable numbered? If it's got a distinct purpose why doesn't the name express it?

rita/src/main.rs
```diff
@@ -77,37 +77,42 @@ fn main() {
...
+    let debt_keeper_input1 = mpsc::Sender::clone(&debt_keeper_input);
     thread::spawn(move || {
         let mut ki = KernelInterface {};
         let mut tm = TunnelManager::new();
-        let mut babel = Babel::new(&"[::1]:8080".parse().unwrap()); //TODO: Do we really want [::1] and not [::0]?
+        let mut babel = Babel::new(&"[::1]:8080".parse().unwrap());
```
drozdziak1: Why does Babel start on 8080? It's a well known port already widely used for user-controlled HTTP servers.

drozdziak1: I mean, what if someone runs an exit and would like to also put up a status webpage on 80 and some unrelated program's admin panel on 8080?

rita/src/main.rs
```diff
@@ -77,37 +77,42 @@ fn main() {
...
+            payment_controller_input1 .send(PaymentControllerMsg::Update);
         };
     });
 
-    let m_tx = Arc::new(Mutex::new(tx.clone()));
-
-    let pc = PaymentController::start(&my_ident, m_tx);
-
-    let pc1 = pc.clone();
+    let payment_controller_input2 = payment_controller_input.clone();
```
drozdziak1: Another numbered thing

rita/src/main.rs
```diff
@@ -122,28 +127,20 @@ fn main() {
         });
     });
 
-    let mut dk = DebtKeeper::new(Int256::from(5), Int256::from(-10));
-
-    for debt_adjustment in rx {
-        match dk.apply_debt(debt_adjustment.ident, debt_adjustment.amount) {
+    for msg in debt_keeper_output {
```
drozdziak1: Why does this loop need to constantly roll, wouldn't it be better to block until something actually happens? That way this thread could just go to sleep on the socket within the debt keeper

num256/src/lib.rs
```diff
@@ -250,6 +290,25 @@ impl CheckedMul for Int256 {
   }
 }
 
+impl Div for Int256 {
+  type Output = Int256;
+  fn div(self, v: Int256) -> Int256 {
+    let num = self.0 / v.0;
```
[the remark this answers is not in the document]
drozdziak1: https://github.com/rust-lang/rfcs/issues/1368
A quote from the issue above:
fwiw, I see no mention here of "thread-directed" signals. IMHO, it's reasonable to not support them as a first pass (they're a totally different thing than process-directed signals), but worth at least being explicit about what's in scope.

As an example of what these are: let's say I do a mmap of a file and try to read from it but another process has truncated the file so the page I'm reading is no longer valid. A SIGBUS is generated and directed to the thread doing the read. It can't be handled by some other thread using signalfd or the like.

Likewise, SIGFPE, SIGSEGV, SIGILL, SIGPIPE, and anything done with pthread_kill/pthread_sigqueue. The last case is sometimes used as a mechanism for CPU / wall clock profiling tools or thread cancellation.

rita/src/main.rs
jtremback: Pretty sure this does block, it is iterating over the channel output

rita/src/main.rs
jtremback: It's a clone of a channel input. I would be curious what naming scheme you'd prefer. We could name it after what it was intended to be used by, but that might get very long winded.

num256/src/lib.rs
```diff
@@ -58,6 +67,13 @@ impl fmt::Display for Uint256 {
     }
 }
 
+impl fmt::Debug for Uint256 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Uint256 {}", self.to_string())
```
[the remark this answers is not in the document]
jtremback: Could be good. I would personally prefer Uint256(3402342), but that's a matter of taste

rita/src/main.rs
jtremback: That is 100k micro ethers which might be around 10 cents or something. Using Int256's minimum value would result in a debt keeper which does not require counterparts to pay at all.

rita/src/main.rs
drozdziak1: Oh, okay, my poor knowledge of Rust strikes again

rita/src/main.rs
drozdziak1: Okay. IMHO we should definitely use an exponent or Rust's underscore notation here, e.g. `-1e14 as i64` or  `-100_000_000_000_000`

rita/src/main.rs
drozdziak1: I understand now, so there's two different components chatting with the payment controller? And shadowing the same variable name would get the borrow checke angry?

drozdziak1: (or simply overwriting the variable)

drozdziak1: @jtremback

integration-tests/rita.sh
```diff
@@ -1,181 +1,7 @@
...
-pass_string "destination: V6(2001::1), bytes: 7240" "rita-n2.log"
-pass_string "Calculated neighbor debt. price: 11, debt: 79640" "rita-n2.log"
-pass_string "Calculated neighbor debt. price: 15, debt: 108600" "rita-n2.log"
-pass_string "prefix: V6(Ipv6Network { network_address: 2001::1, netmask: 128 })" "rita-n2.log"
-pass_string "prefix: V6(Ipv6Network { network_address: 2001::3, netmask: 128 })" "rita-n2.log"
-
-echo "$0 PASS"
\ No newline at end of file
+./rita-deps.sh
+sudo python3 rita.py
```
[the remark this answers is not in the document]
kingoflolz: good point, I'll move it back

num256/src/lib.rs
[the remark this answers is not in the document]
kingoflolz: went with Uint256(3402342) as I think its a bit more "rusty"

num256/src/lib.rs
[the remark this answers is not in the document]
kingoflolz: This is a bigint, so I think it will panic instead of raising a hardware exception, but I have added div by zero checker

rita/payment_controller/src/lib.rs
kingoflolz: Changed all usages to neighbor, although babel uses neighbour

rita/src/main.rs
kingoflolz: I changed it so there is a master, which we clone, and just pass in variables which we just overwrite each time

rita/src/main.rs
kingoflolz: To be fair, I don't really know. That's what @jtremback 's original test script was set to so I just continued using it.

rita/src/main.rs
kingoflolz: This is actually a mistake, it should be something closer in magnitude to 1% the amount of wei sent in the test. I will be changing this to 1 billion (the tests usually require payments ~50Gwei).

PR #35, opened 2018-02-07
Optimizations to Reduce binary size

jkilpatr:
Three major changes have been performed here.

We enable link time optimization, so that all
the libraries we compile are stipped down to
exactly the functions we use.

We enable the 's' optimization flag, which
compiles with size optimization instead of
performance ones.

And we move from the Jemalloc memory allocator
to dynamically linking against the system allocator.

Total reduction in binary size is a little more
than half.


jkilpatr:
@drozdziak1 @kingoflolz so this breaks us on the stable channel because we try and remove jemalloc in place of the system allocator to save some binary size. Decision time, which is more important to us, binary size or running on stable? 

That being said jemalloc removal doesn't save us too much right now, a few hundred kb.

drozdziak1:
@jkilpatr Can't we get away with installing jemalloc on OpenWrt as a dependency and linking dynamically? I've got a faint memory of it being there and already checked as a dep for something else

kingoflolz:
@jkilpatr What did you think of using conditional compilation to make the system allocator conditional? Then we should be able to keep on building on stable.

Syntax would go something like this on the top of the main.rs of rita:
```
#![cfg_attr(feature="system_alloc", feature(alloc_system, global_allocator, allocator_api))]
#[cfg(feature = "system_alloc")]
extern crate alloc_system;

#[cfg(feature = "system_alloc")]
use alloc_system::System;

#[cfg(feature = "system_alloc")]
#[global_allocator]
static A: System = System;
```

And system allocator can be turned on by default on the release channel by adding `system_alloc = []` on the bottom of cargo.toml, while still allowing stable and beta to build the debug builds.

PR #41, opened 2018-02-13
Development branch

janus:
Re-structured files inside althea-types to use macro. This removed redundant codes


janus:
Cleaned out repeated and redundant codes from Althea_rs/althea_type.

PR #47, opened 2018-03-15
Warning hunt!

drozdziak1:
This is a pilot of me handling the unhandled, deleting the unused etc. I've only swatted a couple warnings in `clu` to get to know your comments/concerns/advice.

**Note** I didn't have the time to deal with `rita.sh` which was acting up on my machine due to unrelated reasons, so **please don't merge it** if you don't intend to test yourselves/if I didn't confirm that `rita.sh` is okay/if you're not confident that I don't break anything. Thanks!

### DIFF
--- a/clu/src/lib.rs
+++ b/clu/src/lib.rs
@@ -10,7 +10,7 @@ extern crate settings;
 
 extern crate ipgen;
 extern crate rand;
-use rand::Rng;
+use rand::{thread_rng, Rng};
 
 use std::str;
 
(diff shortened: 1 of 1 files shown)


clu/src/lib.rs
```diff
@@ -205,13 +205,13 @@ fn linux_init(
     let mesh_ip = SETTINGS.read().unwrap().network.own_ip.clone();
     let our_exit_ip = SETTINGS.read().unwrap().exit_client.exit_ip.clone();
 
-    request_own_exit_ip(SETTINGS.clone());
+    request_own_exit_ip(SETTINGS.clone())?;
```
kingoflolz: think we should just delete line 208 and 209, request own exit ip returns error when the exit is unreachable, which it is on init (the loop further down handles retrying that). Early returning from this function breaks rita

jkilpatr:
looks good to me. Based on this how much time will we need to de-warning the codebase?

kingoflolz:
@ttk2 I dewarned most of rita yesterday, I think there might be < 20 warnings across the whole codebase, so likely not very long

PR #55, opened 2018-04-06
rita.py: continue as soon as Babel converges

drozdziak1:
This commit raises the convergence delay to a minute and in exchange
polls for convergence every second so that we don't spend time blindly
waiting for CONVERGENCE_DELAY to pass


jkilpatr:
looks good to me!

PR #62, opened 2018-04-12
make settings more ergonomic and efficient

kingoflolz:


jkilpatr:
Nicely done!

clu/src/lib.rs
```diff
@@ -38,59 +39,27 @@ pub enum CluError {
     RuntimeError(String),
 }
 
-fn linux_generate_wg_keys(config: Arc<RwLock<settings::RitaSettings>>) -> Result<(), Error> {
+fn linux_generate_wg_keys(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), Error> {
```
drozdziak1: Do we need to explicitly state that it's a struct?

PR #66, opened 2018-04-16
Make rita.py check the installed routes

drozdziak1:
This commit makes rita.py ensure that optimum routes are installed when checking convergence.

### DIFF
--- a/integration-tests/rita.py
+++ b/integration-tests/rita.py
@@ -6,8 +6,9 @@ import errno
 import json
 import os
 import random
-import signal
+import re
 import shlex
+import signal
 import subprocess
 import sys
 import time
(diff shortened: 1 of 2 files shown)


integration-tests/rita.py
```diff
@@ -323,6 +375,98 @@ def test_reach_all(self, verbose=True, global_fail=True):
...
+        # all_routes = {
+        #     <where_from>: [
+        #                 (<where_to>, <price>, <next_hop>),
+        #                 [...]
+        #             ],
+        #     [...]
+        # }
+
+        all_routes = {
+                self.nodes[1]: [
```
kingoflolz: I think this shouldn't be self to be consistent with the rest of the dictionary

integration-tests/rita.py
```diff
@@ -414,33 +558,36 @@ def check_log_contains(f, x):
...
-    a = Node(1, 10)
-    b = Node(2, 25)
-    c = Node(3, 60)
-    d = Node(4, 10)
-    e = Node(5, 0)
-    f = Node(6, 50)
-    g = Node(7, 10)
-    # h = Node(8, 0)
-
+    a1 = Node(1, 10)
```
kingoflolz: any reason they got changed to letter-number? did it get too confusing lol

jkilpatr:
nice, so we went with hardcoding? looks like a clean way to do it at least.

integration-tests/rita.py
drozdziak1: True, thanks!

integration-tests/rita.py
drozdziak1: Yeah, didn't want to wonder which IP a node has every time

PR #67, opened 2018-04-17
route only specific lan interfaces over wg_exit

kingoflolz:
(cherry picked from commit 6641d4a)


althea_kernel_interface/src/ip_rule.rs
```diff
@@ -0,0 +1,29 @@
...
+                "route",
+                "add",
+                "default",
+                "via",
+                &format!("{}", exit_internal_ip),
+                "table",
+                "99",
+            ],
+        );
+        self.run_command("ip", &["rule", "add", "iif", interface, "table", "99"]);
```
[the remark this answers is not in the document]
kingoflolz: its just a random table which I picked, and its referenced in the command above

althea_kernel_interface/src/ip_rule.rs
```diff
@@ -0,0 +1,29 @@
...
+        self.run_command("ip", &["route", "flush", "cache"])?;
```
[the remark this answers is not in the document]
kingoflolz: we need to flush the cache so if we started to proxy a new interface over wg exit, routes are switched over ASAP instead of whenever the cache expires

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+        let endpoint_ip = format!("{}", endpoint_ip);
+
+        let mut def_route = Vec::new();
+
+        let mut found = false;
+
+        // find all lines
+        for i in stdout.split("\n") {
+            // starting with default
+            if i.starts_with("default") {
```
drozdziak1: Isn't `ip route list default` output only default rules? Why check for that?

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -21,6 +21,7 @@ impl KernelInterface {
         listen_port: u16,
         private_key_path: &str,
         local_ip: &IpAddr,
+        netmask: u32
```
drozdziak1: `u8` is more than enough for a netmask

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+                for j in i.split(" ") {
```
drozdziak1: Similarly to two comments above you could go `split_whitespace()`

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+        let stdout = String::from_utf8(output.stdout)?;
+
+        let endpoint_ip = format!("{}", endpoint_ip);
+
+        let mut def_route = Vec::new();
+
+        let mut found = false;
+
+        // find all lines
+        for i in stdout.split("\n") {
```
drozdziak1: I'd go with `lines()` here. I know this doesn't need to be too portable (only Linux really has iproute2), but I'd do it for the sake of sticking to good habits.

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+        if !found {
+            // early returning if no default route found, (e.g. on client)
+            return Ok(())
+        }
+
+        let mut default_is_wg_exit = def_route.contains(&String::from("wg_exit"));
+
+        if !default_is_wg_exit {
+            // update the default route if default route is not wg exit
+            *settings_default_route = def_route.clone();
```
drozdziak1: Are we editing the settings here? If so, why?

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        let mut def_route = settings_default_route.clone();
+
+        // change from default ... to <endpoint_ip> ...
+        def_route[0] = format!("{}", endpoint_ip);
```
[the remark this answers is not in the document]
drozdziak1: I'd delete the declaration on 25 and leave this one.

althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -35,11 +34,16 @@ fn is_link_local(ip: IpAddr) -> bool {
...
-fn socket_to_string(endpoint: &SocketAddr, interface_name: String) -> String {
+fn socket_to_string(endpoint: &SocketAddr, interface_name: Option<String>) -> String {
     match endpoint {
         &SocketAddr::V6(endpoint) => {
             if is_link_local(IpAddr::V6(endpoint.ip().clone())) {
-                format!("[{}%{}]:{}", endpoint.ip(), interface_name, endpoint.port())
+                format!(
+                    "[{}%{}]:{}",
+                    endpoint.ip(),
+                    interface_name.expect("Link local without interface"),
```
drozdziak1: Why on earth do we need to accept an `Option` as parameter if we're not willing to deal with a `None` here?

This clearly shows that the responsibility for the interface name lies on caller's part and **creates an avoidable unexpected `panic!()` opportunity for the program**

althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -58,13 +62,21 @@ impl KernelInterface {
...
+        let mut external_peer;
+
         let phy_name = match self.get_device_name(endpoint.ip()) {
-            Ok(phy_name) => phy_name,
-            Err(err) => external_nic.unwrap_or("no_nic".to_string()),
+            Ok(phy_name) => {
+                external_peer = false;
+                Some(phy_name)
+            },
+            Err(err) => {
```
drozdziak1: You're not using the `err`.

althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -58,13 +62,21 @@ impl KernelInterface {
...
+                external_peer = true;
+                external_nic
+            },
         };
-        let socket_connect_str = socket_to_string(endpoint, phy_name);
+        let socket_connect_str = socket_to_string(endpoint, phy_name.clone());
```
drozdziak1: why the `clone` if `phy_name` isn't used anywhere after this statement and `socket_to_string` doesn't mutate it?

althea_types/src/interop.rs
```diff
@@ -46,7 +46,7 @@ pub struct ExitClientIdentity {
 pub struct ExitServerIdentity {
     pub own_local_ip: IpAddr,
     pub server_local_ip: IpAddr,
-    pub netmask: IpAddr,
+    pub netmask: u32,
```
drozdziak1: `u8` would be enough

rita/src/client.rs
```diff
@@ -45,7 +45,7 @@ use actix::registry::SystemService;
 use actix::*;
 use actix_web::*;
 
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
```
drozdziak1: How's this change significant if there's no change whatsoever to any other part of this file?

rita/src/exit.rs
```diff
@@ -62,8 +62,7 @@ use rita_common::dashboard::network_endpoints::get_node_info;
 use rita_common::network_endpoints::{hello_response, make_payments};
 use rita_exit::network_endpoints::{list_clients, setup_request};
 
-use althea_kernel_interface::KI;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
```
drozdziak1: Same here, how does `Mutex` help?

althea_kernel_interface/src/manual_peers_route.rs
kingoflolz: it actually lists all the routes lol

althea_kernel_interface/src/manual_peers_route.rs
kingoflolz: good point

althea_kernel_interface/src/manual_peers_route.rs
kingoflolz: to preserve the default route, as we will overwrite it with a default route with wg exit later

rita/src/client.rs
kingoflolz: sorry these were rebase issues from another change

rita/src/exit.rs
kingoflolz: same here

althea_kernel_interface/src/open_tunnel.rs
kingoflolz: the point of doing an unwrap in this function is that the caller should either:
1. have a ipv6 link local with a interface name
2. have a ipv6 global address with no interface name
2. have a ipv4 global address with no interface name

It would be more complex (I think) to handle this outside the function, as this assumption should be true (basically) all the time.

althea_kernel_interface/src/manual_peers_route.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        self.run_command(
+            "ip",
+            &def_route_ref[..]
+        )?;
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
kingoflolz: Because we have to do the route splitting anyways, it seems harder to do regex for matching if you have to do some manual string splitting anyways

althea_kernel_interface/src/manual_peers_route.rs
drozdziak1: That's unfortunate :D

althea_kernel_interface/src/open_tunnel.rs
kingoflolz: the logic of the function below is we either get a valid device name from the ip (link local neighbor), or we pass in the external nic (which can either be Some or None (depending on client or gateway))

drozdziak1: You're right, sorry. Could you make this function return a Result though? This is a library crate after all and it would be nicer of us to leave the error for the library *user* to handle.

kingoflolz:
superseded by #65

PR #68, opened 2018-04-26
Cleaned up Clu's generate functions

janus:
Re-factored functions and removed repeated codes. 
Removed verbose Boolean logic


janus:
*I like the boolean simplification, not sure about adding the bool to the
kernel interface call. I do agree that the function name is vague though.*
 I lately I was thinking that enum would be a better choice.  Would that be
okay in the case?

There are a couple of functions that have almost same signature except
minor difference knd if there to add the minor difference in their
signature. But changing it to enum and using match is an option I would
want to explore.
As per the name, I just tweaked already  chosen name to avoid something
different.
I would change the fn name, but do you have anything in mind?

PR #71, opened 2018-04-26
Add traffic forwarding rules for exit tunnel

jkilpatr:
This nat's traffic from the lan nics to the exit tunnel, we can't
use the openwrt firewall rules for this like before because it requires
that the interface be there on startup. Also this is a more portable solution
where a random linux machine can run Rita and forward traffic like a router
without anyone having to monkey around.

Two major things to be concerned about when working with this, if you restart
openwrt's firewall Rita will crash and these rules need to be reinserted, maybe
we should put a dep in there somehow so that they get restarted together.

Second mtu clamping, I'm not sure it's perfect yet.

### DIFF
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -89,4 +89,50 @@ impl KernelInterface {
 
         Ok(())
     }
+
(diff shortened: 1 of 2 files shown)


jkilpatr:
Don't merge this yet, it's not done.

jkilpatr:
there's an mtu clamping issue somewhere I'm sure of it.

jkilpatr:
found it.

PR #73, opened 2018-04-28
Some code cleanup

jkilpatr:
Just some basic code cleanup and other helpful stuff. 

No more warnings in Kernel interface, but I've identified a few spots I would like to unit test, mostly to make sure code changes don't screw with proper output parsing for some of these commands. Otherwise the command runner and return codes make for pretty good coverage of failure cases.

### DIFF
--- a/althea_kernel_interface/src/counter.rs
+++ b/althea_kernel_interface/src/counter.rs
@@ -1,10 +1,9 @@
-use super::{KernelInterface, KernelInterfaceError};
+use super::KernelInterface;
 
 use std::collections::HashMap;
 use std::net::IpAddr;
 use std::str::FromStr;
 
-use eui48::MacAddress;
 use regex::Regex;
 
 use failure::Error;
@@ -69,7 +68,7 @@ impl KernelInterface {
                 "inet6",
                 "counters",
             ],
-        );
+        ).expect("ipset failed to create counter");
         self.add_iptables_rule(
             "ip6tables",
             &[
@@ -88,7 +87,7 @@ impl KernelInterface {
                 target.set_name(),
                 &format!("dst,{}", target.interface()),
             ],
-        )?;
+        ).expect("iptables match rule failed");
         Ok(())
     }
 
@@ -106,7 +105,7 @@ impl KernelInterface {
                 "inet6",
                 "counters",
             ],
-        );
+        ).expect("ipset read counters failed!");
 
         self.run_command(
             "ipset",
@@ -115,7 +114,7 @@ impl KernelInterface {
                 &format!("tmp_{}", target.set_name()),
                 target.set_name(),
             ],
-        )?;
+        ).expect("ipset swap failed");
 
         let output = self.run_command("ipset", &["save", &format!("tmp_{}", target.set_name())])?;
         let res = parse_ipset(&String::from_utf8(output.stdout)?);
--- a/althea_kernel_interface/src/create_wg_key.rs
+++ b/althea_kernel_interface/src/create_wg_key.rs
@@ -1,9 +1,9 @@
-use super::{KernelInterface, KernelInterfaceError};
+use super::KernelInterface;
 
 use std::fs::File;
-use std::io::{BufRead, BufReader, BufWriter, Read, Write};
+use std::io::Write;
 use std::path::Path;
-use std::process::{Command, Output, Stdio};
+use std::process::{Command, Stdio};
 
 use failure::Error;
 
@@ -21,13 +21,13 @@ impl KernelInterface {
     }
 
     pub fn create_wg_keypair(&self) -> Result<[String; 2], Error> {
-        let mut genkey = Command::new("wg")
+        let genkey = Command::new("wg")
             .args(&["genkey"])
             .stdout(Stdio::piped())
             .output()
             .unwrap();
 
-        let mut genstdout = genkey.stdout;
+        let genstdout = genkey.stdout;
         let mut pubkey = Command::new("wg")
             .args(&["pubkey"])
             .stdout(Stdio::piped())
@@ -35,7 +35,7 @@ impl KernelInterface {
             .spawn()
             .unwrap();
 
-        pubkey.stdin.as_mut().unwrap().write_all(&genstdout);
+        pubkey.stdin.as_mut().unwrap().write_all(&genstdout).expect("Failure generating wg keypair!");
         let output = pubkey.wait_with_output().unwrap();
 
         let mut privkey_str = String::from_utf8(genstdout)?;
@@ -47,4 +47,4 @@ impl KernelInterface {
     }
 }
 
-//TODO some tests
+//Tested in CLU
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -3,7 +3,6 @@ use super::{KernelInterface, KernelInterfaceError};
 use failure::Error;
 
 use std::net::{IpAddr, SocketAddr};
-use std::str::FromStr;
 
 impl KernelInterface {
     pub fn set_client_exit_tunnel_config(
@@ -35,7 +34,7 @@ impl KernelInterface {
             ],
         )?;
 
-        let output = self.run_command(
+        let _output = self.run_command(
             "ip",
             &[
                 "address",
@@ -66,7 +65,7 @@ impl KernelInterface {
     }
 
     pub fn set_route_to_tunnel(&self, gateway: &IpAddr) -> Result<(), Error> {
-        self.run_command("ip", &["route", "del", "default"]);
+        self.run_command("ip", &["route", "del", "default"]).expect("Failed to delete default route");
 
         let output = self.run_command(
             "ip",
@@ -103,19 +102,19 @@ impl KernelInterface {
                 "-j",
                 "MASQUERADE",
             ],
-        );
+        ).expect("Failure setting client tunnel MASQ");
         self.add_iptables_rule(
             "iptables",
             &[
                 "-A", "FORWARD", "-i", &lan_nic, "-o", "wg_exit", "-j", "ACCEPT",
             ],
-        );
+        ).expect("Failure adding lan to wg_exit forward");
         self.add_iptables_rule(
             "iptables",
             &[
                 "-A", "FORWARD", "-i", "wg_exit", "-o", &lan_nic, "-j", "ACCEPT",
             ],
-        );
+        ).expect("Failure adding wg_exit to lan forward");
         self.add_iptables_rule(
             "iptables",
             &[
@@ -130,7 +129,7 @@ impl KernelInterface {
                 "TCPMSS",
                 "--clamp-mss-to-pmtu", //should be the same as --set-mss 1300
             ],
-        );
+        ).expect("Failure to clamp mss");
         //TODO ipv6 support
 
         Ok(())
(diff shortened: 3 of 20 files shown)


althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -35,7 +34,7 @@ impl KernelInterface {
             ],
         )?;
 
-        let output = self.run_command(
+        let _output = self.run_command(
```
kingoflolz: if we dont use output I dont think we need to bind it to a variable at all

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -66,7 +65,7 @@ impl KernelInterface {
     }
 
     pub fn set_route_to_tunnel(&self, gateway: &IpAddr) -> Result<(), Error> {
-        self.run_command("ip", &["route", "del", "default"]);
+        self.run_command("ip", &["route", "del", "default"]).expect("Failed to delete default route");
```
kingoflolz: we might not have a default route to delete, but we should be a bit more explicit about ignoring errors

althea_kernel_interface/src/exit_server_counter.rs
```diff
@@ -108,7 +107,7 @@ impl KernelInterface {
                 "inet6",
                 "counters",
             ],
-        );
+        ).expect("Failed to create new ipset to swap");
```
kingoflolz: this can crash if rita crashes during the counter swap, and make it crash every 5 seconds, might be better to ignore the error

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -48,7 +46,7 @@ impl KernelInterface {
 
         self.run_command(&command, &arg_str[..])?;
 
-        let output = self.run_command(
+        let _output = self.run_command(
```
kingoflolz: same here

althea_kernel_interface/src/manipulate_uci.rs
```diff
@@ -28,7 +28,7 @@ impl KernelInterface {
 
     //Sets an arbitrary UCI list on OpenWRT
     pub fn set_uci_list(&self, key: &str, value: &[&str]) -> Result<bool, Error> {
-        self.del_uci_var(&key);
+        self.del_uci_var(&key)?;
```
kingoflolz: list might not exist, so it is not necessary to bail here

clu/src/lib.rs
```diff
@@ -137,17 +135,17 @@ pub fn cleanup() -> Result<(), Error> {
 
     for i in interfaces {
         if re.is_match(&i) {
-            KI.del_interface(&i)?;
+            KI.del_interface(&i).expect("failed to delete a w# interface");
         }
     }
 
-    KI.del_interface("wg_exit");
+    KI.del_interface("wg_exit").expect("failed to delete wg_exit");
```
kingoflolz: wg exit might not exist

althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -6,7 +6,7 @@ use std::path::Path;
 use failure::Error;
 
 #[derive(Debug, Fail)]
-pub enum TunnelOpeningError {
+pub enum _TunnelOpeningError {
```
kingoflolz: think we should just delete this if we are not using it

althea_kernel_interface/src/lib.rs
```diff
@@ -30,7 +24,6 @@ mod exit_client_tunnel;
 mod exit_server_counter;
 mod exit_server_tunnel;
 mod get_neighbors;
-mod get_wg_pubkey;
```
kingoflolz: should we delete the entire `get_wg_pubkey` file?

jkilpatr: it doesn't seem to do anything anymore.

althea_kernel_interface/src/counter.rs
```diff
@@ -69,7 +68,7 @@ impl KernelInterface {
                 "inet6",
                 "counters",
             ],
-        );
+        ).expect("ipset failed to create counter");
```
drozdziak1: Wouldn't a `?` fit in this and similar calls better? IMO a library crate like this one (and especially as thin as this one) should leave error handling to the caller at all times.

althea_kernel_interface/src/create_wg_key.rs
```diff
@@ -21,21 +21,21 @@ impl KernelInterface {
...
     pub fn create_wg_keypair(&self) -> Result<[String; 2], Error> {
-        let mut genkey = Command::new("wg")
+        let genkey = Command::new("wg")
             .args(&["genkey"])
             .stdout(Stdio::piped())
             .output()
             .unwrap();
 
-        let mut genstdout = genkey.stdout;
+        let genstdout = genkey.stdout;
```
drozdziak1: Is a 4-character difference in 2 uses worth a separate variable for this?

althea_kernel_interface/src/exit_server_counter.rs
drozdziak1: +1 but I'd also log it with `warn!()`

althea_kernel_interface/src/exit_server_counter.rs
```diff
@@ -117,7 +116,7 @@ impl KernelInterface {
                 &format!("tmp_{}", target.set_name()),
                 target.set_name(),
             ],
-        )?;
```
drozdziak1: What's wrong with using `?`?

althea_kernel_interface/src/manipulate_uci.rs
drozdziak1: I'd log with `trace!` here

althea_kernel_interface/src/open_tunnel.rs
drozdziak1: +1

clu/src/lib.rs
```diff
@@ -65,7 +63,7 @@ fn linux_generate_mesh_ip(config: &mut NetworkSettings) -> Result<(), Error> {
 }
 
 fn validate_wg_key(key: &str) -> bool {
-    key.len() == 44 && key.ends_with("=")
+    key.len() == 44 && key.ends_with("=") && !key.contains(" ")
```
drozdziak1: I wonder if this function wouldn't benefit from returning `Result`, because right now if something's wrong it won't tell the caller what.

drozdziak1: This applies to `validate_mesh_ip` too

clu/src/lib.rs
```diff
@@ -84,11 +82,11 @@ fn linux_setup_exit_tunnel(config: Arc<RwLock<settings::RitaSettingsStruct>>) ->
         details.own_internal_ip,
         details.netmask,
     )?;
-    KI.set_route_to_tunnel(&details.server_internal_ip);
+    KI.set_route_to_tunnel(&details.server_internal_ip).expect("Failed to set default route");
```
drozdziak1: Why not `?`?

scripts/openwrt_build.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-export OPENWRT_DIR=/home/justin/repos/althea-firmware/build
+export OPENWRT_DIR=CHANGEME
```
drozdziak1: simple and useful :+1:

scripts/openwrt_build.sh
```diff
@@ -1,5 +1,5 @@
...
 export TOOLCHAIN=toolchain-mipsel_24kc_gcc-7.3.0_musl
 export TARGET_NAME=target-mipsel_24kc_musl
```
drozdziak1: I wonder if we could ask it to pick the OpenWRT build's current toolchain somehow

scripts/openwrt_upload.sh
```diff
@@ -2,6 +2,4 @@
 set -eux
 bash scripts/openwrt_build.sh
 scp target/mipsel-unknown-linux-musl/debug/rita root@192.168.1.1:/tmp/rita
-scp settings/default.toml root@192.168.1.1:/etc/rita-default.toml
-scp scripts/rita-test.toml root@192.168.1.1:/etc/rita.toml
-ssh root@192.168.1.1 RUST_BACKTRACE=FULL RUST_LOG=TRACE /tmp/rita --config /etc/rita.toml  --default /etc/rita-default.toml --platform openwrt
+ssh root@192.168.1.1 RUST_BACKTRACE=FULL RUST_LOG=TRACE /tmp/rita --config /etc/rita.toml --platform linux &> out.log
```
drozdziak1: Wouldn't it be better to keep the IP in an env? Also, is `/tmp/` guaranteed to be in-RAM on all platforms? Personally I like using `scp` for this on `/usr/sbin/rita` itself cause it'll substitute the binary in-place.

althea_kernel_interface/src/exit_client_tunnel.rs
jkilpatr: I'm pretty sure _ and _output are the same the compiler will optimize out actually saving the value.

althea_kernel_interface/src/create_wg_key.rs
jkilpatr: not really, fixed

althea_kernel_interface/src/exit_server_counter.rs
jkilpatr: nothing, I just swaped this with a warn statement.

clu/src/lib.rs
jkilpatr: well lets think about what could go wrong, it has to be a string, rust will enforce that, it's a string all these functions will work unless there's a stdlib error which I don't think we should concern ourselves with catching at every corner case.

clu/src/lib.rs
jkilpatr: ok this is a good case to propogate up, you could conceivably clean up rotues properly and handle it.

scripts/openwrt_build.sh
jkilpatr: you could try to parse it out of the .config....

scripts/openwrt_upload.sh
jkilpatr: we must use /tmp and yes it's guaranteed to be in ram. Binaries with debug are like 36mb which is more than the storage of lets say an n600 but they fit in ram just fine. So sadly we kinda have to do it this way. As for changing the ip, sure we can environment that out, but it should at least be the default for ease of use.

clu/src/lib.rs
kingoflolz: because the function can only fail one way, I think its fine to use a bool

althea_kernel_interface/src/counter.rs
kingoflolz: actually, I think this error should not be .expected because I think it raises an error if there is already a counter by that name, which will make it crash when it restarts

clu/src/lib.rs
```diff
@@ -137,17 +135,21 @@ pub fn cleanup() -> Result<(), Error> {
 
     for i in interfaces {
         if re.is_match(&i) {
-            KI.del_interface(&i)?;
+            KI.del_interface(&i).expect("failed to delete a w# interface");
```
drozdziak1: Are we not able to at least tell them which interface that was?

drozdziak1: Also, intuition tells me cleanup routines wouldn't usually stop after a problem appears, i.e. if someone manually deletes an interface which is on this list and `del_interface` fails because of that, why should that ever affect the rest of the cleanup?

jkilpatr: this was just an oversight on my part I'll put it back to ?

PR #78, opened 2018-05-04
Fix Babel testing problems

drozdziak1:
This commit fixes an issue where Babel would not work properly with a
SharedMockStream containing more than one pre-pushed babeld output,
which has to work if we want to test Babel consumers, who in turn mostly
issue more than one Babel call.

babel_monitor:
* Fix the buffering issue by moving back to a struct with BufStream
as member and flushing output on writes
* Add an error type for no terminator present in Babel output - we might
want to handle this upon stopping babeld daemons
* Verbosely inform about Babel responses

rita_client::traffic_watcher:
rita_common::traffic_watcher:
rita_exit::traffic_watcher:
* Adjust for changes in Babel
* Add an ignored test for using a localhost:9001 socket to debug the
watcher's behavior; it's got logging enabled

tunnel_manager:
* Adjust for changes in Babel

### DIFF
--- a/rita/Cargo.toml
+++ b/rita/Cargo.toml
@@ -16,35 +16,36 @@ default = []
 system_alloc = []
 
 [dependencies]
-clippy = { version = "*", optional = true }
-babel_monitor = { path = "../babel_monitor" }
-diesel = { version = "1.0.0", features = ["sqlite"] }
-dotenv = "0.9.0"
-althea_types = { path = "../althea_types", features = ["actix"]}
 althea_kernel_interface = { path = "../althea_kernel_interface" }
-exit_db = { path = "../exit_db" }
-settings = { path = "../settings" }
+althea_types = { path = "../althea_types", features = ["actix"]}
+babel_monitor = { path = "../babel_monitor" }
 clu = { path = "../clu" }
+exit_db = { path = "../exit_db" }
 num256 = { path = "../num256" }
-ip_network = "0.1"
-log = "^0.4"
+settings = { path = "../settings" }
+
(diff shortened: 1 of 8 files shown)


rita/Cargo.toml
```diff
@@ -16,35 +16,36 @@ default = []
 system_alloc = []
 
 [dependencies]
-clippy = { version = "*", optional = true }
```
kingoflolz: Does rustfmt start to sort deps now as well?

drozdziak1: No, it's just me, I forgot to split that to another commit, sorry. I divided this into a in-project dep pile and an external dep one.

rita/src/rita_exit/traffic_watcher/mod.rs
```diff
@@ -151,8 +152,15 @@ pub fn watch(mut babel: Box<Babel>, clients: Vec<Identity>) -> Result<(), Error>
...
 #[cfg(test)]
 mod tests {
+    extern crate env_logger;
+
+    use super::*;
+
     #[test]
-    fn it_works() {
-        assert_eq!(2 + 2, 4);
+    #[ignore]
```
jkilpatr: does this mean the tests aren't actually being run?

drozdziak1: A test seemed ideal to put this little routine in. A #[] statement works only for the one thing under it, in this case only this one "test".

drozdziak1: However, one can run it with `cargo test -p rita -- --test <test_name>`

PR #79, opened 2018-05-07
Remove unnecessary use of For loops

janus:
Remove repeated(redundant ) use of For loop. And also remove  For loop for map initialization. 
Clean up unused code.


jkilpatr:
the three rustfmt commits at the end need to be squashed. And I think the redundant code ones should be squashed to although that's more of an opinion.

janus:
About doing that

PR #84, opened 2018-05-14
Update rustfmt syntax

jkilpatr:
Minor rustfmt fix, then a more serious feature addition to handle althea-dash


.travis.yml
```diff
@@ -21,9 +21,9 @@ script:
 matrix:
   exclude:
   - rust: stable
-    env: TEST_COMMAND="cargo install rustfmt-nightly --force && cargo fmt --all -- --write-mode=diff"
+    env: TEST_COMMAND="cargo install rustfmt-nightly --force && cargo fmt --all --"
```
drozdziak1: @jkilpatr By default, `rustfmt` (or `cargo fmt`, the cargo's convenience wrapper around it) will quietly rewrite the project's files into the formatting rules it knows. The `--write-mode` flag used to ask it to do it differently i.e. to just print a diff and exit with non-zero status to indicate misformatted code.

Then, they added a specialized write mode called `check` which would do the same with some minute difference from `diff`.

**As for now, you most likely want to do `cargo fmt --all -- --check`, It seems that they changed `--write-mode=check` into a separate `--check` flag and dropped `--write-mode` altogether**

.travis.yml
```diff
@@ -21,9 +21,9 @@ script:
...
   - rust: beta
-    env: TEST_COMMAND="cargo install rustfmt-nightly --force && cargo fmt --all -- --write-mode=diff"
+    env: TEST_COMMAND="cargo install rustfmt-nightly --force && cargo fmt --all --"
```
drozdziak1: Probably same here

.travis.yml
jkilpatr: thanks, updated.

PR #86, opened 2018-05-15
Update web-interface headers and parsing

jkilpatr:
These changes are required to get the wifi interface page of the dash
working. We need to allow cross site requests from the address of the
server as well as change the way some of the parsing works to be
compatible with the dashboard implementation.


rita/src/middleware.rs
```diff
@@ -0,0 +1,36 @@
+use actix_web::middleware::{Middleware, Response, Started};
+use actix_web::{HttpRequest, HttpResponse, Result};
+use http::{header, HttpTryFrom, Method, StatusCode};
+use regex::Regex;
+
+pub struct Headers; // <- Our middleware
+
+/// Middleware implementation, middlewares are generic over application state,
+/// so you can access state with `HttpRequest::state()` method.
```
jtremback: Are these comments from copy-pasted example code? IMO comments are only useful when you are doing something surprising or very unusual, not annotating obvious code.

kingoflolz: Yeah, think these comments should be removed

PR #87, opened 2018-05-16
Multiexit config

kingoflolz:


rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -29,15 +106,43 @@ impl Handler<Tick> for ExitManager {
...
+            SETTING
+                .get_exit_client()
+                .get_current_exit()
+                .map(|c| c.clone())
+        };
+
+        // code that connects to the current exit server
+        if let Some(exit) = exit_server {
+            if let Some(ref details) = exit.details {
+                linux_setup_exit_tunnel().expect("failure setting up exit tunnel");
```
jkilpatr: this may be safe name just setup_exit_tunnel nothing changes with this operation on opewnrt.

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -29,15 +106,43 @@ impl Handler<Tick> for ExitManager {
...
+                        wg_public_key: details.wg_public_key.clone(),
+                        eth_address: details.eth_address,
+                    },
+                    details.exit_price,
+                ));
+            }
+        }
+
+        // code that manages requesting details to exits
+        let servers = { SETTING.get_exits().clone() };
```
jkilpatr: we send a setup request to every server by default? I guess that's ok, especially since we don't do it again after we get a response and they go to state Registered.

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -1,14 +1,91 @@
...
 use rita_client::traffic_watcher::{TrafficWatcher, Watch};
 
+use althea_types::{ExitServerReply, ExitState};
 use failure::Error;
+use settings::ExitClientDetails;
+use std::net::SocketAddr;
+use KI;
+
+fn linux_setup_exit_tunnel() -> Result<(), Error> {
+    let exit_client = SETTING.get_exit_client();
```
jkilpatr: so current exit will always be in the registered state?

jkilpatr: or do we store a state but still go through the registration process on every exit tunnel setup? To be replaced with a login process at some point.

rita/src/rita_client/exit_manager/mod.rs
kingoflolz: yeah, we will make a registration request to every exit on the list

rita/src/rita_client/exit_manager/mod.rs
kingoflolz: we only run setup_exit_tunnel when we know the current exit is in a registered state, and if it is already registered, we don't try registering again

integration-tests/rita.py
```diff
@@ -78,12 +87,12 @@ def exec_or_exit(command, blocking=True, delay=0.01):
 
 
 def cleanup():
-    os.system("rm -rf *.log *.pid *.toml private-key*")
+    os.system("sudo rm -rf *.log *.pid *.toml private-key*")
```
drozdziak1: There's literally no point as rita.py is run with `sudo` in `rita.sh` at line 58

integration-tests/rita.py
```diff
@@ -109,6 +118,9 @@ def get_veth_interfaces(self):
             interfaces.append("veth-{}-{}".format(self.id, i))
         return interfaces
 
+    def get_pubkey(self):
+        return get_rita_settings(self.id)["network"]["wg_public_key"]
```
drozdziak1: If you're using this method only once then what's the point?

integration-tests/rita.py
```diff
@@ -213,6 +225,10 @@ def save_rita_settings(id, x):
     toml.dump(x, open("rita-settings-n{}.toml".format(id), "w"))
 
 
+def get_rita_settings(id):
```
drozdziak1: Come on, not only are you using this only once, but also in a different, self-contained function where this could be inlined. I'd keep this one and delete `get_pubkey`

integration-tests/rita.py
```diff
@@ -345,12 +361,16 @@ def create(self):
...
+
         time.sleep(1)
 
+        exit_settings["exits"]["exit_a"]["id"]["wg_public_key"] = self.nodes[self.exit].get_pubkey()
+
         print("starting rita")
         for id, node in self.nodes.items():
             if id == self.exit:
-                start_rita_exit(node)
+                pass
```
drozdziak1: Why does this if branch exist if it's not supposed to do anything?

rita/src/exit.rs
```diff
@@ -161,6 +162,7 @@ fn main() {
         App::new()
             .resource("/setup", |r| r.method(Method::POST).with2(setup_request))
             .resource("/list", |r| r.method(Method::POST).with(list_clients))
+            .resource("/exit_info", |r| r.method(Method::GET).with(get_exit_info))
```
drozdziak1: What's the point of a separate endpoint for this? Couldn't this just be `/setup` with GET?

settings/src/lib.rs
```diff
@@ -99,41 +101,84 @@ impl Default for PaymentSettings {
...
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct ExitServer {
+    pub id: Identity,
+    pub registration_port: u16,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub our_details: Option<ExitClientDetails>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub general_details: Option<ExitDetails>,
```
drozdziak1: How's calling this "general" details descriptive of the field's role? Why are ExitClientDetails stored here and why is it called `our_details` when it's obviously not the exit server's details in that field?

settings/src/lib.rs
```diff
@@ -99,41 +101,84 @@ impl Default for PaymentSettings {
...
+    }
+    pub fn get_current_exit_details(&self) -> Option<&ExitClientDetails> {
+        match self.get_current_exit() {
+            Some(ref exit) => match exit.our_details {
+                Some(ref details) => return Some(details),
+                _ => (),
+            },
+            _ => (),
+        }
+        None
```
drozdziak1: Why not use the `?` syntax here? This is far more complicated than it could be.

settings/src/lib.rs
```diff
@@ -99,41 +101,84 @@ impl Default for PaymentSettings {
...
+    }
+    pub fn set_current_exit(&mut self) -> Option<&mut ExitServer> {
```
drozdziak1: change to `get_current_exit_mut`, same for all other "setters"

settings/src/lib.rs
```diff
@@ -99,41 +101,84 @@ impl Default for PaymentSettings {
...
+        }
+        if self.exits.contains_key(self.current_exit.as_ref().unwrap()) == false {
+            return false;
+        }
+        return self.exits[self.current_exit.as_ref().unwrap()]
+            .our_details
+            .is_some();
+    }
+    pub fn reg_details_set(&self) -> bool {
+        self.reg_details.is_some()
```
drozdziak1: Writing the body of this function takes literally 4 more letters. Why does it exist? Also, why the getters and setters for all the different fields, but not `reg_details`?

settings/src/lib.rs
```diff
@@ -151,8 +196,7 @@ pub struct ExitTunnelSettings {
 pub struct RitaSettingsStruct {
     payment: PaymentSettings,
     network: NetworkSettings,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    exit_client: Option<ExitClientSettings>,
+    exit_client: ExitClientSettings,
```
drozdziak1: What's the point of a separate `ExitClientSettings` struct if it's not an Option here?

settings/src/lib.rs
```diff
@@ -534,10 +542,9 @@ where
 
     thread::spawn(move || {
         let old_settings = settings.read().unwrap().clone();
+        thread::sleep(Duration::from_millis(200));
```
drozdziak1: What's this for?

settings/src/lib.rs
```diff
@@ -575,19 +584,14 @@ impl RitaSettingsStruct {
...
-            details.wg_public_key.clone(),
-        ))
+        Some(self.exit_client.get_current_exit().as_ref()?.id.clone())
     }
 }
 
 impl FileWrite for RitaSettingsStruct {
     fn write(&self, file_name: &str) -> Result<(), Error> {
-        let ser = toml::to_string(&self).unwrap();
+        let ser = toml::Value::try_from(self.clone()).unwrap();
```
drozdziak1: unwraps

settings/src/lib.rs
```diff
@@ -618,7 +622,8 @@ impl RitaExitSettingsStruct {
 
 impl FileWrite for RitaExitSettingsStruct {
     fn write(&self, file_name: &str) -> Result<(), Error> {
-        let ser = toml::to_string(&self).unwrap();
+        let ser = toml::Value::try_from(self.clone()).unwrap();
```
drozdziak1: unwraps

settings/src/lib.rs
kingoflolz: I like to segregate stuff which make sense as a single struct, and so that methods can be implemented on it

rita/src/exit.rs
kingoflolz: getting setup doesn't really make sense to me, as it is general info about the exit

settings/src/lib.rs
kingoflolz: `ExitServer` is a struct which represents an exit server on the client, `our details` mean details for the exit which is specific to the client and `general_details` means details which is the same for all clients of that particular exit. Any suggestions for less confusing names?

rita/src/exit.rs
drozdziak1: Well, it's setup both ways - for the client when you `GET` and for the exit when you `POST`

settings/src/lib.rs
drozdziak1: How is putting 4 fields in a separate struct justified? How could it possibly serve readability?

rita/src/rita_client/exit_manager/mod.rs
jkilpatr: then where do we store various connection details? I'm not seeing key and port stored in our config for the long term.

kingoflolz: here's the relevant part of the settings struct where we store these
```
{
   "exit_client":{
      "current_exit":"exit_a",
      "exits":{
         "exit_a":{
            "general_details":{
               "description":"just a normal althea exit",
               "exit_price":50,
               "netmask":24,
               "server_internal_ip":"172.168.1.254",
               "wg_exit_port":59999
            },
            "id":{
               "eth_address":"0x0101010101010101010101010101010101010101",
               "mesh_ip":"fd00::5",
               "wg_public_key":"vUryCFcZwqkhXtaTnYczhWX5so9DZnu5b0Gx6QvcA34="
            },
            "message":"Registration OK",
            "our_details":{
               "client_internal_ip":"172.168.1.102"
            },
            "registration_port":4875,
            "state":"Registered"
         }
      },
      "reg_details":{
         "email":"1234@gmail.com",
         "zip_code":"1234"
      },
      "wg_listen_port":59999
   }
}
```

settings/src/lib.rs
kingoflolz: Hrm, the intention was so that when you say jumped from somewhere that used this struct, you would only find relevant fields and methods to what you were looking for, and also implementing methods on the top level struct is a little painful, as you have to do .read().unwrap() on every access to self. With jump to definition and things like that, I don't think the cons are that bad

kingoflolz: @jkilpatr any thoughts on this issue?

settings/src/lib.rs
drozdziak1: You're right, `our_details` makes sense. How about just `details` for `general_details`?

settings/src/lib.rs
drozdziak1: Okay, my argument is invalid, there's other member structs which would also need to follow the same rules for consistency if we were to go down that path (`payment`, `network` members). And that for real could make stuff hard to read.

jkilpatr:
we really have to figure out these test failures.

PR #90, opened 2018-05-18
Update lock

jkilpatr:
This updates our cargo.lock and moves us away from @kingoflolz pinned Actix version. Just some general housekeeping to keep things on the latest versions. 

I tried using the cargo-exit crates cargo upgrade command to update all of our top level dependencies too, but that brings on a set of futures errors I don't want to go monkeying around with, I'd appreciate from feedback on the state of native futures and their integration into Actix. Is it stable enough that we should be eyeing those changes?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 *.bak
 /target/
+/rls/
 target-mac
 **/*.rs.bk
 .idea
--- a/stats_server/Cargo.toml
+++ b/stats_server/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["kingoflolz <wangben3@gmail.com>"]
 
 [dependencies]
 actix = "0.5.7"
-actix-web = {version = "0.5.7" , default-features = false}
+actix-web = {version = "0.6.7" , default-features = false}
 serde = "1.0.24"
 serde_derive = "1.0.24"
 serde_json = "1.0.8"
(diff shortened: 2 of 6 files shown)


kingoflolz:
@jkilpatr right now you _should_ be able to upgrade to futures 0.1.whatever, but futures 0.2 is a breaking change which actix doesn't support right now. The plan for async await is for part of futures to move into rust stdlib, and make futures 0.3 simply reexport the stdlib types. At that point, actix plans to move to async await syntax along with upgrading to 0.3

stats_server/Cargo.toml
```diff
@@ -4,8 +4,8 @@ version = "0.1.0"
 authors = ["kingoflolz <wangben3@gmail.com>"]
 
 [dependencies]
-actix = {git="https://github.com/kingoflolz/actix", branch="mocking"}
-actix-web = {git="https://github.com/kingoflolz/actix-web", default-features = false}
+actix = "0.5.7"
+actix-web = {version = "0.5.7" , default-features = false}
```
kingoflolz: actix web doesn't follow the same version as actix, so the latest version of actix web is actually 0.6.7

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 *.bak
 /target/
+/rls/
```
drozdziak1: How come RLS makes its directory in project root? For me it's always in `target/rls`

PR #92, opened 2018-05-21
Add version and git hash to rita help text, HTTP endpoint and log output

kingoflolz:


rita/src/rita_common/network_endpoints/mod.rs
```diff
@@ -56,3 +56,11 @@ pub fn hello_response(
         })
         .responder()
 }
+
+pub fn version(_req: HttpRequest) -> String {
+    format!(
+        "crate ver {}\ngit hash {}",
```
drozdziak1: Food for thought: We might want to attach the version as a field of every request and response, like many protocols do.

jkilpatr:
rustfmt needs to be run it seems.

PR #95, opened 2018-05-21
Jehan build

jtremback:
reformat for clarity and functionality


jkilpatr:
So I'll just rename this once I get the toolchains built. But it's a good improvement.

PR #97, opened 2018-05-22
Make Rita understand RTT:

drozdziak1:
This PR makes Rita understand and independently measure the full-path round
trip time to the exit.

### DIFF
--- a/api-docs/exit.md
+++ b/api-docs/exit.md
@@ -1,0 +1,37 @@
+# Rita exit daemon's HTTP endpoints
+This file documents the facilities available on the Althea exit server daemon
+via HTTP.
+
+## Port `registration_port`
+The endpoints below are served on the port configured using the
+`registration_port` config value.
+
+### `/rtt`
+Measure the round trip time to the exit.
+
+* **Method**: `GET`
+* **URL Params**: `None`
+* **Data Params**: `None`
+* **Success Response**:
+  - **Code**: 200 OK
+  - **Contents**:
+```javascript
+// Note: Both exit_rx and exit_tx are a JSON-serialized representation of Rust's
+// std::time::SystemTime.
+{
+  "exit_rx": {
+  "secs_since_epoch":1527106071, // Integer; timestamp seconds at request arrival
+  "nanos_since_epoch":609010634  // Integer; timestamp nanos at request arrival
+  },
+  "exit_tx": {
+  "secs_since_epoch": 1527106071, // Integer; timestamp seconds at response departure
+  "nanos_since_epoch":609011002   // Integer; timestamp nanos at response departure
+  }
+}
+```
+* **Error Response**: `n/a`
+* **Sample call**:
+```sh
+$ curl <exit_ip>:<exit_registration_port>/rtt
+{"exit_rx":{"secs_since_epoch":1527106071,"nanos_since_epoch":609010634},"exit_tx":{"secs_since_epoch":1527106071,"nanos_since_epoch":609011002}}
+```
--- a/babel_monitor/src/lib.rs
+++ b/babel_monitor/src/lib.rs
@@ -287,7 +287,10 @@ ok\n";
     fn mock_dump() {
         let mut s = SharedMockStream::new();
         s.push_bytes_to_read(TABLE.as_bytes());
-        s.write(b"dump\n").unwrap();
+
+        let mut b = Babel::new(s);
+        let dump = b.command("dump").unwrap();
+        assert_eq!(&dump, TABLE);
     }
 
     #[test]
--- a/babel_monitor/src/lib.rs
+++ b/babel_monitor/src/lib.rs
@@ -159,7 +159,7 @@ impl<T: Read + Write> Babel<T> {
     }
 
     pub fn monitor(&mut self, iface: &str) -> Result<(), Error> {
-        let _ = self.command(&format!("interface {}", iface))?;
+        let _ = self.command(&format!("interface {} enable-timestamps true", iface))?;
         info!("Babel started monitoring: {}", iface);
         Ok(())
     }
@@ -217,7 +217,7 @@ impl<T: Read + Write> Babel<T> {
                     prefix: find_babel_val("prefix", entry)?.parse()?,
                     metric: find_babel_val("metric", entry)?.parse()?,
                     refmetric: find_babel_val("refmetric", entry)?.parse()?,
-                    full_path_rtt: find_babel_val("full-path-rtt", entry)?.parse()?,
+                    full_path_rtt: { find_babel_val("full-path-rtt", entry)?.parse()? },
                     price: find_babel_val("price", entry)?.parse()?,
                     fee: find_babel_val("fee", entry)?.parse()?,
                 });
@@ -365,6 +365,7 @@ ok\n";
 
         let route = routes.get(0).unwrap();
         assert_eq!(route.price, 3072);
+        assert_eq!(route.full_path_rtt, 22.805);
 
         b.command("interface wg0").unwrap();
     }
--- a/babel_monitor/src/lib.rs
+++ b/babel_monitor/src/lib.rs
@@ -59,6 +59,7 @@ pub struct Route {
     pub prefix: IpNetwork,
     pub metric: u16,
     pub refmetric: u16,
+    pub full_path_rtt: f32,
     pub price: u32,
     pub fee: u32,
 }
@@ -216,6 +217,7 @@ impl<T: Read + Write> Babel<T> {
                     prefix: find_babel_val("prefix", entry)?.parse()?,
                     metric: find_babel_val("metric", entry)?.parse()?,
                     refmetric: find_babel_val("refmetric", entry)?.parse()?,
+                    full_path_rtt: find_babel_val("full-path-rtt", entry)?.parse()?,
                     price: find_babel_val("price", entry)?.parse()?,
                     fee: find_babel_val("fee", entry)?.parse()?,
                 });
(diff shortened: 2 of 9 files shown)


babel_monitor/src/lib.rs
```diff
@@ -59,6 +59,7 @@ pub struct Route {
     pub prefix: IpNetwork,
     pub metric: u16,
     pub refmetric: u16,
+    pub full_path_rtt: f32,
```
drozdziak1: FYI i chose `f32` for effortless parsing. Babeld operates on `int`s though and should we choose to follow its steps closer, we might want to parse the socket output "on foot".

jkilpatr: f32 is float right? I don't think that's going to parse correctly versus babel's 'thousands' format which isn't even close to real floating point.

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -101,16 +92,52 @@ pub fn watch<T: Read + Write>(
...
+        let RTTimestamps { exit_rx, exit_tx } = client
+            .get(&format!(
+                "http://[{}]:{}/rtt",
+                exit.mesh_ip,
+                match SETTING.get_exit_client().get_current_exit() {
+                    Some(current_exit) => current_exit.registration_port,
+                    None => {
+                        error!("No current exit even though an exit route is present");
+                        return Err(format_err!(
+                            "No current exit even though an exit route is present"
```
jkilpatr: couldn't we not duplicate that string somehow? Minor nit.

rita/src/exit.rs
```diff
@@ -175,6 +175,7 @@ fn main() {
             .resource("/setup", |r| r.method(Method::POST).with2(setup_request))
             .resource("/list", |r| r.method(Method::POST).with(list_clients))
             .resource("/exit_info", |r| r.method(Method::GET).with(get_exit_info))
+            .resource("/rtt", |r| r.method(Method::GET).with(rtt))
```
jkilpatr: this needs to be api doc'd, make a folder doc in the top level and add a markdown format file called endpoints.md with api documentation.

rita/src/rita_exit/network_endpoints/mod.rs
```diff
@@ -69,3 +70,10 @@ pub fn list_clients(_req: HttpRequest) -> Box<Future<Item = Json<Vec<Client>>, E
         .and_then(move |reply| Ok(Json(reply?)))
         .responder()
 }
+
+pub fn rtt(_req: HttpRequest) -> Result<Json<RTTimestamps>> {
```
jkilpatr: needs rustdoc

babel_monitor/src/lib.rs
drozdziak1: @jkilpatr It's actually pretty close, except no exp notation is at play. Babel simply "divides" the value by 1000 and pretends it's float by splitting the formatting:
```c
const char *
format_thousands(unsigned int value)
{
    static char buf[4][15];
    static int i = 0;
    i = (i + 1) % 4;
    snprintf(buf[i], 15, "%d.%.3d", value / 1000, value % 1000);
    return buf[i];
}
```

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -101,16 +92,52 @@ pub fn watch<T: Read + Write>(
     trace!("exit price {}", exit_price);
 
     if destinations.contains_key(&exit.mesh_ip) {
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(5))
```
kingoflolz: maybe a shorter timeout is better, because we have this running every 5 seconds and we don't want them to pile up

kingoflolz: Also maybe async would be a good thing to add here, as we don't want to block too much, it's a bit tricky in rust but let me know if you want to give it a shot

rita/src/rita_client/traffic_watcher/mod.rs
kingoflolz: if we print out the return value from this function, the error! would be unnecessary

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -101,16 +92,52 @@ pub fn watch<T: Read + Write>(
...
+        trace!(
+            "RTTs: per-hop {}ms, inner {}ms",
+            destinations[&exit.mesh_ip].full_path_rtt,
+            inner_rtt_millis
+        );
+
         trace!(
             "exit destination price {}",
-            destinations[&exit.mesh_ip].clone() + exit_price
+            Int256::from(destinations[&exit.mesh_ip].clone().price) + exit_price
```
kingoflolz: I think .price.clone() would be better (if it works)

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -101,16 +92,52 @@ pub fn watch<T: Read + Write>(
...
         );
+        trace!("Exit ip: {:?}", exit.mesh_ip);
+        trace!("Exit destination:\n{:#?}", destinations[&exit.mesh_ip]);
 
         owes += Int256::from(exit_price * output);
 
-        owes += (destinations[&exit.mesh_ip].clone() + exit_price) * input;
+        owes += (Int256::from(destinations[&exit.mesh_ip].clone().price) + exit_price) * input;
```
kingoflolz: same here

kingoflolz:
just thinking out loud, I wonder if it would be a good idea to test this in our integration tests by introducing artificial latency into the network (network-lab supports using `tc` per link, so it should be a reasonably straightforward thing to add (assuming it can work consistently and doesn't exacerbate our CI failure rate lol))

rita/src/rita_client/traffic_watcher/mod.rs
drozdziak1: @kingoflolz Long RTTs will be entirely possible in congested networks. As for async the delay between taking the client_tx timestamp and the actual transfer of the request is important to be minimized - introducing an additional potentially non-deterministic delay between `client_tx` assignment and when `futures` schedules to make the request would be a big footshot.

api-docs/exit.md
```diff
@@ -0,0 +1,37 @@
...
+  "nanos_since_epoch":609011002   // Integer; timestamp nanos at response departure
+  }
+}
+```
+* **Error Response**: `n/a`
+* **Sample call**:
+```sh
+$ curl <exit_ip>:<exit_registration_port>/rtt
+{"exit_rx":{"secs_since_epoch":1527106071,"nanos_since_epoch":609010634},"exit_tx":{"secs_since_epoch":1527106071,"nanos_since_epoch":609011002}}
+```
```
drozdziak1: This is how I'd do the API docs. I'd put **all** exit entrypoints in here and segregate them by port as shown above. I'd generally keep three files like this one for the client, exit and common stuff, respectively.

babel_monitor/src/lib.rs
jkilpatr: do you think 'pretty close' is a one of the sources of the current inaccuracy?

rita/src/rita_client/traffic_watcher/mod.rs
jkilpatr: it might be better to have a thread and block, just to make sure we don't get into any futures bs.

api-docs/exit.md
jkilpatr: seems like a good system @kingoflolz thoughts?

jkilpatr:
@drozdziak1 what do you think about unit/integration tests for this functionality?

rita/src/rita_client/traffic_watcher/mod.rs
drozdziak1: I'm using `reqwest` synchronously and this fragment of code doesn't depend on anything async, presently async problems shouldn't be the case here.

babel_monitor/src/lib.rs
drozdziak1: @jkilpatr Lol I just checked `ping` in Babel's tests vs. the RTT. I might've been barking up the wrong tree the whole time, as the value wildly diverge there too

drozdziak1:
@jkilpatr I think integration testing this can get difficult pretty quick. We'd need severe network obstructions (so that the noise in readings can be kept at bay) which would impact test duration. it'd be good to have this be a benchmark of the RTT's accuracy.

babel_monitor/src/lib.rs
jkilpatr: by 'this problem' do you mean parsing the thousands value and doing operations on it or just time measurement with the endpoints?

jkilpatr:
yeha I'm not going to propose anything comprehensive yet, just enough to make sure the basics work as that's where we are at now.

babel_monitor/src/lib.rs
drozdziak1: I meant the measurement

api-docs/exit.md
jkilpatr: what's the resolution of system time by default?

drozdziak1: @jkilpatr It's nanos

jkilpatr:
@kingoflolz could you update your review here?

rita/src/rita_client/traffic_watcher/mod.rs
kingoflolz: makes sense

api-docs/exit.md
kingoflolz: Good idea, I'll wait til this lands to change all existing rita docs over

PR #99, opened 2018-05-25
Gateway robustness (WIP)

kingoflolz:
- [x] Add manual routes for DNS resolvers
- [ ] Add gateway detection
- [ ] Stop opening per hop tunnels with hostnames when not a gateway


rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -78,6 +81,21 @@ impl SystemService for TunnelManager {
             self.listen_interfaces.insert(i);
         }
         trace!("Loaded listen interfaces {:?}", self.listen_interfaces);
+
+        if !cfg!(test) {
+            Arbiter::registry().init_actor(|_| {
+                //TODO: make the configurable when trust-dns-resolver serde issue is solved
+                //default is 8.8.8.8
+                Connector::new(ResolverConfig::default(), ResolverOpts::default())
```
jkilpatr: wait we're not even configuring the server? that's one hell of a default.

althea_kernel_interface/src/interface_tools.rs
```diff
@@ -24,6 +25,15 @@ impl KernelInterface {
         self.run_command("ip", &["link", "del", "dev", name])?;
         Ok(())
     }
+
+    pub fn iface_status(&self, iface: &str) -> Result<String, Error> {
+        // cat so we can mock
+        let output = self.run_command("cat", &[&format!("/sys/class/net/{}/operstate", iface)])?;
```
jkilpatr: so we cat here rather than just reading the file for easy mocking? Could we mock the file stream instead? Not sure if that's actually better though.

kingoflolz: We already have the infra to mock commands, so this is the past of least resistance (we almost ever read files)

jkilpatr:
@kingoflolz so right now the DNS is hardcoded to 8.8.8.8? Is changing that the extent of what we are waiting for?

kingoflolz:
@ttk2 yeah, basically

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -78,6 +83,24 @@ impl SystemService for TunnelManager {
             self.listen_interfaces.insert(i);
         }
         trace!("Loaded listen interfaces {:?}", self.listen_interfaces);
+
+        #[cfg(not(test))]
+        {
+            Arbiter::registry().init_actor(|_| {
+                //TODO: make the configurable when trust-dns-resolver serde issue is solved
+                //default is 8.8.8.8
+                Connector::new(ResolverConfig::default(), ResolverOpts::default())
```
jkilpatr: have you tried setting dnssec to true?

kingoflolz: not yet, I think we can try that when can change the settings via config file instead of messing with the code, but it _does_ say dnssec validation is implemented (FWIW)

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -17,6 +17,10 @@ use std::net::SocketAddr;
 use KI;
 
 fn linux_setup_exit_tunnel() -> Result<(), Error> {
+    {
+        KI.update_settings_route(&mut SETTING.get_network_mut().default_route)?;
+    }
```
drozdziak1: How does keeping this scope here help?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -664,6 +723,7 @@ mod tests {
             }
 
             counter += 1;
+            trace!("program {:?}, args {:?}, ctr {}", program, args, counter);
```
drozdziak1: ctr -> counter?

jkilpatr:
So testing this on my router it doesn't seem to work. Did you try it on yours?

PR #102, opened 2018-05-28
Fix neighbors http

jkilpatr:
Taken from #100 implements the plumbing for Babel to query neighbours. Disables a lot of the breaking rtt stuff that @drozdziak1 is probably working on in #97 

I need to talk with @drozdziak1 and confirm a few things about our babel interface. 

Also I'm starting to speculate that our endpoints are very variable in response time (between .2 seconds and several seconds) because we have a single thread that may be blocked by various tasks.

### DIFF
--- a/api-docs/router-dashboard/get-neighbors.md
+++ b/api-docs/router-dashboard/get-neighbors.md
@@ -18,20 +18,20 @@ GET
 
 ```
 [
-   	{
-   		"nickname": "fd00::2",
-   		"route_metric_to_exit": 0,
-   		"total_debt": 0,
-   		"current_debt": 0,
-         "link_cost": 0
-   	},
-   	{
-   		"nickname": "fd00::7",
-   		"route_metric_to_exit": 0,
-   		"total_debt": 0,
-   		"current_debt": 0,
-         "link_cost": 0
-   	}
+   {
+      "nickname": "fd00::2",
+      "route_metric_to_exit": 0,
+      "total_debt": 0,
+      "current_debt": 0,
+      "link_cost": 0
+   },
+   {
+      "nickname": "fd00::7",
+      "route_metric_to_exit": 0,
+      "total_debt": 0,
+      "current_debt": 0,
+      "link_cost": 0
+   }
 ]
 ```
 
--- a/api-docs/router-dashboard/get-neighbors.md
+++ b/api-docs/router-dashboard/get-neighbors.md
@@ -17,18 +17,20 @@ GET
 ### Content
 
 ```
-   [
+[
    	{
    		"nickname": "fd00::2",
    		"route_metric_to_exit": 0,
-   		"total_payments": 0,
-   		"debt": 0
+   		"total_debt": 0,
+   		"current_debt": 0,
+         "link_cost": 0
    	},
    	{
    		"nickname": "fd00::7",
    		"route_metric_to_exit": 0,
-   		"total_payments": 0,
-   		"debt": 0
+   		"total_debt": 0,
+   		"current_debt": 0,
+         "link_cost": 0
    	}
 ]
 ```
--- a/api-docs/router-dashboard/get-neighbors.md
+++ b/api-docs/router-dashboard/get-neighbors.md
@@ -23,14 +23,16 @@ GET
       "route_metric_to_exit": 0,
       "total_debt": 0,
       "current_debt": 0,
-      "link_cost": 0
+      "link_cost": 0,
+      "price_to_exit": 0
    },
    {
       "nickname": "fd00::7",
       "route_metric_to_exit": 0,
       "total_debt": 0,
       "current_debt": 0,
-      "link_cost": 0
+      "link_cost": 0,
+      "price_to_exit": 0
    }
 ]
 ```
--- a/babel_monitor/src/lib.rs
+++ b/babel_monitor/src/lib.rs
@@ -30,6 +30,8 @@ pub enum BabelMonitorError {
     ReadFailed(String),
     #[fail(display = "No terminator after Babel output:\n{}", _0)]
     NoTerminator(String),
+    #[fail(display = "No Neighbor was found matching address:\n{}", _0)]
+    NoNeighbor(String),
 }
 
 use BabelMonitorError::*;
@@ -49,7 +51,7 @@ fn find_babel_val(val: &str, line: &str) -> Result<String, Error> {
     return Err(VariableNotFound(String::from(val), String::from(line)).into());
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Route {
     pub id: String,
     pub iface: String,
@@ -63,9 +65,10 @@ pub struct Route {
     pub fee: u32,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Neighbor {
     pub id: String,
+    pub address: IpAddr,
     pub iface: String,
     pub reach: u16,
     pub txcost: u16,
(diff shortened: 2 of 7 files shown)


rita/src/rita_client/dashboard/mod.rs
```diff
@@ -156,3 +162,75 @@ impl Handler<SetWifiConfig> for Dashboard {
...
+    fn handle(&mut self, _msg: GetNodeInfo, _ctx: &mut Self::Context) -> Self::Result {
+        Box::new(
+            DebtKeeper::from_registry()
+                .send(Dump {})
+                .and_then(|res| {
+                    let stream = TcpStream::connect::<SocketAddr>(
+                        format!("[::1]:{}", SETTING.get_network().babel_port)
+                            .parse()
+                            .unwrap(),
+                    ).expect("Failed to connect to Babel!");
```
jkilpatr: bunches of expects here, mainly because the question of "where to propagate the error to?" seems to be nowhere. We can't handle this outside of the request function because the response will never reach the requester. So all errors must be resolved here.

babel_monitor/src/lib.rs
```diff
@@ -186,12 +189,13 @@ impl<T: Read + Write> Babel<T> {
...
                 vector.push_back(Neighbor {
                     id: find_babel_val("neighbour", entry)?,
+                    address: find_babel_val("address", entry)?.parse()?,
                     iface: find_babel_val("if", entry)?,
                     reach: u16::from_str_radix(&find_babel_val("reach", entry)?, 16)?,
                     txcost: find_babel_val("txcost", entry)?.parse()?,
                     rxcost: find_babel_val("rxcost", entry)?.parse()?,
-                    rtt: find_babel_val("rtt", entry)?.parse()?,
-                    rttcost: find_babel_val("rttcost", entry)?.parse()?,
+                    rtt: 0.0,
```
jkilpatr: @drozdziak1  fixing these in the firmware and the like is part of the task of full path rtt.

babel_monitor/src/lib.rs
```diff
@@ -223,6 +227,38 @@ impl<T: Read + Write> Babel<T> {
         }
         Ok(vector)
     }
+
+    /// In this function we take a route snapshot then loop over the routes list twice
```
jkilpatr: as much as I complain about the speed of this at 2*routes time it's within constant time of the fastest it could be (assuming we don't make babel sort routes somehow so we can search them more efficiently). So it can probably stay unless there's an obvious way to get the neighbor ip's I'm missing.

rita/src/exit.rs
```diff
@@ -189,7 +189,6 @@ fn main() {
             // assuming exit nodes dont need wifi
             //.resource("/wifisettings", |r| r.route().filter(pred::Get()).h(get_wifi_config))
             //.resource("/wifisettings", |r| r.route().filter(pred::Post()).h(set_wifi_config))
```
jkilpatr: we will need a different endpoint for the exit.

api-docs/router-dashboard/get-neighbors.md
```diff
@@ -17,19 +17,23 @@ GET
...
-   		"total_payments": 0,
-   		"debt": 0
-   	},
-   	{
-   		"nickname": "fd00::7",
-   		"route_metric_to_exit": 0,
-   		"total_payments": 0,
-   		"debt": 0
-   	}
+[
```
drozdziak1: So what doc convention are rolling with? Is there any reason why #97 can't go in first and for other api-touching PRs to follow suit?

rita/src/client.rs
```diff
@@ -162,7 +162,7 @@ fn main() {
                 r.method(Method::POST).with2(make_payments)
             })
             .resource("/hello", |r| r.method(Method::POST).with2(hello_response))
-    }).workers(1)
+    }).workers(2)
```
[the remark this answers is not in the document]
jkilpatr: I was testing response latency, this didn't help, let me fix it.

api-docs/router-dashboard/get-neighbors.md
jkilpatr: sure I guess it could, I just pulled it over here for ease of review.

babel_monitor/src/lib.rs
jkilpatr: @drozdziak1 is #97 ready for me to merge/base on? I'm really trying to cut a stable version this week so I don't want to merge bigger changes.

babel_monitor/src/lib.rs
kingoflolz: One thing we might want to do in the future is have all babel methods operate over some sort of BabelSnapshot, so we don't need to parse routes/hit babel every time

rita/src/rita_client/dashboard/mod.rs
```diff
@@ -156,3 +162,75 @@ impl Handler<SetWifiConfig> for Dashboard {
...
+                                price_to_exit: route.price,
+                            })
+                        } else {
+                            output.push(NodeInfo {
+                                nickname: serde_json::to_string(&identity.mesh_ip).unwrap(),
+                                route_metric_to_exit: 0,
+                                total_payments: debt_info.total_payment_recieved.clone().into(),
+                                debt: debt_info.debt.clone().into(),
+                                link_cost: 0,
+                                price_to_exit: 0,
```
kingoflolz: these should probrobly be something other than 0, probably an Option (as price_to_exit _might_ be 0)

rita/src/rita_client/dashboard/mod.rs
```diff
@@ -156,3 +162,75 @@ impl Handler<SetWifiConfig> for Dashboard {
...
+                            })
+                        }
+                    }
+
+                    futures::future::ok(output)
```
kingoflolz: maybe we can return a Result instead of a Future, so you can use ?, then the consumer of the message can handle the errors

babel_monitor/src/lib.rs
```diff
@@ -223,6 +227,38 @@ impl<T: Read + Write> Babel<T> {
...
+    ) -> Result<Route, Error> {
+        let routes = self.parse_routes()?;
+        // First find the neighbors route to itself to get the local address
+        for neigh_route in routes.iter() {
+            // This will fail on v4 babel routes etc
+            if let IpNetwork::V6(ref ip) = neigh_route.prefix {
+                if ip.get_network_address() == neigh_mesh_ip {
+                    let neigh_local_ip = neigh_route.neigh_ip;
+                    // Now we take the neigh_local_ip and search for a route via that
+                    for route in routes.iter() {
```
kingoflolz: IMO I would break this into 2 sequential loops instead of a nested one (unless there is a particular reason for it)

rita/src/rita_client/dashboard/mod.rs
jkilpatr: How does that make a difference here?

rita/src/rita_client/dashboard/mod.rs
jkilpatr: but what consumer? the actix library?

babel_monitor/src/lib.rs
jkilpatr: mostly me not wanting to make a placeholder neighbour object to deal with the scoping issue, it was more annoying that it should have been.

rita/src/rita_client/dashboard/mod.rs
kingoflolz: Tells the JS that this is an invalid value, so it can not show it on the UI

rita/src/rita_client/dashboard/mod.rs
kingoflolz: The HTTP handler functions in network_endpoints

babel_monitor/src/lib.rs
kingoflolz: Ok fair enough

rita/src/rita_client/dashboard/mod.rs
jkilpatr: isn't that as high as our handling can go?

rita/src/rita_client/dashboard/mod.rs
jkilpatr: so return 404? that's probably better I agree.

jkilpatr: @kingoflolz so I played around with this some more, the types just really don't work out with the box's. Any ideas? I notice you did the same thing in wifi settings as I'm doing here so I'm pretty sure you didn't find a better way back then.

PR #104, opened 2018-05-31
tunnel manager refactor

kingoflolz:
- [x] fix tests
- [x] write up how it works
- [x] test behind NAT ect

### DIFF
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -162,12 +162,7 @@ fn main() {
             .resource("/make_payment", |r| {
                 r.method(Method::POST).with2(make_payments)
             })
-            .resource("/hello", |r| {
-                r.method(Method::POST).with2(hello_response_old)
-            })
-            .resource("/hello", |r| {
-                r.method(Method::POST).with2(hello_response_new)
-            })
+            .resource("/hello", |r| r.method(Method::POST).with2(hello_response))
     }).workers(1)
         .bind(format!("[::0]:{}", SETTING.get_network().rita_hello_port))
         .unwrap()
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -162,7 +162,12 @@ fn main() {
             .resource("/make_payment", |r| {
                 r.method(Method::POST).with2(make_payments)
             })
-            .resource("/hello", |r| r.method(Method::POST).with2(hello_response))
+            .resource("/hello", |r| {
+                r.method(Method::POST).with2(hello_response_old)
+            })
+            .resource("/hello", |r| {
+                r.method(Method::POST).with2(hello_response_new)
+            })
     }).workers(1)
         .bind(format!("[::0]:{}", SETTING.get_network().rita_hello_port))
         .unwrap()
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -74,15 +74,17 @@ use rita_common::network_endpoints::*;
 struct Args {
     flag_config: String,
     flag_platform: String,
+    flag_future: bool,
 }
 
 #[cfg(not(test))]
 lazy_static! {
     static ref USAGE: String = format!(
-        "Usage: rita_exit --config=<settings> --platform=<platform>
+        "Usage: rita_exit --config=<settings> --platform=<platform> [--future]
 Options:
-    -c, --config=<settings>   Name of config file
+    -c, --config=<settings>     Name of config file
     -p, --platform=<platform>   Platform (linux or openwrt)
+    --future                    Enable B side of A/B releases
 About:
     Version {}
     git hash {}",
@@ -123,6 +125,8 @@ lazy_static! {
         let platform = args.flag_platform;
 
         let s = RitaSettingsStruct::new_watched(&settings_file).unwrap();
+
+        s.set_future(args.flag_future);
 
         clu::init(&platform, s.clone());
 
--- a/rita/src/rita_common/http_client/mod.rs
+++ b/rita/src/rita_common/http_client/mod.rs
@@ -13,7 +13,7 @@ use actix::registry::SystemService;
 
 use serde_json;
 
-use althea_types::LocalIdentity;
+use althea_types::{Identity, LocalIdentity};
 
 use failure::Error;
 
@@ -53,7 +53,7 @@ impl Actor for HTTPSyncExecutor {
 
 #[derive(Debug, Eq, PartialEq)]
 pub struct Hello {
-    pub my_id: LocalIdentity,
+    pub my_id: Identity,
     pub to: SocketAddr,
 }
 
--- a/rita/src/rita_common/http_client/mod.rs
+++ b/rita/src/rita_common/http_client/mod.rs
@@ -15,6 +15,9 @@ use serde_json;
 
 use althea_types::{Identity, LocalIdentity};
 
+use settings::RitaCommonSettings;
+use SETTING;
+
 use failure::Error;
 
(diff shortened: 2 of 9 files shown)


jkilpatr:
@kingoflolz this doesn't seem to setup peer tunnels properly.

rita/src/client.rs
```diff
@@ -162,7 +162,12 @@ fn main() {
             .resource("/make_payment", |r| {
                 r.method(Method::POST).with2(make_payments)
             })
-            .resource("/hello", |r| r.method(Method::POST).with2(hello_response))
+            .resource("/hello", |r| {
+                r.method(Method::POST).with2(hello_response_old)
+            })
+            .resource("/hello", |r| {
+                r.method(Method::POST).with2(hello_response_new)
+            })
```
drozdziak1: Just how hacky is this? Is it something like a fallback, where when one request fails Actix tries the other?

rita/src/rita_common/http_client/mod.rs
```diff
@@ -73,7 +73,12 @@ impl Handler<Hello> for HTTPSyncExecutor {
 
     fn handle(&mut self, msg: Hello, _: &mut Self::Context) -> Self::Result {
         info!("sending {:?}", msg);
-        let my_id = serde_json::to_string(&msg.my_id)?;
+
+        // TODO: REMOVE IN ALPHA 5
```
drozdziak1: @kingoflolz @jkilpatr Where would be the best place to note this? Our Trello usage somewhat died of natural causes, maybe we need a different approach? It would be nice to have a fix for this actually scheduled somewhere

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -390,6 +418,31 @@ impl TunnelManager {
         babel.monitor(&tunnel.iface_name)?;
         Ok(())
     }
+
+    /// Given a LocalIdentity, connect to the neighbor over wireguard
```
drozdziak1: How's this function distinct from `open_tunnel`? It's got the same description and yet it exists

rita/src/rita_common/http_client/mod.rs
jkilpatr: https://forum.altheamesh.com/t/tunnel-manager-change/38

We put it there but I'll admit trello is nice for this sort of stuff. What are your thoughts?

jkilpatr:
@drozdziak1 what do you think of similar backwards compatiblity testing as we have in Babel?

rita/src/client.rs
kingoflolz: yeah, it tries the handlers in order, (they take different json payloads, and its not a "match" (but also not a fail) in actix's eyes if the payload fails to deserialize)

rita/src/rita_common/tunnel_manager/mod.rs
kingoflolz: This opens a tunnel, but _does not connect over to the neighbor, only opens a wg interface which listens_, sorry I'll update the docs

rita/src/client.rs
drozdziak1: Okay, so I understood correctly that this isn't a good practice? would you mind marking this as "not a good practice" in a comment?

rita/src/rita_common/http_client/mod.rs
drozdziak1: Awesome! Yeah, I don't think we will find anything great anytime soon, each one of us probably has their favourite way of juggling tasks. Maybe some good Trello <-> Github integration would help because all the Trello goodness is already in place but you need to manually go and mark the cards done when coding work is done, which probably is the root cause of us abandoning Trello. That and the size of our team (and project) which IMHO still isn't big enough to be seriously threatened by forgotten hacks.

drozdziak1: Another option is to move all code-related tasks to GitHub issues altogether and only manage theory/design problems on Trello

jkilpatr: I'm not opposed to issues, we just have to keep up with them. I've liked the forum so far in that a post makes reporting easier.

jkilpatr:
tested and validated with old exit -> new client

still needs testing with old client -> new exit (is borked.altheamesh.com running this @kingoflolz ?)

jkilpatr:
old client -> new exit does not seem to work, no per hop tunnel when pointed at borked.altheamesh.com

kingoflolz:
@jkilpatr should be working now, tested on my edgerouter

rita/src/rita_common/http_client/mod.rs
```diff
@@ -73,7 +76,16 @@ impl Handler<Hello> for HTTPSyncExecutor {
 
     fn handle(&mut self, msg: Hello, _: &mut Self::Context) -> Self::Result {
         info!("sending {:?}", msg);
-        let my_id = serde_json::to_string(&msg.my_id)?;
+
+        let my_id = if SETTING.get_future() {
```
jkilpatr: I don't think this works, this says that we take either or not that we can take both?

jkilpatr: what if half the nodes have updated and the other half have not?

jkilpatr:
tested old exit new client, new client new exit + future, new exit new client no future. All seem to have worked. 

But when trying old client + new client behind the same Nat I got a crash on rita-exit 

Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: INFO 2018-06-03T18:00:22Z: rita_exit::rita_common::network_endpoints: sending to old hello handler
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: INFO 2018-06-03T18:00:22Z: rita_exit::rita_common::network_endpoints: Got Hello from Some("[::ffff:172.125.171.96]:47768")
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: INFO 2018-06-03T18:00:22Z: rita_exit::rita_common::tunnel_manager: creating new wg interface TunnelData { iface_name: "wg1", listen_port: 60001 }
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: INFO 2018-06-03T18:00:22Z: althea_kernel_interface: Command "ip" ["link", "set", "dev", "wg1", "up"] returned: an error Output { status: ExitStatus(ExitStatus(512)), stdout: "", stderr: "RTNETLINK answers: Address already in use\n" }
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RuntimeError("received error setting wg interface up: RTNETLINK answers: Address already in use\n")', libcore/result.rs:945:5
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: note: Run with `RUST_BACKTRACE=1` for a backtrace.
Jun 03 14:00:22 vps171948.vps.ovh.ca rita_exit[3232]: ERROR 2018-06-03T18:00:22Z: actix_web::pipeline: Error occured during request handling, status: 500 Internal Server Error Mailbox has closed
Jun 03 14:00:22 vps171948.vps.ovh.ca systemd[1]: rita-exit.service: main process exited, code=exited, status=101/n/a
Jun 03 14:00:22 vps171948.vps.ovh.ca systemd[1]: Unit rita-exit.service entered failed state.
Jun 03 14:00:22 vps171948.vps.ovh.ca systemd[1]: rita-exit.service failed.

PR #107, opened 2018-06-04
Document ExitState enum in api-docs

jkilpatr:


jkilpatr:
not sure how we want to do this.

api-docs/router-dashboard/set-settings.md
```diff
@@ -1,5 +1,16 @@
 # Router dashboard API
 
+## COMMENTS
```
drozdziak1: This changeset is still using endpoint-per-file approach which is not very readable. @kingoflolz Didn't you want to rewrite your docs in the service-per-file fashion? Your stuff probably got merged because of no activity. Can we hold our horses on merging this? Here's an [issue](https://github.com/althea-mesh/althea_rs/issues/113) so that we don't forget about this.

PR #114, opened 2018-06-05
DON'T MERGE Attempt at making wan always work

jkilpatr:
I'm not trying to get this merged, as it seems to break more than
it fixes, it's mostly here to discuss some code.

### DIFF
--- a/althea_kernel_interface/src/get_neighbors.rs
+++ b/althea_kernel_interface/src/get_neighbors.rs
@@ -1,5 +1,6 @@
 use super::KernelInterface;
 
+use std::collections::HashSet;
 use std::net::IpAddr;
 use std::str::FromStr;
 
@@ -25,8 +26,8 @@ impl KernelInterface {
         Ok(vec)
     }
 
-    pub fn trigger_neighbor_disc(&self) -> Result<(), Error> {
-        for interface in self.get_interfaces()? {
+    pub fn trigger_neighbor_disc(&self, interfaces: &HashSet<String>) -> Result<(), Error> {
+        for interface in interfaces {
             self.run_command("ping6", &["-c1", "-I", &interface, "ff02::1"])?;
         }
         Ok(())
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -44,6 +44,8 @@ type Connector = actors::Connector;
 pub enum TunnelManagerError {
     #[fail(display = "DNS lookup error")]
     DNSLookupError,
+    #[fail(display = "No Dev for peer!")]
+    NoDevError,
 }
 
 #[derive(Debug, Clone)]
@@ -251,7 +253,7 @@ impl TunnelManager {
     /// This gets the list of link-local neighbors, and then contacts them to get their
     /// Identity using `neighbor_inquiry` as well as their wireguard tunnel name
     pub fn get_neighbors(&mut self) -> ResponseFuture<Vec<(LocalIdentity, String, IpAddr)>, Error> {
-        KI.trigger_neighbor_disc().unwrap();
+        KI.trigger_neighbor_disc(&self.listen_interfaces).unwrap(); // Pings broadcast address to populate ip neigh
         let neighs: Vec<
             Box<Future<Item = Option<(LocalIdentity, String, IpAddr)>, Error = ()>>,
         > = KI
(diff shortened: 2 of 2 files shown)


althea_kernel_interface/src/get_neighbors.rs
```diff
@@ -25,8 +26,8 @@ impl KernelInterface {
         Ok(vec)
     }
 
-    pub fn trigger_neighbor_disc(&self) -> Result<(), Error> {
-        for interface in self.get_interfaces()? {
+    pub fn trigger_neighbor_disc(&self, interfaces: &HashSet<String>) -> Result<(), Error> {
```
jkilpatr: this may be a decent change, I have no idea why we find neighs on all interfaces then discard the ones not on peer interfaces.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -261,18 +263,24 @@ impl TunnelManager {
...
+                    let idx = settings
+                        .default_route
+                        .iter()
+                        .position(|s| s == "dev")
+                        .unwrap();
+                    iface = Some(settings.default_route[idx + 1].clone());
+                }
+
+                for i in settings.manual_peers.clone() {
+                    out.push((i, iface.clone()))
```
jkilpatr: while this gets us the right interface it doesn't actually cause the peer discovery to work. I assume the http request doens't get routed right even though ping -i <wan dev> 1.1.1.1 works regardless of the default route, wonder why that is.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -310,15 +318,17 @@ impl TunnelManager {
         let iface_index = if let Some(dev) = dev.clone() {
             KI.get_iface_index(&dev).unwrap()
         } else {
-            0
+            return Box::new(futures::future::err(TunnelManagerError::NoDevError.into()));
```
jkilpatr: What does dev "none" normally do exactly? does it mean it takes the default route?

jkilpatr:
I think the real conclusion here may just be that screwing with the default route enough to make this work is a bad idea.

althea_kernel_interface/src/get_neighbors.rs
kingoflolz: Looks good to me, I just wanted to make it work 100% of the time so I just pinged everyone, but this should be more efficient. I wonder if there can be any edge case where pinging one interface gets you a response from another interface, but in that case the network is so broken that its not worth trying to probe

rita/src/rita_common/tunnel_manager/mod.rs
kingoflolz: 0 means do whatever the kernel routing table tells us to do, instead of specifying a specific interface (should not be an error)

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -310,15 +318,17 @@ impl TunnelManager {
...
 
+        let dev = dev.unwrap();
+
         Box::new(
             Connector::from_registry()
                 .send(actors::Resolve::host(their_hostname.clone()))
                 .from_err()
                 .and_then(move |res| {
-                    let url = format!("http://[{}%{:?}]:4876/hello", their_hostname, dev);
+                    let url = format!("http://[{}%{}]:4876/hello", their_hostname, dev);
```
kingoflolz: we don't actually use URL anywhere except for debug IIRC, so this shouldn't effect anything

rita/src/rita_common/tunnel_manager/mod.rs
kingoflolz: None stands for the wan interface (sorry probs not very clear, so I think what you did to get the wan interface name is redundant (but it might be better to always explicitly specify interfaces))

rita/src/rita_common/tunnel_manager/mod.rs
drozdziak1: Why do we take `dev` as an `Option` here if we're not going to handle a `None` gracefuly?

rita/src/rita_common/tunnel_manager/mod.rs
drozdziak1: @jkilpatr you probably want the `find()` method here instead of `position()`. Also, why the `unwrap`?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -251,7 +253,7 @@ impl TunnelManager {
     /// This gets the list of link-local neighbors, and then contacts them to get their
     /// Identity using `neighbor_inquiry` as well as their wireguard tunnel name
     pub fn get_neighbors(&mut self) -> ResponseFuture<Vec<(LocalIdentity, String, IpAddr)>, Error> {
-        KI.trigger_neighbor_disc().unwrap();
+        KI.trigger_neighbor_disc(&self.listen_interfaces).unwrap(); // Pings broadcast address to populate ip neigh
```
drozdziak1: Why the `unwrap`? Is the situation that hopeless here?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -310,15 +318,17 @@ impl TunnelManager {
         let iface_index = if let Some(dev) = dev.clone() {
             KI.get_iface_index(&dev).unwrap()
         } else {
-            0
+            return Box::new(futures::future::err(TunnelManagerError::NoDevError.into()));
         };
 
+        let dev = dev.unwrap();
```
drozdziak1: `unwrap`, contradictory with the explicit handling above

PR #115, opened 2018-06-05
Only ping interfaces configured for peer discovery

jkilpatr:
It's wasteful to ping all interfaces and then remove entries we don't
need.


jkilpatr:
so these tests pass on my machine but not in CI, interesting.

jkilpatr:
ah found it, was looking at the test command wrong.

jkilpatr:
don't really understand how these tests are laid out. I'll pair with you @kingoflolz

PR #117, opened 2018-06-06
Add backwards-compat to integration tests

drozdziak1:
This commit enables us and CI to perform our integration tests between
two arbitrary althea_rs releases. By default it compares
althea-mesh/althea_rs' master and release branches.

Note: the tests might go nuts if the way we read data in the tests off
Rita changes - i.e. a situation where two releases which work perfectly
in production and don't pass integration tests is possible.

.gitignore:
* Start ignoring temporary release repo directories

.travis.yml:
* Start testing against the random, old_exit and new_exit compat test
layouts

rita.py:
* Add a mapping between nodes and assigned releases; use each release's
binaries accordingly
* Display release information in printouts
* Make some of the global variables full caps

rita.sh:
* Remove redundant functions
* Automatically decide whether to pull and build Babel
* If COMPAT_LAYOUT is chosen pull and build the specified releases or
use the defaults; if not, fall back to previous behavior


jkilpatr:
impressive, now we can make sure it works, then make sure it fails by comparing to a very old version. 

@drozdziak1 could you add a doc that goes over what the integration tests cover somewhere? I'm not sure where we would want to put it.

PR #121, opened 2018-06-09
Only ping interfaces configured for peer discovery

kingoflolz:
It's wasteful to ping all interfaces and then remove entries we don't
need.

supersedes #115

### DIFF
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -251,7 +251,8 @@ impl TunnelManager {
     /// This gets the list of link-local neighbors, and then contacts them to get their
     /// Identity using `neighbor_inquiry` as well as their wireguard tunnel name
     pub fn get_neighbors(&mut self) -> ResponseFuture<Vec<(LocalIdentity, String, IpAddr)>, Error> {
-        KI.trigger_neighbor_disc().unwrap();
+        KI.trigger_neighbor_disc(&SETTING.get_network().peer_interfaces)
+            .unwrap();
         let neighs: Vec<
             Box<Future<Item = Option<(LocalIdentity, String, IpAddr)>, Error = ()>>,
         > = KI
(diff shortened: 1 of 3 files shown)


rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -268,11 +269,6 @@ impl TunnelManager {
             })
             .filter_map(|(ip_address, dev)| {
                 info!("neighbor at interface {:?}, ip {}", dev, ip_address,);
-                if let Some(dev) = dev.clone() {
```
kingoflolz: @jkilpatr any reason you removed this?

jkilpatr: this is where we filter out entries we don't need right?

kingoflolz: yeah, but why was it removed? so now we are only triggereing neighbor disc on listen interfaces but we connect on all interfaces

jkilpatr: I'll need to look at this more carefully. Do you think it's causing the connection loss problem?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -853,213 +857,91 @@ mod tests {
...
+                            })
+                        }
+                        _ => {
+                            panic!("ip link called too many times")
+                        }
+                    }
                 }
-                _ => panic!("command called too many times"),
+                (program, args) => (
+                    panic!("unimplemented program! {} {:?}", program, args)
```
drozdziak1: I'd change the `!` to `:`, it's a bit more clear that way

jkilpatr:
@drozdziak1 this actually breaks networking! So lets not merge it until we figure that out.

jkilpatr:
or maybe it's fixed now?

PR #122, opened 2018-06-09
fix exit geoip

kingoflolz:
fix exit geoip by stopping rita from crashing when it strays off the happy path (request did not succeed) and return more descriptive errors to the client


jkilpatr:
please fill out a description and --amend your commit with a description.

PR #126, opened 2018-06-12
switch althea types to reexporting parity/primitives

kingoflolz:
serialization format remains the same, should not be a breaking change


rita/src/rita_exit/db_client/mod.rs
```diff
@@ -89,6 +89,7 @@ fn get_country(ip: &IpAddr) -> Result<String, Error> {
 }
 
 #[test]
+#[ignore]
```
jkilpatr: is this sneaking in from another change?

rita/src/rita_common/payment_controller/mod.rs
```diff
@@ -312,15 +312,15 @@ mod tests {
...
             .create();
 
         // mock bounty hunter
         let __m = mock("POST", "/update")
             .with_status(200)
             .with_body("bounty OK")
             .match_body("{\"from\":{\"mesh_ip\":\"1:1:1:1:1:1:1:1\",\
-            \"eth_address\":\"0x0101010101010101010101010101010101010101\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"balance\":\"-1\",\"tx\":{\"to\":{\"mesh_ip\":\"1:1:1:1:1:1:1:1\",\"eth_address\":\"0x0101010101010101010101010101010101010101\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"from\":{\"mesh_ip\":\"1:1:1:1:1:1:1:1\",\"eth_address\":\"0x0101010101010101010101010101010101010101\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"amount\":\"0x1\"}}")
+            \"eth_address\":\"0x0000000000000000000000000000000000000001\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"balance\":\"-1\",\"tx\":{\"to\":{\"mesh_ip\":\"1:1:1:1:1:1:1:1\",\"eth_address\":\"0x0000000000000000000000000000000000000001\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"from\":{\"mesh_ip\":\"1:1:1:1:1:1:1:1\",\"eth_address\":\"0x0000000000000000000000000000000000000001\",\"wg_public_key\":\"AAAAAAAAAAAAAAAAAAAA\"},\"amount\":\"0x1\"}}")
```
jkilpatr: why are we changing all of these?

kingoflolz: its easier to initialize to a low constant value as you can just do like 1.into(), instead of our previous array initialization approach

rita/src/rita_exit/db_client/mod.rs
kingoflolz: I was playing with the tests and this broke on me because my internet was a bit flaky, which made me realize that we shouldn't run this by default

kingoflolz:
Confirmed not to break backwards compat with master

kingoflolz:
seems to somehow break backwards compat, investigating

jkilpatr:
@kingoflolz has this been resolved?

settings/src/lib.rs
```diff
@@ -637,10 +639,12 @@ impl FileWrite for RitaSettingsStruct {
     fn write(&self, file_name: &str) -> Result<(), Error> {
         let ser = toml::Value::try_from(self.clone())?;
         let ser = toml::to_string(&ser)?;
-        let mut file = File::create(file_name)?;
-        file.write_all(ser.as_bytes())?;
-        file.flush().unwrap();
-
+        {
+            let mut file = File::create(file_name)?;
```
jkilpatr: what do the brackets achieve here?

settings/src/lib.rs
```diff
@@ -672,9 +678,13 @@ impl FileWrite for RitaExitSettingsStruct {
     fn write(&self, file_name: &str) -> Result<(), Error> {
         let ser = toml::Value::try_from(self.clone())?;
         let ser = toml::to_string(&ser)?;
-        let mut file = File::create(file_name)?;
-        file.write_all(ser.as_bytes())?;
-        file.flush().unwrap();
+        {
```
jkilpatr: is there an easy way to combine these duplicate functions?

settings/src/lib.rs
kingoflolz: the brackets make sure we close the file handle because it goes out of scope, before we do the sync command

PR #131, opened 2018-06-19
Correct default exit ip range

jkilpatr:
So the 172 range is 16 not 168 for the private section. We're lucky
this doesn't seem to break many sites because the block is assigned
to AOL and presumably used for Dialup users.

This also expands the assignment range to the maximum allowed, which
is more than a million people on a single exit server, not somthing
I ever expect to see.

### DIFF
--- a/settings/src/lib.rs
+++ b/settings/src/lib.rs
@@ -291,9 +291,9 @@ impl Default for ExitNetworkSettings {
             exit_hello_port: 4875,
             wg_tunnel_port: 59999,
             exit_price: 10,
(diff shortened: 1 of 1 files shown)


jkilpatr:
@kingoflolz for this I think we'll need a 'reset' sign up state than the exit can send such that people have to be assigned new details. Otherwise switching the ip range is impossible.

kingoflolz:
if we are ok with making users sign up again when the server changes stuff around significantly, we can just drop the users table and make everyone sign up again

PR #132, opened 2018-06-20
B side of Alpha 4 changes

jkilpatr:
This removes things slated to be removed in this release.

### DIFF
--- a/rita/src/rita_common/network_endpoints/mod.rs
+++ b/rita/src/rita_common/network_endpoints/mod.rs
@@ -8,8 +8,6 @@ use futures::Future;
 use settings::RitaCommonSettings;
 use SETTING;
 
-use serde_json;
-
 use rita_common;
 use rita_common::payment_controller::PaymentController;
 use rita_common::tunnel_manager::{GetWgInterface, OpenTunnelListener, TunnelManager};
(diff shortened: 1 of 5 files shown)


rita/src/rita_common/network_endpoints/mod.rs
```diff
@@ -53,25 +53,20 @@ pub fn hello_response(
     their_id: Json<serde_json::Value>,
```
kingoflolz: we can change this to Json<Identity>

rita/src/rita_common/network_endpoints/mod.rs
```diff
@@ -53,25 +53,20 @@ pub fn hello_response(
...
-    let new_id: Result<Identity, serde_json::Error> = serde_json::from_value(their_id.clone());
-    let old_id: Result<LocalIdentity, serde_json::Error> = serde_json::from_value(their_id.clone());
+    let id: Result<Identity, serde_json::Error> = serde_json::from_value(their_id.clone());
 
-    match new_id {
+    match id {
         Ok(new_id) => {
             info!("sending to new hello handler");
-            hello_response_new(new_id, req)
+            hello_response_handler(new_id, req)
```
kingoflolz: maybe we should inline this?

kingoflolz:
I think there is one thing I forgot to comment, try searching for `get_future()`, we need to make the future behavior the normal behavior

rita/src/rita_common/network_endpoints/mod.rs
jkilpatr: should we just remove hello_response in general and do the deserialization in the calling function?

jkilpatr:
not seeing anything else with get future.

jkilpatr:
backwards compat changes tested manually works well.

PR #133, opened 2018-06-20
Add exits endpoint for liveness

jkilpatr:
The new exit's endpoint is somewhat redundant compared to the settings endpoint
but provides liveness info like the reachability and tunnel state.


jkilpatr:
still needs testing

jkilpatr:
yes we do, I neglected that

PR #134, opened 2018-06-25
Exit signup refactor

kingoflolz:
lets see if travis gets it right this time


jkilpatr:
what's up with the tests here?

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -110,121 +108,243 @@ fn verify_identity(details: &ExitRegistrationDetails, request_ip: &IpAddr) -> Re
...
+        zip: client
+            .reg_details
+            .zip_code
+            .clone()
+            .unwrap_or("".to_string()),
+        country,
+    }
+}
+
+fn email_ver_done(_mesh_ip: &IpAddr, _conn: &SqliteConnection) -> Result<bool, Error> {
```
jkilpatr: does this need to be finished, also where is sending the email itself?

jkilpatr: ok so we still need integration with fastmail / some mail service

kingoflolz: yeah, planning to do that in a new PR as it is pretty self contained (and this one is big enough as is)

PR #143, opened 2018-07-04
email verification

kingoflolz:


althea_types/src/interop.rs
```diff
@@ -50,6 +53,8 @@ pub enum ExitState {
     Pending {
         general_details: ExitDetails,
         message: String,
+        #[serde(default)]
```
jkilpatr: so we have the dashboard post this in?

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -259,15 +262,14 @@ impl Handler<Tick> for ExitManager {
                     ));
                 }
                 ExitState::Registering { .. }
-                | ExitState::Pending { .. }
                 | ExitState::GotInfo {
                     auto_register: true,
                     ..
                 } => {
-                    Arbiter::handle().spawn(exit_setup_request(k.clone()).then(move |res| {
+                    Arbiter::handle().spawn(exit_setup_request(k.clone(), None).then(move |res| {
```
jkilpatr: what's the arbiter doing here?

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -208,12 +228,48 @@ fn client_to_db_client(
...
+        // TODO: maybe have a proper templating engine
+        .text(format!("Your althea verification code is [{}]", client.email_code))
+        .build()
+        .unwrap();
+
+    if SETTING.get_mailer().unwrap().test {
+        let mut mailer = FileEmailTransport::new(&SETTING.get_mailer().unwrap().test_dir);
+        mailer.send(&email)?;
+    } else {
+        // TODO add serde to lettre
```
jkilpatr: what needs to happen here?

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -265,23 +338,22 @@ impl Handler<SetupClient> for DbClient {
 
                         let new_ip = incr_dummy(&conn)?;
 
-                        //TODO: Send email
-
                         let user_country = if SETTING.get_allowed_countries().is_empty() {
```
jkilpatr: should we mark this for removal in some version?

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -208,12 +228,48 @@ fn client_to_db_client(
...
+        email_code: format!("{:06}", rand_code),
+        verified: false,
     }
 }
 
-fn email_ver_done(_mesh_ip: &IpAddr, _conn: &SqliteConnection) -> Result<bool, Error> {
-    //TODO actually check db
-    Ok(true)
+fn email_ver_done(client: &models::Client) -> Result<bool, Error> {
+    Ok(client.verified || SETTING.get_mailer().is_none())
```
jkilpatr: so if we disable mailer instant verification, I guess that's a good way to handle it.

rita/src/rita_client/exit_manager/mod.rs
jkilpatr: why can't we use a more normal future?

rita/src/rita_exit/db_client/mod.rs
jkilpatr: I'm talking about the country code stuff. We don't send it in future versions right?

rita/src/rita_client/exit_manager/mod.rs
kingoflolz: Normally you will have to spawn your future into a tokio core, which you start yourself, but since we have a cute running anyways from actix, we may as well use it because it's inefficient to run 2 event loops when you can do one

rita/src/rita_exit/db_client/mod.rs
kingoflolz: Yeah good point, it should be removed fire alpha 6

integration-tests/rita.py
```diff
@@ -312,6 +313,33 @@ def register_to_exit(node):
...
+
+
+
+def read_email(node):
+    id = node.id
+    # TODO: this is O(n^2)
+    for mail in os.listdir("mail"):
+        with open(os.path.join("mail", mail)) as mail_text:
+            mail = json.load(mail_text)
+            if mail["envelope"]["forward_path"][0] == "{}@althea.com".format(id):
```
drozdziak1: `althea.com` is an existing domain name, apparently with a mailing system set up. I don't think we want to necessarily ever risk spamming someone's server. `example.com` OTOH is specifically designated not to be used for anything other than testing.

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -3,6 +3,8 @@ use actix_web::*;
 
 use futures::Future;
 
+use failure::Error;
```
drozdziak1: Is this really necessary? This is the only change in the file.

kingoflolz: yes, its a result of a related change which borked type inference

integration-tests/rita.py
kingoflolz: ok, will change

integration-tests/rita.py
```diff
@@ -125,7 +125,8 @@ def exec_or_exit(command, blocking=True, delay=0.01):
 
 
 def cleanup():
-    os.system("rm -rf *.log *.pid private-key*")
+    os.system("rm -rf *.log *.pid private-key* mail")
+    os.system("mkdir mail")
```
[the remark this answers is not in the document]
kingoflolz: we delete it in the previous line (also this makes sure it operates with a clean checkout)

althea_types/src/interop.rs
```diff
@@ -30,6 +30,9 @@ pub struct ExitRegistrationDetails {
 
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub email: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub email_code: Option<String>,
```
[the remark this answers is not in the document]
kingoflolz: yes, and also the flow is 2 stage, we submit a setup request without email code and them submit one with a email code to get access

integration-tests/rita.py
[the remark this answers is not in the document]
drozdziak1: I can see that, but why not just have `rm -rf [...] mail/*` in there?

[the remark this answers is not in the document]
kingoflolz: we need to somehow create the mail dir if its a clean checkout, as its not committed into git

PR #145, opened 2018-07-10
Allow WAN toggle during operation

jkilpatr:
Previously per hop tunnels for gateways would only operate properly
when you started Rita with the machine in a totally ready state.
Now per hop gateways are reactive and will open at any time when
a device gets a default route on the external_nic. This should
solve several hard to debug issues.


jkilpatr:
rustfmt needs fixing, everything else looks good.

PR #146, opened 2018-07-11
Clean up verbose code

janus:
Cleaned up some codes inside althea_kernel_interface folder .

### DIFF
--- a/althea_kernel_interface/src/ip_route.rs
+++ b/althea_kernel_interface/src/ip_route.rs
@@ -6,15 +6,12 @@ use failure::Error;
 
 impl KernelInterface {
     fn get_default_route(&self) -> Option<Vec<String>> {
-        let output = self
-            .run_command("ip", &["route", "list", "default"])
+        let output = self.run_command("ip", &["route", "list", "default"])
             .unwrap();
 
         let stdout = String::from_utf8(output.stdout).unwrap();
 
         let mut def_route = Vec::new();
-
-        let mut found = false;
 
         // find all lines
         for i in stdout.lines() {
@@ -25,16 +22,11 @@ impl KernelInterface {
                         def_route.push(String::from(j));
                     }
                 }
-                found = true;
-                break;
+                return Some(def_route);
             }
         }
 
-        if found {
-            Some(def_route)
-        } else {
-            None
-        }
+        None
     }
 
     fn set_route(&self, to: &IpAddr, route: &Vec<String>) -> Result<(), Error> {
@@ -53,9 +45,7 @@ impl KernelInterface {
     }
 
     fn set_default_route(&self, route: &Vec<String>) -> Result<(), Error> {
-        let mut def_route_ref: Vec<&str> = vec!["route", "add"];
-
-        def_route_ref.push("default");
+        let mut def_route_ref: Vec<&str> = vec!["route", "add", "default"];
 
         for i in 1..route.len() {
             def_route_ref.push(route[i].as_str())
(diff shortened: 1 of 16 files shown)


althea_kernel_interface/src/ip_route.rs
```diff
@@ -6,15 +6,14 @@ use failure::Error;
...
-            .run_command("ip", &["route", "list", "default"])
+        let output = self.run_command("ip", &["route", "list", "default"])
             .unwrap();
 
         let stdout = String::from_utf8(output.stdout).unwrap();
 
         let mut def_route = Vec::new();
 
-        let mut found = false;
+        //let mut found = false;
```
kingoflolz: delete the code if it is not needed

althea_kernel_interface/src/ip_route.rs
```diff
@@ -53,9 +46,9 @@ impl KernelInterface {
     }
 
     fn set_default_route(&self, route: &Vec<String>) -> Result<(), Error> {
-        let mut def_route_ref: Vec<&str> = vec!["route", "add"];
+        let mut def_route_ref: Vec<&str> = vec!["route", "add", "default"];
 
-        def_route_ref.push("default");
+        //def_route_ref.push("default");
```
kingoflolz: same here

althea_kernel_interface/src/lib.rs
```diff
@@ -100,6 +100,7 @@ impl CommandRunner for LinuxCommandRunner {
     }
 
     fn set_mock(&self, _mock: Box<FnMut(String, Vec<String>) -> Result<Output, Error> + Send>) {
+        println!("unimplemented!()");
```
kingoflolz: Don't think this is needed, the unimplemented macro tells us that it was hit

althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -9,7 +9,7 @@ fn to_wg_local(ip: &IpAddr) -> IpAddr {
     match ip {
         &IpAddr::V6(ip) => {
             let mut seg = ip.segments();
-            assert_eq!((seg[0] & 0xfd00), 0xfd00);
+            assert_eq!((seg[0] & 0xffff), 0xfd00);
```
kingoflolz: we only want to check the first 9 bits, not the whole first 16

althea_kernel_interface/src/ip_route.rs
jkilpatr: I like this change you can just delete this line though, no need to leave it commented out.

althea_kernel_interface/src/stats.rs
```diff
@@ -5,62 +5,51 @@ use std::fs::File;
 use std::io::Read;
 
 impl KernelInterface {
-    pub fn get_proc_stat(&self) -> Result<String, Error> {
-        debug!("getting proc stat");
-        let mut f = File::open("/proc/stat")?;
-
+    fn get_stat(&self, path_str: &str) -> Result<String, Error> {
```
kingoflolz: maybe this should be read_file ?

althea_kernel_interface/src/ip_route.rs
jkilpatr: same thing here.

althea_kernel_interface/src/lib.rs
jkilpatr: not needed the unimplemented!() should print something like that for us.

jkilpatr:
looks good, I'll let @kingoflolz review again then I'll merge

drozdziak1:
Yup @janus, like @kingoflolz said you need to run cargo-fmt on your code. Make sure it's up-to-date with `cargo install rustfmt-nightly --force` and then do `cargo fmt` inside the project and then commit and push.

jkilpatr:
https://github.com/althea-mesh/althea_rs/pull/150

cleaned up here, will merge once tested there.

PR #154, opened 2018-07-24
Add a POST /exits endpoints to the dashboard

drozdziak1:


rita/src/rita_client/dashboard/mod.rs
```diff
@@ -330,3 +331,48 @@ impl Handler<GetExitInfo> for Dashboard {
...
+    type Result = Result<(), Error>;
+}
+
+impl Handler<SelectOrResetExit> for Dashboard {
+    type Result = Result<(), Error>;
+
+    fn handle(&mut self, msg: SelectOrResetExit, _ctx: &mut Self::Context) -> Self::Result {
+        debug!("POST /exits received request: {:?}", msg);
+
+        if msg.reset {
```
kingoflolz: what if its neither reset nor select? that should be an error right?

PR #155, opened 2018-07-26
Updated integration test runner

nathande:
Need to specify bash as the shell intepreter because dash is the default on Ubuntu. Also removed kv_store crate as nothing referenced it.


integration-tests/rita.py
```diff
@@ -42,7 +42,7 @@
 COMPAT_LAYOUT = os.getenv('COMPAT_LAYOUT', None)
 
 BACKOFF_FACTOR = float(os.getenv('BACKOFF_FACTOR', 1))
-CONVERGENCE_DELAY = float(os.getenv('CONVERGENCE_DELAY', 50))
+CONVERGENCE_DELAY = float(os.getenv('CONVERGENCE_DELAY', 100))
```
jkilpatr: We don't need to do this, as this value is carefully tuned to make the CI tests run faster. If you want it to be slower for you you can set the environmental var CONVERGENCE_DELAY=100 ./rita.sh

integration-tests/rita.sh
```diff
@@ -74,4 +75,4 @@ else
   popd
 fi
 
-sudo -E python3 rita.py $@
+sudo -E PATH="$PATH:$HOME/.cargo/bin" python3 rita.py $@
```
jkilpatr: did you not add cargo to your own path? That should fix this problem too. I guess maybe we shouldn't assume people did that.

integration-tests/readme.md
```diff
@@ -7,14 +7,13 @@ althea-mesh/
...
 
-The tests need to be run on ubuntu 16.04 as root (`sudo su`).
-
 Network lab needs to be installed using `bpkg`.
 
 Example:
 ```
-sudo su
-bpkg install sudomesh/network-lab
+sudo apt-get install -y libsqlite3-dev iperf3 python3-pip bridge-utils wireguard linux-source linux-headers-$(uname -r) curl git libssl-dev pkg-config build-essential ipset
```
jkilpatr: this should tell you to go and install packages for your platform

integration-tests/rita.sh
drozdziak1: @jkilpatr There were some hiccups with PATH passing in `sudo -E` on Travis during the A/B in CI ordeal, perhaps @nathande faced the same problem. I solved it for CI then by adding `cp <cargo_bin>/diesel /usr/bin/diesel` as a CI prep step. It is a bit problematic though because `~/.cargo/bin` being the cargo installation path is only a default and assuming that will be annoying to people with custom setups later on. On CI however we have control over that so the copying is sane there.

drozdziak1: @nathande Maybe make this into an env like `CARGO_BIN_PATH` defaulting to `~/.cargo/bin` which gets appended to `PATH` so that people with different setups can change that easily?

jkilpatr:
@nathande why make a new commit why not ammend the old one? Keeps the commit history cleaner

jkilpatr:
Looks good to me, will merge after tests

PR #156, opened 2018-07-27
Explicit peer discovery

jkilpatr:
This creates a new module PeerListener that is responsible for finding peers and then integrates it into tunnel manager. 

The end result is a pretty large refactor and a significant replacement of shelling out with pure rust code. This already seems faster and more stable than the previous version. 

What it does still need is some tests for both tunnel manager and peer listener and some more IRL testing. This should be backwards compatible.


rita/src/exit.rs
```diff
@@ -176,6 +178,7 @@ fn main() {
     assert!(rita_common::tunnel_manager::TunnelManager::from_registry().connected());
     assert!(rita_common::http_client::HTTPClient::from_registry().connected());
     assert!(rita_common::traffic_watcher::TrafficWatcher::from_registry().connected());
+    assert!(PeerListener::new().is_ok());
```
kingoflolz: What does this do? Looks like it just creates a peer listener, check if it succeeds and then throws it away?

rita/src/rita_common/network_endpoints/mod.rs
```diff
@@ -44,52 +47,56 @@ pub fn make_payments(
...
-    let old_id: Result<LocalIdentity, serde_json::Error> = serde_json::from_value(req.0.clone());
-
-    // remove in Alpha 6
-    let their_id = match new_id {
-        Ok(new_id) => {
-            info!("got new id sending new response");
-            new_id
+    let id: Result<LocalIdentity, serde_json::Error> = serde_json::from_value(req.0.clone());
+
+    let their_id = match id {
```
kingoflolz: We should just be able to make req take a Json<LocalIdentity> instead of a serde value, should make this code cleaner

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+    buf.put_u16_be(22);
+    let ipaddr_bytes: [u8; 16] = addr.octets();
+    for i in 0..16 {
+        buf.put_u8(ipaddr_bytes[i]);
+    }
+    trace!("Encoded packet {:x?}", buf);
+    buf
+}
+
+fn decode_im_here(buf: &mut Vec<u8>) -> Result<Option<Ipv6Addr>, io::Error> {
```
kingoflolz: Why do you return a Resul<Option<>> here? I think it would be better to just return a Result<Ipv6Addr, failure::Error>

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+}
+
+fn recieve_imhere(interfaces: &mut Vec<ListenInterface>) -> Result<Vec<Peer>, Error> {
+    trace!("About to dequeue ImHere");
+    let mut output = Vec::new();
+    for listen_interface in interfaces.iter_mut() {
+        // Since the only datagrams we are interested in are very small (22 bytes plus overhead)
+        // this buffer is kept intentionally small to discard larger packets earlier rather than later
+        let mut datagram: [u8; 100] = [0; 100];
+        let res = listen_interface.multicast_socket.recv_from(&mut datagram);
```
kingoflolz: Shouldn't we keep on receiving packets until there is none left in the buffer?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -88,30 +91,23 @@ impl Default for TunnelManager {
...
-    type Result = ();
-
-    fn handle(&mut self, listen: Listen, _: &mut Context<Self>) -> Self::Result {
-        self.listen_interfaces.insert(listen.0);
-        SETTING.get_network_mut().peer_interfaces = self.listen_interfaces.clone();
+    fn handle(&mut self, msg: IdentityCallback, _: &mut Context<Self>) -> Self::Result {
+        let local_id = msg.0;
+        let peer = msg.1;
+        let res = self.open_tunnel(local_id, peer);
+        if res.is_err() {
```
kingoflolz: I think the pattern below can be applied to this section and many other sections of code, so you don't do unwraps, even if you know they will work (makes for easier grepping of unwraps to find potential failure points)
```
match res {
    Ok(res) => {
        Some(res)
    }
    Err(e) => {
        warn!("Open Tunnel failed with {:?}", e);
    }
}
```

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,306 +150,167 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+            }
+        }
+        for manual_peer in SETTING.get_network().manual_peers.iter() {
+            let ip = manual_peer.parse::<IpAddr>();
+            let port = SETTING.get_network().rita_hello_port;
+            if ip.is_ok() {
+                let ip = ip.unwrap();
+                let socket = SocketAddr::new(ip, port);
+                let man_peer = Peer {
+                    linklocal_ip: ip, // not actually linklocal this time
```
kingoflolz: maybe contact_ip is a better name? or maybe the LLIP is actually redundant when we have contact_socket

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,306 +150,167 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-        if self.tunnel_map.contains_key(&ip) {
-            trace!("found existing wg interface for {}", ip);
-            self.tunnel_map[&ip].clone()
-        } else {
-            trace!("creating new wg interface for {}", ip);
-            let new = self.new_if();
-            self.tunnel_map.insert(ip.clone(), new.clone());
-            new
+        TunnelManager {
+            ports: ports,
```
kingoflolz: maybe this should be called free ports?

settings/src/lib.rs
```diff
@@ -72,6 +72,8 @@ pub struct NetworkSettings {
     pub own_ip: IpAddr,
     /// Mesh IP of bounty hunter (in fd00::/8)
     pub bounty_ip: IpAddr,
+    /// Broadcast ip address used for peer discovery (in ff02::/8)
+    pub discovery_ip: Ipv6Addr,
```
kingoflolz: we need a default function and serde attribute if we want to remain config compatible

rita/src/exit.rs
jkilpatr: previously new did a lot, now we need the same stuff as the other actors I guess.

rita/src/rita_common/network_endpoints/mod.rs
jkilpatr: so like Json<Hello>?

rita/src/rita_common/peer_listener/mod.rs
jkilpatr: because it's possible everything is ok but we didn't find a proper packet

settings/src/lib.rs
jkilpatr: done

rita/src/rita_common/network_endpoints/mod.rs
jkilpatr: sorry not really seeing the syntax for this. Don't we have to parse?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -460,635 +387,7 @@ impl TunnelManager {
...
-
-        let sys = System::new("test");
-
-        System::current().registry().set(
-            HTTPClient::mock(Box::new(|msg, _ctx| {
-                assert_eq!(
-                    msg.downcast_ref::<Hello>(),
-                    Some(&Hello {
-                        my_id: SETTING.get_identity(),
-                        to: SocketAddr::V4(SocketAddrV4::new("1.1.1.1".parse().unwrap(), 4876))
```
jkilpatr: this is the sort of test we want to readd

rita/src/rita_common/network_endpoints/mod.rs
kingoflolz: do something like  req: (Json<LocalIdentity>, HttpRequest)

althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -17,7 +17,7 @@ impl KernelInterface {
         let cap = re.captures(&str);
         if let Some(cap) = cap {
             trace!("got link local IP of {} from device {}", &cap[1], &dev);
-            Ok(cap[1].parse()?)
+            Ok(cap[1].parse::<Ipv6Addr>()?)
```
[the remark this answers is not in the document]
jkilpatr: it would make an IpAddr which could be ipv4 which would be wrong.

rita/src/client.rs
```diff
@@ -189,10 +188,10 @@ fn main() {
...
         })
     }).workers(1)
-    .bind(format!("[::0]:{}", SETTING.get_network().rita_contact_port))
-    .unwrap()
-    .shutdown_timeout(0)
-    .start();
+        .bind(format!("[::0]:{}", SETTING.get_network().rita_contact_port))
+        .unwrap()
+        .shutdown_timeout(0)
+        .start();
```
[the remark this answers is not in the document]
jkilpatr: sure easy.

rita/src/rita_client/dashboard/mod.rs
```diff
@@ -124,23 +125,23 @@ impl Handler<SetWifiConfig> for Dashboard {
             let iface_number = i.section_name.clone().chars().last();
 
             if i.mesh && iface_number.is_some() {
-                let iface_name = format!("rita_wlan{}", iface_number.unwrap());
+                let iface_name = format!("wlan{}", iface_number.unwrap());
```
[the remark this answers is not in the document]
jkilpatr: intended, the rita_wlan dummy objects aren't really needed and actually break new peer discovery, they where nice for some legacy firewall stuff in the firmware we don't even use anymore.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +211,201 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-    }
+    Ok(())
 }
 
 impl TunnelManager {
     pub fn new() -> Self {
+        let start = SETTING.get_network().wg_start_port;
+        let mut ports = Vec::<u16>::new();
+        for i in start..65500 {
+            ports.push(i);
```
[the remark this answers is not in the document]
jkilpatr: this is a good point, if we can manage to implement it correctly.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -38,33 +37,75 @@ type Resolver = Mocker<resolver::Resolver>;
...
-    DNSLookupError,
+/* Uncomment when tunnel state handling is added
+#[derive(Debug, Clone)]
+pub enum TunnelState {
+    Init,
+    Open,
+    Throttled,
+    Closed,
 }
+*/
```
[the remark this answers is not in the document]
jkilpatr: because we will want it before the end of the month when we implement tunnel closing based on blockchain stuff.

jkilpatr:
So I'm 80% sure that the optimal route failure just a parsing problem @drozdziak1. Backwards compat takes the same topology and puts a rubber stamp on it. if you could investigate I would appreciate it.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+            multicast_socketaddr: multicast_socketaddr,
+            linklocal_ip: link_ip,
+        })
+    }
+}
+
+fn encode_im_here(addr: Ipv6Addr) -> Vec<u8> {
+    let mut buf = Vec::new();
+    buf.put_u32_be(MSG_IM_HERE);
+    buf.put_u16_be(22);
```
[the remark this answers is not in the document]
mpapierski: We'll never exceed MTU so usually it would be below ~1500bytes at all times. If anything I'd just shrink magic key to 16 bits to save 2 bytes. In theory magic key could be even `u8` as we won't exceed 255 distinct message types. But at least for now it could be like this as well.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+            Err(e) => {
+                error!("Receiving ImHere failed with {:?}", e);
+            }
+        }
+
+        Ok(())
+    }
+}
+
+pub struct Listen(pub String);
```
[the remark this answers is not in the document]
mpapierski: +1 TBH I'd suggest naming fields in structs at all times. The code will be more readable with explicit names instead of `.0`, `.1` etc.

rita/src/rita_common/rita_loop/mod.rs
```diff
@@ -100,32 +108,48 @@ impl Handler<Tick> for RitaLoop {
...
                         })
                 }),
         );
+
+        trace!("Starting PeerListener tick");
+        Arbiter::spawn(
+            PeerListener::from_registry()
+                .send(Tick {})
+                .then(|res| {
+                    trace!("PeerListener said {:?}", res);
```
drozdziak1: Mention here that whatever it said was after Tick

rita/src/rita_common/rita_loop/mod.rs
```diff
@@ -100,32 +108,48 @@ impl Handler<Tick> for RitaLoop {
...
+                    res
+                }).then(|_| Ok(())),
+        );
+
+        trace!("Getting Peers from PeerListener to pass to TunnelManager");
+        Arbiter::spawn(
+            PeerListener::from_registry()
+                .send(GetPeers {})
+                .and_then(|peers| {
+                    trace!("Got peers from PeerListener, passing to TunnelManager");
```
drozdziak1: Mention what the peers are.

rita/src/rita_common/rita_loop/mod.rs
```diff
@@ -100,32 +108,48 @@ impl Handler<Tick> for RitaLoop {
...
+                    TunnelManager::from_registry().send(PeersToContact(peers.unwrap())) // GetPeers never fails so unwrap is safe
```
drozdziak1: If it never fails then why does it need a `Result`? (unless it's forced by actix?)

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -38,33 +44,75 @@ type Resolver = Mocker<resolver::Resolver>;
...
+
+        //let init = TunnelState::Init;
+        let tunnel = Tunnel {
+            ip: ip,                       //tunnel endpoint
+            iface_name: iface_name,       //name of wg#
+            listen_ifidx: ifidx,          //The physical port this tunnel is listening on
+            listen_port: our_listen_port, //the port this tunnel resides on
+            //            tunnel_state: init,           //how this tunnel feels about it's life
+            localid: their_id.clone(), // the identity of the counterparty tunnel once we have it
+        };
```
drozdziak1: Thanks for commenting! Could you please move those to the struct *definition* above? That's where people curious about `Tunnel` will look.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -38,33 +44,75 @@ type Resolver = Mocker<resolver::Resolver>;
...
+
+        let network = SETTING.get_network().clone();
+
+        // TODO you have the iface index use it
```
drozdziak1: If you don't think this is a good moment for implementing this -> file an issue. We should centralize PM things and this is one of them IMO

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -38,33 +44,75 @@ type Resolver = Mocker<resolver::Resolver>;
...
     }
 }
 
 pub struct TunnelManager {
-    pub port: u16,
-
-    tunnel_map: HashMap<IpAddr, TunnelData>,
-    listen_interfaces: HashSet<String>,
+    free_ports: Vec<u16>,
+    tunnels: Vec<Tunnel>,
```
drozdziak1: We most likely want sets for these

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-}
-
 pub struct GetNeighbors;
 impl Message for GetNeighbors {
     type Result = Result<Vec<(LocalIdentity, String, IpAddr)>, Error>;
 }
 
 impl Handler<GetNeighbors> for TunnelManager {
-    type Result = ResponseFuture<Vec<(LocalIdentity, String, IpAddr)>, Error>;
+    type Result = Result<Vec<(LocalIdentity, String, IpAddr)>, Error>;
```
drozdziak1: Again, probably we want a set

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-}
-
-impl Handler<GetLocalIdentity> for TunnelManager {
-    type Result = MessageResult<GetLocalIdentity>;
-
-    fn handle(&mut self, their_id: GetLocalIdentity, _: &mut Context<Self>) -> Self::Result {
-        MessageResult(self.get_local_identity(their_id.from))
+        let mut res = Vec::new();
+        for tunnel in self.tunnels.iter() {
+            res.push((tunnel.localid.clone(), tunnel.iface_name.clone(), tunnel.ip));
```
drozdziak1: Might wanna `map().collect()` here instead, which could make the handler leaner. I understand the paing of tracking long method chains, but that's not really the case here.

rita/src/rita_common/tunnel_manager/mod.rs
drozdziak1: Also, isn't a cloned tunnels set/vec what we could just return? A `memcpy` would be quicker than the iteration below, and a lock on a reference would be even quicker that.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+        }
+        Ok(res)
     }
 }
 
-pub struct OpenTunnel(pub LocalIdentity, pub IpAddr);
+pub struct PeersToContact(pub Vec<Peer>);
```
drozdziak1: Document, please!

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
     type Result = ();
-
-    fn handle(&mut self, their_id: OpenTunnel, _: &mut Context<Self>) -> Self::Result {
-        self.open_tunnel(their_id.0, their_id.1).unwrap();
-        ()
+    fn handle(&mut self, peers: PeersToContact, _ctx: &mut Context<Self>) -> Self::Result {
+        trace!("TunnelManager contacting peers");
+        for peer in peers.0.iter() {
+            let res = self.neighbor_inquiry(peer.clone());
+            if res.is_err() {
```
drozdziak1: You want an `if let Err(e) here. Also, what's the use of knowing it's an error without saying what the error is?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+                warn!("Neighbor inqury for {:?} failed!", peer);
+            }
+        }
+        // Do not contact manual peers if we are not a gateway
+        if SETTING.get_network().is_gateway {
+            for manual_peer in SETTING.get_network().manual_peers.iter() {
+                let ip = manual_peer.parse::<IpAddr>();
+                let port = SETTING.get_network().rita_hello_port;
+                if ip.is_ok() {
```
drozdziak1: You want an `if let Ok(ip) = ip.unwrap()` here

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+                    let ip = ip.unwrap();
+                    let socket = SocketAddr::new(ip, port);
+                    let man_peer = Peer {
+                        contact_ip: ip,
+                        ifidx: 0,
+                        contact_socket: socket,
+                    };
+                    let res = self.neighbor_inquiry(man_peer);
+                    if res.is_err() {
```
drozdziak1: You want `if let`

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+                        warn!("Neighbor inqury for {:?} failed!", manual_peer);
+                    }
+                } else {
+                    let res = self.neighbor_inquiry_hostname(manual_peer.to_string());
+                    if res.is_err() {
```
drozdziak1: same

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+                        warn!("Neighbor inqury for {:?} failed!", manual_peer);
```
drozdziak1: Again why notify of error if we don't get to know what the error is.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-impl Message for OpenTunnelListener {
-    type Result = ();
-}
+/// Sets out to contact a neighbor, takes a speculative port (only assigned if the neighbor
+/// responds successfully)
+fn contact_neighbor(peer: Peer, our_port: u16) -> Result<(), Error> {
+    KI.manual_peers_route(
+        &peer.contact_ip,
+        &mut SETTING.get_network_mut().default_route,
+    ).unwrap();
```
drozdziak1: You probably want `?`, unless you can guarantee this doesn't fail.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-        ()
-    }
+    Ok(())
 }
 
 impl TunnelManager {
     pub fn new() -> Self {
+        let start = SETTING.get_network().wg_start_port;
+        // TODO we can eliminate the possiblitly of port leaks and save 256kb of ram by
+        // going through and using the operating system to keep track of free ports
```
drozdziak1: Issue please

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+        let mut ports = Vec::<u16>::new();
+        for i in start..65500 {
```
drozdziak1: Go all the way to `65535`. Also, you can probably just go `let mut ports = (start..65500).collect()`. As mentioned in other parts that touch `free_ports`, the ports most likely should be a set. It would suck big time to iterate this many times to know that a port is not taken.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
-            self.tunnel_map[&ip].clone()
-        } else {
-            trace!("creating new wg interface for {}", ip);
-            let new = self.new_if();
-            self.tunnel_map.insert(ip.clone(), new.clone());
-            new
+    pub fn get_port(&mut self) -> u16 {
+        let port = self.free_ports.pop();
+        if port.is_none() {
+            panic!("Tunnelmanager ran out of ports!")
```
drozdziak1: This sucks big time. We wanna just refuse to mesh with them here. In the future in dense areas we'll probably want neighbor limiting.

drozdziak1: I suggest you pass on the `Option` from `pop()` and have the caller decide what to do.

drozdziak1: Or better yet get rid of this function and just inline that where necessary.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+        let our_port = self.get_port();
+
+        let res = Resolver::from_registry()
+            .send(resolver::Resolve::host(their_hostname.clone()))
+            .then(move |res| match res {
+                Ok(Ok(dnsresult)) => {
+                    let port = SETTING.get_network().rita_hello_port;
+                    let url = format!("http://[{}]:{}/hello", their_hostname, port);
+                    info!("Saying hello to: {:?} at ip {:?}", url, dnsresult);
+                    if dnsresult.len() > 0 && SETTING.get_network().is_gateway {
```
drozdziak1: What if we're not a gateway and the DNS request hangs? Several seconds of wasted time

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -154,941 +218,203 @@ impl Handler<GetPhyIpFromMeshIp> for TunnelManager {
...
+    pub fn open_tunnel(
+        &mut self,
+        their_id: LocalIdentity,
+        peer: Peer,
+        our_port: u16,
+    ) -> Result<Tunnel, Error> {
+        trace!("TunnelManager getting existing tunnel or opening a new one");
+        for tunnel in self.tunnels.iter() {
+            // This is deceptively simple, see the commit message
+            // TODO delete ifidx zero tunnels once we start deleting tunnels at all
```
drozdziak1: Issue please

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+pub struct PeerListener {
+    interfaces: Vec<ListenInterface>,
+    peers: Vec<Peer>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Peer {
+    pub contact_ip: IpAddr,
+    pub ifidx: u32,
+    pub contact_socket: SocketAddr,
```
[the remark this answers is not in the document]
jkilpatr: fixed.

rita/src/rita_common/peer_listener/mod.rs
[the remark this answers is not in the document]
jkilpatr: done

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+                    .insert(new_iface_name);
+            }
+            Err(e) => {
+                error!("Peer listener failed to listen on {:?}", e);
+            }
+        }
+    }
+}
+
+pub struct UnListen(pub String);
```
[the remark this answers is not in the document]
jkilpatr: done

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+                .get_network_mut()
+                .peer_interfaces
+                .remove(&ifname_to_delete);
+        } else {
+            error!("Peer listener failed to unlisten on {:?}", ifname_to_delete);
+        }
+    }
+}
+
+pub struct GetListen;
```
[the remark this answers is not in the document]
jkilpatr: done.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+impl Message for GetListen {
+    type Result = Result<HashSet<String>, Error>;
+}
+
+impl Handler<GetListen> for PeerListener {
+    type Result = Result<HashSet<String>, Error>;
+    fn handle(&mut self, _: GetListen, _: &mut Context<Self>) -> Self::Result {
+        let mut res = HashSet::new();
+        for item in self.interfaces.iter() {
```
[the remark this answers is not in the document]
jkilpatr: turns out we don't even need this, deleted.

rita/src/rita_common/peer_listener/mod.rs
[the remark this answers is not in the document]
jkilpatr: I made the magic long to make attempts to parse other protocols messages less probable, but even being on the right port and ip is already vanishingly unlikely, I'll go for u8

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+use std::io;
+use std::io::Cursor;
+use std::net::{IpAddr, Ipv6Addr, SocketAddr, SocketAddrV6, UdpSocket};
+
+use rita_common::rita_loop::Tick;
+
+use KI;
+use SETTING;
+
+pub const MSG_IM_HERE: u32 = 0x5b6d4158;
```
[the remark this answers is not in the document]
jkilpatr: shortened down to 5b, should still be random enough to prevent collisions with other protocols

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+    if packet_magic != MSG_IM_HERE {
+        trace!(
+            "Recieved an ImHere packet with an invalid magic: {:?}",
+            packet_magic
+        );
+        return Ok(None);
+    }
+
+    let packet_size = pointer.read_u16::<BigEndian>()?;
+    if packet_size < 22 as u16 {
```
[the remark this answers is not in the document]
jkilpatr: agree on the global len, I'm trying to make sure we have enough bytes to get the ipaddr but also leave room for us to tack stuff on in the future without breaking parsing.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+    let peer_address = Ipv6Addr::new(
+        peer_address_arr[7],
+        peer_address_arr[6],
+        peer_address_arr[5],
+        peer_address_arr[4],
+        peer_address_arr[3],
+        peer_address_arr[2],
+        peer_address_arr[1],
+        peer_address_arr[0],
+    );
```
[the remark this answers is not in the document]
jkilpatr: There's no ergonomic way to get it to reverse the order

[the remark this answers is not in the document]
jkilpatr: well we could do reverse iter then collect but that's actually less efficient.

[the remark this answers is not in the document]
jkilpatr: I'll leave this up to you.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+}
+
+fn recieve_imhere(interfaces: &mut Vec<ListenInterface>) -> Result<Vec<Peer>, Error> {
+    trace!("About to dequeue ImHere");
+    let mut output = Vec::<Peer>::new();
+    for listen_interface in interfaces.iter_mut() {
+        let mut socket_empty = false;
+        // Since the only datagrams we are interested in are very small (22 bytes plus overhead)
+        // this buffer is kept intentionally small to discard larger packets earlier rather than later
+        while !socket_empty {
```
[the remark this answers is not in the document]
jkilpatr: this is multi tlv proof, we get all the messages we want off of the socket. then move on with live.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+            let mut datagram: [u8; 100] = [0; 100];
+            let res = listen_interface.multicast_socket.recv_from(&mut datagram);
+            if res.is_err() {
+                trace!("Could not recv ImHere");
+                // TODO Consider we might want to remove interfaces that produce specific types
+                // of errors from the active list
+                socket_empty = true;
```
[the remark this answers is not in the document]
jkilpatr: we discard it and wait to get a full copy.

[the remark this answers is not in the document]
jkilpatr: we only pull full messages off of this buffer as far as I can tell,

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+
+fn recieve_imhere(interfaces: &mut Vec<ListenInterface>) -> Result<Vec<Peer>, Error> {
+    trace!("About to dequeue ImHere");
+    let mut output = Vec::<Peer>::new();
+    for listen_interface in interfaces.iter_mut() {
+        let mut socket_empty = false;
+        // Since the only datagrams we are interested in are very small (22 bytes plus overhead)
+        // this buffer is kept intentionally small to discard larger packets earlier rather than later
+        while !socket_empty {
+            let mut datagram: [u8; 100] = [0; 100];
```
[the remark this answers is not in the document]
jkilpatr: like the parsing we want to be able to parse some extra info if we tack it on in the future.

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+            if res.is_none() {
+                trace!("ImHere decode was unsuccessful!");
+                continue;
+            }
+            let ipaddr = res.unwrap();
+
+            if ipaddr == listen_interface.linklocal_ip {
+                trace!("Got ImHere from myself");
+                continue;
+            }
```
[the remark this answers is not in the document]
jkilpatr: nesting matches gets messy IMO, and you need more than two.

rita/src/rita_common/rita_loop/mod.rs
jkilpatr: forced by actix

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: https://github.com/althea-mesh/althea_rs/issues/176

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: don't see the advantage for port, we never search and never check membership. No duplicates maybe,

jkilpatr: nah hashset doesn't make sense for ports, no push or pop makes getting a port (we don't care which one) difficult.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: this is a point of integration with traffic watcher which is IMO out of scope to modify with this pull.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: we don't ever have to do that!, we just pop a port off of the list.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: https://github.com/althea-mesh/althea_rs/issues/177

rita/src/rita_common/peer_listener/mod.rs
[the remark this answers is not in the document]
drozdziak1: Can't it be sent in reverse then?

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+            "Recieved an ImHere packet with an invalid size: {:?}",
+            packet_size
+        );
+        return Ok(None);
+    }
+
+    let mut peer_address_arr: [u16; 8] = [0xFFFF; 8];
+    for i in (0..8).rev() {
+        peer_address_arr[i] = pointer.read_u16::<BigEndian>()?;
+    }
```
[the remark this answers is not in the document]
drozdziak1: Ah yeah you want the endianness, fair enough.

rita/src/rita_common/peer_listener/mod.rs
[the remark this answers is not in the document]
drozdziak1: OK (talked over in appear)

rita/src/rita_common/peer_listener/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+        peer_address_arr[1],
+        peer_address_arr[0],
+    );
+
+    if peer_address.is_unspecified() || peer_address.is_loopback() || peer_address.is_multicast() {
+        trace!(
+            "Recieved a valid ImHere with an invalid ip address: {:?}",
+            peer_address,
+        );
+        return Ok(None);
```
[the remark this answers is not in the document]
drozdziak1: Conclusion from Appear: Go with either: use a Result and handle/ignore the error externally

rita/src/rita_common/peer_listener/mod.rs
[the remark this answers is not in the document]
drozdziak1: REsolved, thanks!

rita/src/rita_common/tunnel_manager/mod.rs
drozdziak1: Conclusion from appear: this will be handled in a refactor issue @jkilpatr made

rita/src/rita_common/tunnel_manager/mod.rs
drozdziak1: This is thought out exactly right! Big reviewer trap though. The delay is neligible thanks to Actix' and futures' shenanigans

jkilpatr:
@drozdziak1 one last review then merge, hopefully.

rita/src/rita_common/peer_listener/message.rs
```diff
@@ -0,0 +1,208 @@
...
+                    );
+                    return Err(MessageError::InvalidIpAddress);
+                }
+
+                trace!("ImHere decoding completed successfully {:?}", peer_address);
+                Ok(PeerMessage::ImHere(peer_address))
+            }
+            _ => {
+                trace!(
+                    "Recieved an ImHere packet with an invalid magic: {:X?}",
```
drozdziak1: What makes it ImHere specifically if the magic is wrong? Say that it's just an unknown magic

rita/src/rita_common/peer_listener/message.rs
```diff
@@ -0,0 +1,208 @@
...
+                }
+                trace!("Encoded ImHere packet {:x?}", buf);
+                return buf;
+            }
+        }
+    }
+    /**
+     * Decode buffer of data into a ImHere message
+     * Message format is very simple
+     * Magic <u8>, Size <u16>, Ipaddr <u128>
```
drozdziak1: The u128 plan didn't work right? Can you say that it's in fact a `&[u16; 8]`?

jkilpatr:
@drozdziak1 been testing this on real routers and exits + integration tests locally all day, it's rock solid as far as I can tell. We also really need it in to send out testing builds by Friday (aka Thursday night)

PR #157, opened 2018-07-31
Removed unused zip and luci_pass fields from database

nathande:


integration-tests/rita.py
```diff
@@ -341,7 +339,7 @@ def read_email(node):
             mail = json.load(mail_file_handle)
             if mail["envelope"]["forward_path"][0] == "{}@example.com".format(id):
                 return ''.join(chr(i) for i in mail["message"])
-    print("cannot find email for node {}".format(id))
+    raise Exception("cannot find email for node {}".format(id))
```
drozdziak1: Why use an exception? How is it any better than `teardown()` and then `sys.exit(1)`?

exit_db/migrations/2018-07-30-160806_remove_unused/up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN TRANSACTION;
+CREATE TEMPORARY TABLE clients_backup(mesh_ip, wg_pubkey, wg_port, internal_ip, email, country);
```
drozdziak1: You're ommitting several fields here, take a look at the previous migration, the ones you're adding here are not the complete list minus {zip,luci_pass}

PR #158, opened 2018-07-31
(WIP) Pre-release fixes

drozdziak1:
* We no longer include debug symbols in release builds
* Tunnel manager now unwraps the interface name before formating it into a URL
* The unused zip and luci_pass field migration now has all the other fields intact (Cherry-picked over and amended @nathande's commit making #157 obsolete)
* Cargo.lock lists rita at latest release tag
* The new /exits endpoints now have docs


integration-tests/rita.py
```diff
@@ -341,7 +339,7 @@ def read_email(node):
             mail = json.load(mail_file_handle)
             if mail["envelope"]["forward_path"][0] == "{}@example.com".format(id):
                 return ''.join(chr(i) for i in mail["message"])
-    print("cannot find email for node {}".format(id))
+    raise Exception("cannot find email for node {}".format(id))
```
mpapierski: I remember from other PR that this change introduces a side effect. Is it intentional?

mpapierski:
Looks good, but Travis fails on `cargo test` for some reason:

```
failures:

---- rita_common::tunnel_manager::tests::test_neighbor_inquiry_domain stdout ----
	thread 'rita_common::tunnel_manager::tests::test_neighbor_inquiry_domain' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- rita_common::tunnel_manager::tests::test_neighbor_inquiry_ip stdout ----
	thread 'rita_common::tunnel_manager::tests::test_neighbor_inquiry_ip' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21


failures:
    rita_common::tunnel_manager::tests::test_neighbor_inquiry_domain
    rita_common::tunnel_manager::tests::test_neighbor_inquiry_ip
```

integration-tests/rita.py
drozdziak1: Failures should generally exit. I was previously using `sys.exit(1)` rightaway. Come to think of it an exception might be better for just running `teardown()` rightaway. I didn't pay that much attention because I have rita.py overhauled in a nother branch.

drozdziak1:
Yup, I've got that sorted locally, this PR is still WIP because of that.

drozdziak1:
Pushed

PR #159, opened 2018-08-01
Add ipq40xx arm target

jkilpatr:
Turns out the glb1300 is a little different than the Omnia, this accounts
for that.

### DIFF
--- a/scripts/openwrt_build_ipq40xx.sh
+++ b/scripts/openwrt_build_ipq40xx.sh
@@ -5,14 +5,14 @@ if [[ ! -d $DIR/staging_dir ]]; then
     wget -N https://updates.altheamesh.com/staging.tar.xz -O staging.tar.xz > /dev/null; tar -xf staging.tar.xz
 fi
 
-export TOOLCHAIN=toolchain-arm_cortex-a9+vfpv3_gcc-7.3.0_musl_eabi
+export TOOLCHAIN=toolchain-arm_cortex-a7+neon-vfpv4_gcc-7.3.0_musl_eabi
 export TARGET_CC=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-gcc
 export TARGET_LD=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-ld
 export TARGET_AR=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-ar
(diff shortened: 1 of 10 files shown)


scripts/openwrt_build_ipq40xx.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
drozdziak1: Could you in fact list that this file helps with the gl-b1300?

scripts/openwrt_build_mvebu.sh
```diff
@@ -9,12 +9,12 @@ export TOOLCHAIN=toolchain-arm_cortex-a9+vfpv3_gcc-7.3.0_musl_eabi
 export TARGET_CC=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-gcc
 export TARGET_LD=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-ld
 export TARGET_AR=$DIR/staging_dir/$TOOLCHAIN/bin/arm-openwrt-linux-ar
-export CARGO_TARGET_ARM_UNKNOWN_LINUX_MUSLEABIHF_LINKER=$TARGET_CC
-export CARGO_TARGET_ARM_UNKNOWN_LINUX_MUSLEABIHF_AR=$TARGET_AR
+export CARGO_TARGET_ARMV7_UNKNOWN_LINUX_MUSLEABIHF_LINKER=$TARGET_CC
```
drozdziak1: Is this a fix for the Illegal Instruction problem you mentioned on Riot?

scripts/openwrt_build_ipq40xx.sh
drozdziak1: Is there docs for these scripts?

mpapierski:
Looks good. I retried one build on integration tests that seemed to be stalled. Also, one build fails because it seems that we have misformatted code with `cargo +nightly fmt`. This could be addressed in a separate PR, though.

scripts/openwrt_build_ipq40xx.sh
jkilpatr: not as much as I would like.

scripts/openwrt_build_mvebu.sh
jkilpatr: well the new ipq40xx stuff is the fix for that, armv7 is just me realizing that rust would generate compatible but not as capable code

scripts/openwrt_build_ipq40xx.sh
drozdziak1: Looks great, thanks!

PR #161, opened 2018-08-02
Increase test timeout to 30 minutes.

mpapierski:
This increases timeout on tests from default 10 minutes, to 30 minutes.
This was observed at least once in #159, where integration tests build
failed because no output was received for 10 minutes.

More details: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Possible solution to #163


mpapierski:
I'm closing it for now as this wouldn't solve the possible root cause.

Issue #164, opened 2018-08-03
Release with rust stable


mpapierski:
Yesterday [Rust 1.28 was released](https://blog.rust-lang.org/2018/08/02/Rust-1.28.html). So far we use `nightly` because of `-z` flag that allows us to compress the binaries. According to @ttk2 it wasn't possible before to ditch `-z` and use `stable` because of no working binary compressor - UPX devel branch makes MIPS executables segfault at startup.

We need to verify that we can safely use `stable` for future releases.

There are some warnings as well because some stuff got deprecated.

mpapierski:
Solved in #167

PR #167, opened 2018-08-03
Fix warnings after moving to 1.28 stable.

mpapierski:
Enough said. Compiles without warnings on 1.28 stable. Part of effort on #164

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
   - which diesel || cargo install diesel_cli --no-default-features --features sqlite
   - sudo cp $(which diesel) /usr/bin
 env:
-  - TEST_COMMAND="cargo +nightly fmt --all -- --check"
+  - TEST_COMMAND="rustup component add rustfmt-preview && cargo +nightly fmt --all -- --check"
   - TEST_COMMAND="cargo build --verbose --all"
   - TEST_COMMAND="cargo test --verbose --all" RUST_TEST_THREADS=1
 rust:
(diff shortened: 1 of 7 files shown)


jkilpatr:
@mpapierski please go into the .travis.yml and make it stable sot hat we can have one pull for moving to stable.

mpapierski:
@jkilpatr good point. Added beta, nightly builds to `allowed_failure` section. Main point of builds is `stable` from now on. Reduced from 13 to 7 builds, the rest doesn't affect the stability of the build. Should improve verification speed.

.travis.yml
```diff
@@ -8,34 +8,26 @@ before_install:
...
-  - rust: stable
-    env: TEST_COMMAND="./integration-tests/rita.sh" REVISION_B=master COMPAT_LAYOUT="inner_ring_old" INITIAL_POLL_INTERVAL=5 BACKOFF_FACTOR="1.5" VERBOSE=1
-  - rust: beta
-    env: TEST_COMMAND="./integration-tests/rita.sh" REVISION_B=master COMPAT_LAYOUT="inner_ring_old" INITIAL_POLL_INTERVAL=5 BACKOFF_FACTOR="1.5" VERBOSE=1
+  allow_failures:
+    - rust: beta
+    - rust: nightly
+  include:
+    - name: "cargo fmt"
+      script: rustup component add rustfmt-preview --toolchain nightly && cargo fmt --all -- --check
```
jkilpatr: this should not be toolchain nightly

mpapierski: As discussed there seems to be a problem with `rustfmt` on stable. It produces a relatively big diff, which I wouldn't expect to happen. For some reason default set of settings is different on nightly, even after recent release of rust `1.28`. Lets keep it consistent at least for now. I'll just update this PR with `rustfmt.toml` that would hopefully make `rustfmt` deterministic in future. Also pre-commit hook would be nice.

PR #170, opened 2018-08-03
OpenWRT upload script should kill running Rita

jkilpatr:
This changes the upload script to use the most common device type (mips) and to automatically kill any running instances of Rita, allowing for easier use especially since the last instance in a debugging session usually remains stubbornly alive until you login to fix it.

### DIFF
--- a/scripts/openwrt_upload.sh
+++ b/scripts/openwrt_upload.sh
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 set -eux
-export TARGET=mipsel
-export TRIPLE=mipsel-unknown-linux-musl
+export TARGET=mips
+export TRIPLE=mips-unknown-linux-musl
 export ROUTER_IP=192.168.10.1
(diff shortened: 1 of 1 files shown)


scripts/openwrt_upload.sh
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 set -eux
-export TARGET=mipsel
-export TRIPLE=mipsel-unknown-linux-musl
+export TARGET=mips
+export TRIPLE=mips-unknown-linux-musl
 export ROUTER_IP=192.168.10.1
 bash scripts/openwrt_build_$TARGET.sh
+set +e
+ssh root@$ROUTER_IP killall -9 rita
```
drozdziak1: Why not just `|| exit 1`?

jkilpatr: we don't want to exit, we specifically want to continue.

drozdziak1: True, confused the sign

PR #171, opened 2018-08-03
Wifi endpoint upgrade

jkilpatr:
At first I tried a simple path based endpoint, but that was too restrictive
as you could not set a wifi SSID or password that contained spaces, at least
without careful escaping. Now you post very simple json objects as documented
in the docs to modify the wifi password, ssid or mesh toggle.

Testing and working, although the wifi mesh setting needs more testing to be really sure.


docs/api/router-dashboard.md
```diff
@@ -364,3 +364,76 @@ This file documents the dashboard API found in Rita client.
...
+
+```
+{}
+```
+
+- Error Response: `500 Server Error`
+
+- Sample Call:
+
+`curl -XPOST 127.0.0.1:<rita_dashboard_port>/settings -H 'Content-Type: application/json' -i -d '{"radio":"wlan0", "ssid": "this is a freeform ssid"}'`
```
asoltys: The curl example here is posting to /settings rather than /wifi_settings/ssid

docs/api/router-dashboard.md
```diff
@@ -364,3 +364,76 @@ This file documents the dashboard API found in Rita client.
...
+
+```
+{}
+```
+
+- Error Response: `500 Server Error`
+
+- Sample Call:
+
+`curl -XPOST 127.0.0.1:<rita_dashboard_port>/settings -H 'Content-Type: application/json' -i -d '{"radio":"wlan0", "pass": "this is a freeform password"}'`
```
asoltys: Same here

docs/api/router-dashboard.md
```diff
@@ -364,3 +364,76 @@ This file documents the dashboard API found in Rita client.
...
+
+```
+{}
+```
+
+- Error Response: `500 Server Error`
+
+- Sample Call:
+
+`curl -XPOST 127.0.0.1:<rita_dashboard_port>/settings -H 'Content-Type: application/json' -i -d '{"radio":"wlan0", "mesh": true}'`
```
asoltys: And here

PR #172, opened 2018-08-06
Exit registration endpoints

drozdziak1:
This PR adds two `/exits/{name}/register` and `/exits/{name}/verify/{code}` endpoints for handling both stages of exit registration. It supersedes the `auto_register` settings option.

Stuff still wrong/missing:
* [x] A `wait()` on the exit_setup_request() future seems to be hogging a resource for all "/exits"-suite requests.
* [x] Docs
* [x] `auto_register` and `email_code` still exist despite not being used - **A different PR will be open for this (issue #181)
* [x] Give explicit error reasons for failures and indicate appropriately with the status code

### DIFF
--- a/integration-tests/rita.py
+++ b/integration-tests/rita.py
@@ -312,24 +312,20 @@ def switch_binaries(node_id):
             "BOUNTY_HUNTER:\t{}").format(RITA, RITA_EXIT, BOUNTY_HUNTER))
 
 def register_to_exit(node):
-    id = node.id
-    os.system("ip netns exec netlab-{id} curl -XPOST 127.0.0.1:4877/settings -H 'Content-Type: application/json' -i -d '{data}'"
-              .format(id=id, data=json.dumps({"exit_client": EXIT_SELECT})))
+    os.system(("ip netns exec netlab-{} curl -XPOST " +
+        "127.0.0.1:4877/exits/exit_a/register").format(node.id))
 
 def email_verif(node):
-    id = node.id
-
     email_text = read_email(node)
 
     code = re.search(r"\[([0-9]+)\]", email_text).group(1)
 
-    print(code)
+    print("Email code for node {} is {}".format(node.id, code))
 
-    os.system("ip netns exec netlab-{id} curl -XPOST 127.0.0.1:4877/settings -H 'Content-Type: application/json' -i -d '{data}'"
-              .format(id=id, data=json.dumps({"exit_client": {"exits": {"exit_a": {"email_code": code}}}})))
-
-    os.system("ip netns exec netlab-{id} curl 127.0.0.1:4877/settings"
-              .format(id=id))
+    exec_or_exit(("ip netns exec netlab-{} curl -XPOST " +
+        "127.0.0.1:4877/exits/exit_a/verify/{}").format(node.id, code))
+    exec_or_exit(("ip netns exec netlab-{} curl " +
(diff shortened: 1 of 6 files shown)


integration-tests/rita.py
```diff
@@ -312,24 +312,20 @@ def switch_binaries(node_id):
             "BOUNTY_HUNTER:\t{}").format(RITA, RITA_EXIT, BOUNTY_HUNTER))
 
 def register_to_exit(node):
-    id = node.id
-    os.system("ip netns exec netlab-{id} curl -XPOST 127.0.0.1:4877/settings -H 'Content-Type: application/json' -i -d '{data}'"
-              .format(id=id, data=json.dumps({"exit_client": EXIT_SELECT})))
+    os.system(("ip netns exec netlab-{} curl -XPOST " +
```
mpapierski: Its safer to use `subprocess.call` or `subprocess.check_output` when you need to gather stdout.

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -54,24 +58,102 @@ pub fn get_exit_info(_req: HttpRequest) -> Box<Future<Item = Json<Vec<ExitInfo>>
...
+    Box::new(exit_setup_request(exit_name, None).then(|res| {
+        let mut ret = HashMap::new();
+        match res {
+            Ok(_) => future::ok(HttpResponse::Ok().json(ret)),
+            Err(e) => {
+                error!("exit_setup_request() failed with: {:?}", e);
+                ret.insert("error".to_owned(), "Exit setup request failed".to_owned());
+                ret.insert("rust_error".to_owned(), format!("{:?}", e));
+                future::ok(
+                    HttpResponse::new(StatusCode::BAD_REQUEST)
```
mpapierski: Woudln't server error (500+) code fit more here? `BAD_REQUEST` is a client error which would imply bad user input or something similiar, and we clearly handle an error here.

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -54,24 +58,102 @@ pub fn get_exit_info(_req: HttpRequest) -> Box<Future<Item = Json<Vec<ExitInfo>>
...
+    Box::new(exit_setup_request(exit_name, Some(code)).then(|res| {
+        let mut ret = HashMap::new();
+        match res {
+            Ok(_) => future::ok(HttpResponse::Ok().json(ret)),
+            Err(e) => {
+                error!("exit_setup_request() failed with: {:?}", e);
+                ret.insert("error".to_owned(), "Exit setup request failed".to_owned());
+                ret.insert("rust_error".to_owned(), format!("{:?}", e));
+                future::ok(
+                    HttpResponse::new(StatusCode::BAD_REQUEST)
```
mpapierski: Same as above - we handle error condition here so perhaps 500+ error would be better?

rita/src/rita_client/dashboard/network_endpoints.rs
drozdziak1: Yeah but problem is exit_setup_request only checks if the exit exists inside, so I have no way of knowing from the outside, and checking again outside is kinda janky.

rita/src/rita_client/dashboard/network_endpoints.rs
drozdziak1: same response

PR #179, opened 2018-08-07
Catch `ValueError` regardless of py version.

mpapierski:
A small adhoc change. 

It turns out on travis sometimes we run rita.py on py2, where JSONDecodeError doesn't exist.

Example failed job: https://travis-ci.org/althea-mesh/althea_rs/jobs/413146414

### DIFF
--- a/integration-tests/rita.py
+++ b/integration-tests/rita.py
@@ -686,16 +686,16 @@ class World:
                 + "netlab-{} curl -sfg6 [::1]:4877/neighbors".format(node.id)),
                 stdout=subprocess.PIPE)
             assert_test(not result.wait(), "curl-ing /neighbors")
+            stdout = result.stdout.read().decode('utf-8')
             try:
                 print("Received neighbors:")
                 if VERBOSE:
-                    neighbors = json.loads(result.stdout.read().decode('utf-8'))
+                    neighbors = json.loads(stdout)
                     pprint(neighbors)
                 else:
-                    print(result.stdout.read().decode('utf-8'))
+                    print(stdout)
             except ValueError as e:
-                if VERBOSE:
-                    print('Unable to decode JSON {!r}: {!r}'.format(result, e))
+                print('Unable to decode JSON {!r}: {}'.format(stdout, e))
                 assert_test(False, "Decoding the neighbors JSON")
 
             # /exits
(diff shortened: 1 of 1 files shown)


integration-tests/rita.py
```diff
@@ -693,7 +693,9 @@ def test_endpoints_all(self):
                     pprint(neighbors)
                 else:
                     print(result.stdout.read().decode('utf-8'))
-            except json.JSONDecodeError as e:
+            except ValueError as e:
+                if VERBOSE:
+                    print('Unable to decode JSON {!r}: {!r}'.format(result, e))
```
drozdziak1: Unnecessary, the `assert_test` below will print `<red plus> Decoding the JSON Failed`

integration-tests/rita.py
```diff
@@ -711,7 +713,9 @@ def test_endpoints_all(self):
                     pprint(exits)
                 else:
                     print(result.stdout.read().decode('utf-8'))
-            except json.JSONDecodeError as e:
+            except ValueError as e:
+                if VERBOSE:
+                    print('Unable to decode JSON {!r}: {!r}'.format(result, e))
```
drozdziak1: same here

integration-tests/rita.py
```diff
@@ -729,7 +733,9 @@ def test_endpoints_all(self):
                     pprint(info)
                 else:
                     print(result.stdout.read().decode('utf-8'))
-            except json.JSONDecodeError as e:
+            except ValueError as e:
+                if VERBOSE:
+                    print('Unable to decode JSON {!r}: {!r}'.format(result, e))
```
drozdziak1: here

integration-tests/rita.py
```diff
@@ -747,7 +753,9 @@ def test_endpoints_all(self):
                     pprint(settings)
                 else:
                     print(result.stdout.read().decode('utf-8'))
-            except json.JSONDecodeError as e:
+            except ValueError as e:
+                if VERBOSE:
+                    print('Unable to decode JSON {!r}: {!r}'.format(result, e))
```
drozdziak1: and here

mpapierski:
@drozdziak1 As discussed I removed `if VERBOSE` branch and it will report invalid responses unconditionally. Updated with some more changes because because I realized I'll be reporting `Popen` object instead of actual `stdout` output.

PR #182, opened 2018-08-08
Add /wipe endpoint

### LINKED ISSUE #173: Add a /wipe for dev builds
It could be really handy for development to add a `/wipe` HTTP endpoint that would wipe possible state in rita:

- Remove `wg` keys
- Remove `eth` keys

The most important thing to make sure is that this endpoint has to be hidden in release builds. An integration test is welcome that would verify release build doesn't have this endpoint.

It seems perfectly reasonable to add 'dangerous' endpoints during development to speed development.

mpapierski:
This adds `/wipe` endpoint to dashboard server. It removes WG interface, and recreates keys both in settings & on disk. This should fix #173.

- [x] 404 based on `--release`, wipes data on `--debug`.
- [x] WireGuard keys in settings
- [x] Mesh IP address regenerated
- [x] WG Priv key on disk recreated
- [x] ~Eth address recreate~ before guac integration which would handle Eth addresses we can ship what we have for now. Opinions?

### DIFF
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -222,6 +222,7 @@ fn main() {
                 verify_on_exit_with_code,
             ).route("/info", Method::GET, get_own_info)
             .route("/version", Method::GET, version)
+            .route("/wipe", Method::POST, wipe)
     }).workers(1)
     .bind(format!(
         "[::0]:{}",
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -156,6 +156,12 @@ fn main() {
     openssl_probe::init_ssl_cert_env_vars();
     env_logger::init();
 
+    if cfg!(feature = "development") {
+        println!("Warning!");
+        println!("This build is meant only for development purposes.");
+        println!("Running this on production is unsupported and not safe!");
+    }
+
     let args: Args = Docopt::new((*USAGE).as_str())
         .and_then(|d| d.deserialize())
         .unwrap_or_else(|e| e.exit());
--- a/rita/src/exit.rs
+++ b/rita/src/exit.rs
@@ -229,6 +229,7 @@ fn main() {
             .route("/settings", Method::POST, set_settings)
             .route("/version", Method::GET, version)
             .route("/wipe", Method::POST, wipe)
+            .route("/database", Method::DELETE, nuke_db)
     }).bind(format!(
         "[::0]:{}",
         SETTING.get_network().rita_dashboard_port
--- a/rita/src/exit.rs
+++ b/rita/src/exit.rs
@@ -228,6 +228,7 @@ fn main() {
             .route("/settings", Method::GET, get_settings)
             .route("/settings", Method::POST, set_settings)
             .route("/version", Method::GET, version)
+            .route("/wipe", Method::POST, wipe)
     }).bind(format!(
         "[::0]:{}",
         SETTING.get_network().rita_dashboard_port
--- a/rita/src/exit.rs
+++ b/rita/src/exit.rs
@@ -156,6 +156,12 @@ fn main() {
     openssl_probe::init_ssl_cert_env_vars();
     env_logger::init();
 
+    if cfg!(feature = "development") {
+        println!("Warning!");
+        println!("This build is meant only for development purposes.");
+        println!("Running this on production as an exit node is unsupported and not safe!");
+    }
+
     let args: Args = Docopt::new((*USAGE).as_str())
         .and_then(|d| d.deserialize())
         .unwrap_or_else(|e| e.exit());
--- a/rita/src/rita_common/dashboard/network_endpoints.rs
+++ b/rita/src/rita_common/dashboard/network_endpoints.rs
@@ -54,9 +54,12 @@ pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
 pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
     // Clean up existing WG interfaces
     match cleanup() {
-        Ok(_) => trace!("wipe: Cleanup success!"),
+        Ok(_) => trace!("wipe: WireGuard interfaces cleanup success!"),
         Err(e) => {
-            warn!("wipe: Unable to complete cleanup: {:?}", e);
+            warn!(
+                "wipe: Unable to complete WireGuard interfaces cleanup: {:?}",
+                e
+            );
             return Err(e);
         }
     }
--- a/rita/src/rita_common/dashboard/network_endpoints.rs
+++ b/rita/src/rita_common/dashboard/network_endpoints.rs
@@ -54,35 +54,35 @@ pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
 pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
     // Clean up existing WG interfaces
     match cleanup() {
-        Ok(_) => trace!("Cleanup success!"),
+        Ok(_) => trace!("wipe: Cleanup success!"),
         Err(e) => {
-            warn!("Unable to complete cleanup: {:?}", e);
+            warn!("wipe: Unable to complete cleanup: {:?}", e);
             return Err(e);
         }
     }
 
     // Restore default route
     match KI.restore_default_route(&mut SETTING.get_network_mut().default_route) {
-        Ok(_) => trace!("Restore default route success!"),
+        Ok(_) => trace!("wipe: Restore default route success!"),
         Err(e) => {
-            warn!("Unable to restore default route: {:?}", e);
+            warn!("wipe: Unable to restore default route: {:?}", e);
             return Err(e);
         }
     }
 
     // Create new WireGuard keys
     match linux_generate_wg_keys(&mut SETTING.get_network_mut()) {
-        Ok(_) => trace!("Generated new WireGuard keys"),
+        Ok(_) => trace!("wipe: Generated new WireGuard keys"),
         Err(e) => {
-            warn!("Unable to generate new WireGuard keys: {:?}", e);
+            warn!("wipe: Unable to generate new WireGuard keys: {:?}", e);
             return Err(e);
         }
     }
     // Generate new mesh IP
     match linux_generate_mesh_ip(&mut SETTING.get_network_mut()) {
-        Ok(_) => trace!("Generated new mesh IP"),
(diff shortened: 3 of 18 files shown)


rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
 
     JsonStatusResponse::new(Ok("New settings applied".to_string()))
 }
+
+pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
+    if cfg!(not(debug_assertions)) {
```
jkilpatr: what does this do exactly?

rita/src/client.rs
```diff
@@ -210,6 +210,7 @@ fn main() {
             .route("/wifi_settings/mesh", Method::POST, set_wifi_mesh)
             .route("/info", Method::GET, get_own_info)
             .route("/version", Method::GET, version)
+            .route("/wipe", Method::POST, wipe)
```
drozdziak1: Make the method DELETE, it better fits the purpose and requires an extra step which should help ensure that the person making the request truly knows what they want.

rita/src/exit.rs
```diff
@@ -223,6 +223,7 @@ fn main() {
             .route("/settings", Method::GET, get_settings)
             .route("/settings", Method::POST, set_settings)
             .route("/version", Method::GET, version)
+            .route("/wipe", Method::POST, wipe)
```
drozdziak1: Same here. BTW did you test that this isn't visible to machines on the outside? Imagine someone developing on `borked` and suddenly someone starts messing with them by wiping the exit.

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+        return Ok(HttpResponse::NotFound().finish());
+    }
+
+    // Clean up existing WG interfaces
+    match cleanup() {
+        Ok(_) => trace!("Cleanup success!"),
+        Err(e) => {
+            warn!("Unable to complete cleanup: {:?}", e);
+            return Err(e);
```
drozdziak1: Cleanups shouldn't interrupt on errors but just move on with other stuff on their agenda. You should want to discard this error.

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+        }
+    }
+
+    // Restore default route
+    match KI.restore_default_route(&mut SETTING.get_network_mut().default_route) {
+        Ok(_) => trace!("Restore default route success!"),
+        Err(e) => {
+            warn!("Unable to restore default route: {:?}", e);
+            return Err(e);
```
drozdziak1: Same here

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+        }
+    }
+
+    // Create new WireGuard keys
+    match linux_generate_wg_keys(&mut SETTING.get_network_mut()) {
+        Ok(_) => trace!("Generated new WireGuard keys"),
+        Err(e) => {
+            warn!("Unable to generate new WireGuard keys: {:?}", e);
+            return Err(e);
```
drozdziak1: same here

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+        }
+    }
+    // Generate new mesh IP
+    match linux_generate_mesh_ip(&mut SETTING.get_network_mut()) {
+        Ok(_) => trace!("Generated new mesh IP"),
+        Err(e) => {
+            warn!("Unable to generate new mesh IP: {:?}", e);
+            return Err(e);
```
drozdziak1: Same here

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+
+    // Creates file on disk containing key
+    match KI.create_wg_key(
+        &Path::new(&SETTING.get_network().wg_private_key_path),
+        &SETTING.get_network().wg_private_key,
+    ) {
+        Ok(_) => trace!("Generated new WireGuard keys"),
+        Err(e) => {
+            warn!("Unable to generate new WireGuard keys: {:?}", e);
+            return Err(e);
```
drozdziak1: Same here

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -38,3 +43,58 @@ pub fn set_settings(
...
+pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
+    if cfg!(not(debug_assertions)) {
+        return Ok(HttpResponse::NotFound().finish());
+    }
+
+    // Clean up existing WG interfaces
+    match cleanup() {
+        Ok(_) => trace!("Cleanup success!"),
+        Err(e) => {
+            warn!("Unable to complete cleanup: {:?}", e);
```
drozdziak1: Unable to complete cleanup of *what*?

drozdziak1: Also, prepend each message with `wipe: ` so that people know where this happens, as the logs only show the origin of a message up to module level.

rita/src/exit.rs
jkilpatr: that's currently a firewall thing the installer should *not* open the port these endpoints reside on. Now that's not the best way to handle things long term but it works for now. 

In general unauthenticaed endpoints are an unavoidable problem on clients, but on exits they should probably filter for localhost connections only or force auth.

rita/src/rita_common/dashboard/network_endpoints.rs
jkilpatr: maybe you want a stack trace?

rita/src/rita_common/dashboard/network_endpoints.rs
mpapierski: `debug_assertions` are true for `--debug` build and false for `--release`. Couldn't find better way than this. I'll add comment to this part.

rita/src/rita_common/dashboard/network_endpoints.rs
drozdziak1: I mean, we don't want anyone to confuse the warning with something truly alarming.

rita/src/client.rs
mpapierski: Hmm... Didn't think of it. Seems like good fit. Speaking of semantics that would be more appropriate to call `DELETE /wireguard`...

drozdziak1:
Also, docs?

rita/src/rita_common/dashboard/network_endpoints.rs
drozdziak1: @jkilpatr A way to do both is what we need IMHO. However @mpapierski don't Rust errors stringify along with the backtrace?

rita/src/exit.rs
mpapierski: Its on the same address as dashboards: `[::0]:$PORT` - isn't `::0` all addresses so probably all dashboard endpoints are possibly visible outside?

jkilpatr: we format in the port, but we bind to all addresses but block the port with the firewall (or should, I'll check today)

rita/src/rita_common/dashboard/network_endpoints.rs
mpapierski: I'm not convinced on ignoring errors, though. I'd rather return error 500 at any point that would indicate that you'd have to do some action manually, or perhaps retry a HTTP request later from client side.

rita/src/rita_common/dashboard/network_endpoints.rs
mpapierski: This cleanup is called at the beggining of client/exit which wipes wg interfaces. Good point on logging.

rita/src/rita_common/dashboard/network_endpoints.rs
```diff
@@ -89,9 +95,9 @@ pub fn wipe(_req: HttpRequest) -> Result<HttpResponse, Error> {
         &Path::new(&SETTING.get_network().wg_private_key_path),
         &SETTING.get_network().wg_private_key,
     ) {
-        Ok(_) => trace!("Generated new WireGuard keys"),
+        Ok(_) => trace!("wipe: Generated new WireGuard keys"),
```
drozdziak1: This should be called writing the key to disk, not generating it.

rita/src/rita_common/dashboard/network_endpoints.rs
drozdziak1: @mpapierski Requiring developers to do stuff manually on a failed step of cleanup defeats the purpose of said cleanup or even purpose as a development enhancement. I say we return a JSON like:
```json
{
"wireguard_cleanup": "OK",
"restore_default_route": "OK",
"new_wg_key": "Failed",
"wg_key_write_to_disk": "Failed",
"new_mesh_ip": "OK",
}
```

asoltys:
I tried the /wipe endpoint and it reset my key properly and blew away my tunnels but the tunnels didn't get recreated. I didn't try the DELETE /database endpoint yet since I haven't updated the server. I'm wondering, is there going to be an endpoint on the client for that to cause the client to hit the server endpoint?

I had to comment out the #[cfg(debug_assertions)] to get /wipe working because I was getting 404 responses at first even though I built with debug = true

rita/src/exit.rs
jkilpatr: should we just bind to 192.168.10.1? or whatever the configured lan is? but that's out of scope for this pull.

scripts/build_common.sh
```diff
@@ -0,0 +1,35 @@
+# Builds release profile unless specified other
+PROFILE="--release"
+# Features switch that will be passed to cargo in a docker container
+FEATURES=""
```
jkilpatr: an example usage would be nice, I'm not quite sure why this is here.

mpapierski: I'll add some help. It is meant to parse arguments and return variables such as $PROFILE and $FEATURES so it could be used in `cargo build` calls. It is `source`-ed from build scripts.

PR #187, opened 2018-08-09
Use openssl-probe crate to fix cert errors.

### LINKED ISSUE #183: StarTLS with Lettre fails when using MUSL static builds
https://gist.github.com/jkilpatr/c69d57f7bd84cb65847a5ec7e065ec52

mpapierski:
Reproduced the issue with `linux_build_static.sh` and using a simple `reqwest` HTTPS call.

This should fix #183.

### DIFF
--- a/rita/src/client.rs
+++ b/rita/src/client.rs
@@ -79,6 +79,8 @@ use rita_client::dashboard::network_endpoints::*;
 use rita_common::dashboard::network_endpoints::*;
 use rita_common::network_endpoints::*;
 
+extern crate openssl_probe;
+
 #[derive(Debug, Deserialize)]
 struct Args {
     flag_config: String,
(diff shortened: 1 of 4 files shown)


rita/src/client.rs
```diff
@@ -77,6 +77,8 @@ use rita_client::dashboard::network_endpoints::*;
 use rita_common::dashboard::network_endpoints::*;
 use rita_common::network_endpoints::*;
 
+extern crate openssl_probe;
```
drozdziak1: Put this along with other `extern crate`s. Our imports look like garbage in many places, but let's not make it worse.

rita/src/exit.rs
```diff
@@ -81,6 +81,7 @@ use std::sync::{Arc, RwLock};
 #[cfg(test)]
 use std::sync::Mutex;
 
+extern crate openssl_probe;
```
drozdziak1: same here

mpapierski: Makes sense.

PR #189, opened 2018-08-09
Read /etc/resolv.conf for dns routes

jkilpatr:
For some reason TrustDNS does not seem to like non 8.8.8.8 servers, but this does the routing components correctly.


rita/src/rita_common/rita_loop/mod.rs
```diff
@@ -86,14 +83,17 @@ impl Handler<Tick> for RitaLoop {
...
-            #[cfg(not(test))]
-            for i in ResolverConfig::default().name_servers() {
-                trace!("TrustDNS default {:?}", i);
-                KI.manual_peers_route(
-                    &i.socket_addr.ip(),
-                    &mut SETTING.get_network_mut().default_route,
-                ).unwrap();
+
+            let servers = KI.get_resolv_servers();
+            match servers {
```
mpapierski: Interesting that according to TrustDNS docs it supports /etc/resolv.conf. Anyway I'm +1 on this if it solves a problem.

althea_kernel_interface/src/dns.rs
```diff
@@ -0,0 +1,34 @@
+use super::KernelInterface;
+
+use std::fs::File;
+use std::io::Error;
+use std::io::Read;
+use std::net::IpAddr;
+
+impl KernelInterface {
+    /// Get's a list of IP addresses of nameservers from /etc/resolv.conf, may be v6, v4 or both
```
drozdziak1: s/get's/gets/g

drozdziak1: s/get's/gets/g

althea_kernel_interface/src/dns.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// generally ignores malformed lines but produces IO errors
+    pub fn get_resolv_servers(&self) -> Result<Vec<IpAddr>, Error> {
+        let mut f = File::open("/etc/resolv.conf")?;
+        let mut contents = String::new();
+        f.read_to_string(&mut contents)?;
+
+        let mut res = Vec::new();
+        for line in contents.lines() {
```
drozdziak1: Can you know for a fact there isn't gonna be comments in the file before the nameserver entries? I think a regex might fit here better.

rita/src/rita_common/rita_loop/mod.rs
drozdziak1: The `servers` variable has no other use than this match, could you inline it?

althea_kernel_interface/src/dns.rs
jkilpatr: I've never seen nameserver #comment ip but I guess the format actually allows it. 

The issue is that we would want a pretty complex ipv4 or v6 regex.

drozdziak1: It's good, just got confused, sorry!

PR #190, opened 2018-08-09
Fix unwrap in dashboard

jkilpatr:
Sometimes the info endpoint crashed, I can't reproduce it after doing this. Could never get a proper stack trace out of it before.


mpapierski:
I'm landing it

PR #194, opened 2018-08-10
WIP: Add build script for Pi3

jkilpatr:
This one requires a special workaround in the rust linker args, but is otherwise fine.


jkilpatr:
doesn't work for some reason.

jkilpatr:
dropping in favour of just building for the pi as a firmware target .

Issue #197, opened 2018-08-10
Tunnel State disagreement


jkilpatr:
When we open a per hop tunnel we perform a very simple consensus protocol. Each party exchanges some details and waits for a response, if they get a response they consider it to be an approval of the details they sent and settle on a state representative of the details sent and recieved. 

When one side is rebooted and the other is not this result in a consensus issue, one side thinks the process has already completed and the other things no such negotiation has taken place. My current proposal is to include a 'new' flag in such scenarios so that we can recreate tunnels as needed.

PR #198, opened 2018-08-10
Resolve tunnel state disagreement

jkilpatr:
Before the Peer discovery refactor we would always mutate the existing
tunnel during a successful identity exchange. Now we maintain a tunnel
state and only make modifications when required.

This works fine until one side reboots and the other does not, now there
is disagreement over what the tunnel state is. We need to introduce a single
bit of extra state. We can't do this by not sending hellos when tunnels are
already open. Because in NAT punching cases this prevents tunnel opening totally.

Accepting that we need constant hello sending as a liveness measure we need a way
for one side to indicate to the other that a problem exits. So we send a bool have_tunnel
with our identity, indicating that we do or do not have a tunnel with the peer already.

To complicate matters we don't actually know when we send our hello if we already have
a tunnel, because we send hello's to IP's but open tunnels with identities. To resolve
this bootstrapping problem the bool is wrapped in an option.

So the total system has two bools wrapped in options, that's effectively 16 possible
states. The possible states are reduced to 6 due to these constraints.

Hello responses always include Some(bool) since the hello has an identity we can use for the
lookup.

When a None is recieved we interpret it as 'True' so that we don't delete a tunnel that
might be perfectly fine

Therefore there is only one state in which we take corrective action based on this bool, which is when we have a tunnel but the hello response said they do not. In that case we delete our current tunnel
and create a new one matching the local identity sent in the message that's currently being
processed.

With this system no matter what state any side is in one side is garunteed to either be in
the right state already or take action to correct the tunnel state.

Even in the case that the option is somehow eliminated in the future everything will work, because only the state (true, false) results in a tunnel reset.

### DIFF
--- a/althea_types/src/interop.rs
+++ b/althea_types/src/interop.rs
@@ -146,6 +146,7 @@ impl Message for Identity {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct LocalIdentity {
     pub wg_port: u16,
+    pub have_tunnel: Option<bool>, // If we have an existing tunnel, None if we don't know
     pub global: Identity,
 }
 
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -140,16 +140,17 @@ impl Default for TunnelManager {
 
 pub struct IdentityCallback(pub LocalIdentity, pub Peer, pub Option<u16>);
 impl Message for IdentityCallback {
-    type Result = Option<Tunnel>;
+    type Result = Option<(Tunnel, bool)>;
 }
 
 // An attempt to contact a neighbor has succeeded or a neighbor has contacted us, either way
 // we need to allocate a tunnel for them and place it onto our local storage.  In the case
 // that a neighbor contacts us we don't have a port already allocated and we need to choose one
 // in the case that we have atempted to contact a neighbor we have already sent them a port that
-// we now must attach to their tunnel entry.
+// we now must attach to their tunnel entry. If we also return a bool for if the tunnel already
+// exists
 impl Handler<IdentityCallback> for TunnelManager {
-    type Result = Option<Tunnel>;
+    type Result = Option<(Tunnel, bool)>;
 
     fn handle(&mut self, msg: IdentityCallback, _: &mut Context<Self>) -> Self::Result {
         let peer_local_id = msg.0;
@@ -316,6 +317,7 @@ fn contact_neighbor(peer: Peer, our_port: u16) -> Result<(), Error> {
         my_id: LocalIdentity {
             global: SETTING.get_identity(),
             wg_port: our_port,
+            have_tunnel: None,
         },
         to: peer,
     });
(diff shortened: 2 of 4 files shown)


mpapierski:
This looks like a fix for #197

althea_types/src/interop.rs
```diff
@@ -146,6 +146,7 @@ impl Message for Identity {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct LocalIdentity {
     pub wg_port: u16,
+    pub have_tunnel: Option<bool>, // If we have an existing tunnel, None if we don't know
```
mpapierski: Maybe there is some more information we can put here instead of just bool?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -403,30 +406,64 @@ impl TunnelManager {
...
+        trace!("getting existing tunnel or opening a new one");
+        // ifidx must be a part of the key so that we can open multiple tunnels
+        // if we have more than one physical connection to the same peer
         let key = &(their_localid.global.clone(), peer.ifidx);
-        if self.tunnels.contains_key(key) {
+
+        let we_have_tunnel = self.tunnels.contains_key(key);
+        let they_have_tunnel = match their_localid.have_tunnel {
+            Some(v) => v,
+            None => true, // when we don't take the more conservative option
```
mpapierski: This reads like its wrong: `they_have_tunnel` is `true` when `their_lcoalid.have_tunnel` is `None`?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -375,6 +377,7 @@ impl TunnelManager {
                 }
                 Err(e) => {
                     warn!("Actor mailbox failure from DNS resolver! {:?}", e);
+                    // We might need a port callback here
```
mpapierski: TODO?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -403,30 +406,64 @@ impl TunnelManager {
...
+                    iface_name, res
+                );
+            }
+
+            // In the case that we have a tunnel and they don't we drop our existing one
+            // and agree on the new parameters in this message
+            self.tunnels.remove(key);
+            self.free_ports.push(port);
+            return_bool = true;
         }
```
mpapierski: What about case for `!we_have_tunnel`? Shouldn't this handler return some kind of error?

althea_types/src/interop.rs
jkilpatr: A bool is all we really need, although the option + bool setup really makes it the same as two bools.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: it's an assumption, we don't know if they have a tunnel so we take the more conservative option and assume they do.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: maybe, I'm not sure yet, just a hunch I got while writing this but didn't want to forget.

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: that case is the mirror of the case on the other side where the other side takes action to fix the problem.

PR #223, opened 2018-08-20
Rustfmt stable

jkilpatr:
moves us back to rustfmt stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
   - which diesel || cargo install diesel_cli --no-default-features --features sqlite
   - sudo cp $(which diesel) /usr/bin
 env:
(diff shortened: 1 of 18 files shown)


.travis.yml
```diff
@@ -8,7 +8,7 @@ before_install:
   - which diesel || cargo install diesel_cli --no-default-features --features sqlite
   - sudo cp $(which diesel) /usr/bin
 env:
-  - TEST_COMMAND="rustup component add rustfmt-preview --toolchain nightly && cargo +nightly fmt --all -- --check"
+  - TEST_COMMAND="rustup component add rustfmt-preview --toolchain stable && cargo fmt --all && git diff-index --cached --quiet HEAD --ignore-submodules -- || exit 1"
```
mpapierski: I think in this case we can ignore cargo fmt for nightly/beta and when it will run on stable we can be sure cargofmt is present (actually I'm not sure 100% if rust on travis pull cargofmt by default).

PR #225, opened 2018-08-20
Added debts API

### LINKED ISSUE #214: Expose debts through API
Implement an API endpoint that would display list of debts. This API will be consumed in dashboard. Should work on both exit nodes, and client nodes.

# Example

Request

```
curl -H'Accept: application/json' http://127.0.0.1:12345/api/debts
```

Example response

```json
{
  "fd80::f00": 12345,
  "fd80::f01": 12346
}
```

Keys are the mesh IPs, and values are debt amounts.

It could be a up to a consumer how to display this data. For example it could be sorted by debt value, and IP address.

mpapierski:
This PR implements debts API. Unfortunately we can't just simply transform `HashMap<K, V>` into a JSON object when `K` is a struct. Because of that this API returns a list of tuples similiar to `Vec<(K, V)>`.

Closes #214

### DIFF
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -559,7 +559,7 @@ Format:
       "wg_public_key": "pubkey"
     },
     {
-      "total_payment_recieved": "0x0",
+      "total_payment_received": "0x0",
       "total_payment_sent": "0x0",
       "debt": "0",
       "incoming_payments": "0"
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -534,7 +534,7 @@ Calling HTTP `DELETE` request on this endpoint causes all tables to be wiped out
 
 ## /debts
 
-Calling HTTP `GET` request on this endpoint returns a list of debts. Each element of the resulting list contains a 2-elements array where first elements contains `Identity` data, and second element contains details about the debt owed.
+Calling HTTP `GET` request on this endpoint returns a list of debts. Each element of the resulting list contains a dictionary with two keys: `identity` with a dictionary with identity-related information, and `payment_details` key with a value of payments related informations.
 
 - URL: `<rita ip>:<rita_dashboard_port>/debts`
 - Method: `GET`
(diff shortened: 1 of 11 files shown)


docs/api/router-dashboard.md
```diff
@@ -531,3 +531,36 @@ Calling HTTP `DELETE` request on this endpoint causes all tables to be wiped out
...
+  - Code: 200 OK
+  - Contents: `JSON` structured message. See below for an example format.
+- Error Response: `500 Server Error`
+- Sample Call
+
+`curl 127.0..1:<rita_dashboard_port>/debts`
+
+Format:
+```
+[
```
jkilpatr: is this in json style with field names or just the data?

rita/src/client.rs
```diff
@@ -229,6 +229,7 @@ fn main() {
             ).route("/info", Method::GET, get_own_info)
             .route("/version", Method::GET, version)
             .route("/wipe", Method::POST, wipe)
+            .route("/debts", Method::GET, get_debts)
```
jkilpatr: why not in the rita_common endpoints?

rita/src/exit.rs
```diff
@@ -236,6 +236,7 @@ fn main() {
             .route("/version", Method::GET, version)
             .route("/wipe", Method::POST, wipe)
             .route("/database", Method::DELETE, nuke_db)
+            .route("/debts", Method::GET, get_debts)
```
jkilpatr: same as for clients

docs/api/router-dashboard.md
mpapierski: Yeah. Actually both keys, and values are dictionaries. Updating right now.

rita/src/exit.rs
jkilpatr: this is me being confused, ignore me.

docs/api/router-dashboard.md
jtremback: @mpapierski could we have it be an object?

```json
{
   identity: {...}
   payment_details: {...}
}
```

mpapierski: Good point. Not sure why I didn't think of it before. Feels much better. Thanks!

docs/api/router-dashboard.md
```diff
@@ -549,17 +549,22 @@ Calling HTTP `GET` request on this endpoint returns a list of debts.
...
-   debt_buffer]],
-  ...
+  [
+    {
+      "mesh_ip": "a:b:c:d:e:f:g:h",
+      "eth_address": "0x0101010101010101010101010101010101010101",
+      "wg_public_key": "pubkey"
+    },
+    {
+      "total_payment_recieved": "0x0",
```
asoltys: should be received not recieved

mpapierski: Thanks for pointing out @asoltys. I'll run `sed` to replace all occurrences globally in repo.

PR #227, opened 2018-08-21
Improve pre-commit hook.

mpapierski:
Now this script will list all inconsistent files, and should give more
accurate information about how to fix the problems.


mpapierski:
I'm merging it - this is adhoc change in git hook that doesn't influence any other part of the system.

Issue #215, opened 2018-08-17
Tunnel State Management


jkilpatr:
Not to be confused with tunnel lifetime management, Tunnel state management involves checking that a user is on the subnet dao and allowing or disallowing them based on that state. 

This same system should be easily generalizable for payment and non payment. Although we shouldn't include anything for that now. 

The Tunnel struct in TunnelManager should have a 'state' enum this enum will include possible states from in registry to not in registry, to later be expanded to included 'in registry and paid, in registry and unpaid' you get the idea. 

Create a new Actor message 'tick' that is called in rita loop every system tick that will iterate over the tunnels and perform a membership query on the subnet dao (to be created as a separate task) before handling state changes. Commands (shell commands) should only be run when the state has changed, not if it remains the same. To avoid excessive shelling out.

jkilpatr:
resolved

PR #229, opened 2018-08-22
[WIP] Tunnel state manipulation

mpapierski:
Work in progress. Some fighting was done to allow tests for `Tunnel` struct.

Implementation for #215

### DIFF
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -557,7 +557,7 @@ impl TunnelManager {
             );
             let tunnel = tunnels
                 .get(&peer.ifidx)
-                .expect("Unable to find tunnel by ifidx {}", peer.ifidx);
+                .unwrap_or_else(|| panic!("Unable to find tunnel by ifidx {}", peer.ifidx));
             return Ok((tunnel.clone(), true));
         }
 
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -60,7 +60,9 @@ pub enum TunnelManagerError {
 #[derive(Debug, Clone)]
 pub enum TunnelAction {
     /// Received confirmed membership of an identity
-    MembershipConfirmed(Identity),
+    MembershipConfirmed,
+    /// Membership expired for an identity
+    MembershipExpired,
 }
 
 impl fmt::Display for TunnelAction {
@@ -118,8 +120,8 @@ impl Tunnel {
             listen_ifidx: ifidx,
             listen_port: our_listen_port,
             localid: their_id.clone(),
-            // By default new tunnels are in NotRegistered state
-            state: TunnelState::NotRegistered,
+            // By default new tunnels are in Registered state
+            state: TunnelState::Registered,
         }
     }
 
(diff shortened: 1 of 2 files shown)


rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -50,6 +52,46 @@ type Resolver = resolver::Resolver;
 pub enum TunnelManagerError {
     #[fail(display = "Port Error: {:?}", _0)]
     PortError(String),
+    #[fail(display = "Invalid state")]
+    InvalidStateError,
+}
+
+/// Action that progresses the state machine
```
jkilpatr: what does this provide that just setting the enum doesn't?

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -59,48 +101,80 @@ pub struct Tunnel {
...
         ip: IpAddr,
+        iface_name: String,
         our_listen_port: u16,
         ifidx: u32,
         their_id: LocalIdentity,
-    ) -> Result<Tunnel, Error> {
-        let iface_name = KI.setup_wg_if().unwrap();
-
-        let tunnel = Tunnel {
+    ) -> Tunnel {
```
jkilpatr: ok I'm concerned about this, we need to run new -> open -> monitor and in tests you just run new -> monitor that's the idea right? 

My big concern is that I designed new to encapsulate everything so that it was impossible for us to have a tunnel object without having the real tunnel. This reduces the risk of desyncing of the real system tunnel state from what we have represented here. I guess we only actually create tunnels in one spot so it shouldn't be too hard to keep straight. But that's my worry.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -511,30 +585,156 @@ impl TunnelManager {
...
+                error!(
+                    "Unable to execute babel monitor on tunnel {:?}: {}",
+                    tunnel, e
+                );
+                Err(e)
             }
         }
     }
+
+    pub fn find_tunnel_by_id(&mut self, identity: &Identity) -> Option<&mut Tunnel> {
```
jkilpatr: why this instead of HashMap<Identity, Vec<tunnel>>?

rita/src/rita_common/tunnel_manager/mod.rs
mpapierski: My idea was to differentiate actions that could be applied to a tunnel, from a state itself. More like an "event" that happens.

rita/src/rita_common/tunnel_manager/mod.rs
mpapierski: There is no safety measurements between opening a tunnel and inserting a tunnel in the hashmap. After opening a tunnel there are few steps that might fail - `TcpStream::connect`, `tunnel.restrict`, `tunnel.monitor`. What if we'd add new state i.e. "TunnelState::New" before openning a tunnel, and change it to "TunnelState::NotRegistered" after all steps will pass? Of course `TunnelState::New` would be already in the internal hashmap. This way you could wipe tunnels which are in `TunnelState::New` for too long? Or just simply wipe the tunnel in any case one of the operations would fail?

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: The question here is if that will simplify anything.

jkilpatr: I could see it being useful in very complicated state machines. Not sure we're at that level

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: why don't we wait to insert the tunnel until these operations have succeeded? That way we don't need any garbage collection on the tunnels list.

mpapierski: @jkilpatr my point is that first we do `open_tunnel` which already does changes in the system. If any further operation before inserting into hashmap fails we can have a tunnel which doesn't reflect our state, becase we return early with Err (thanks to `?` operator). What do you think about such case?

PR #230, opened 2018-08-24
Update actix and actix-web to use official repos.

### LINKED ISSUE #217: Remove dependency on kindiana's fork of Actix
https://github.com/Kimundi/owning-ref-rs/pull/44

No longer needed.

mpapierski:
This was smooth update. All fixes of @kingoflolz are merged upstream already so no need to pin to older and we can benefit from newer versions.

Closes #217

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "actix"
-version = "0.7.0"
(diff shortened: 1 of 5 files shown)


mpapierski:
Blocked by pending upstream PR that re-introduces `actix::actors::Mocker`. See #217 for more context.

PR #231, opened 2018-08-24
SubnetDAO querying and enforcement

jkilpatr:
This is WIP in that it needs to be actually tested, but other than minor changes to the request/response JSON I think it's done and should be reviewed

### DIFF
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -567,7 +567,7 @@ Format:
 
 ## /dao_list
 
-Calling HTTP `GET` request on this endpoint returns a list of EthAddress' for a configured subnet dao
+Calling HTTP `GET` request on this endpoint returns a list of EthAddresses for a configured subnet DAO. If no DAO is configured it will return an empty list. 
 
 - URL: `<rita ip>:<rita_dashboard_port>/dao_list`
 - Method: `GET`
@@ -594,7 +594,7 @@ Format:
 
 ## /dao_list/add/{address}
 
-Calling HTTP `POST` request on this endpoint removes the provided address from the configured list of
+Calling HTTP `POST` request on this endpoint adds the provided address to the configured list of
 SubnetDAO's
 
 - URL: `<rita ip>:<rita_dashboard_port>/dao_list/add/{address}`
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -25,8 +25,6 @@ This file documents the dashboard API found in Rita client.
 
 `curl 127.0.0.1:4877/info`
 
-
-
 ## /neighbors
 
 - URL: `<rita ip>:<rita_dashboard_port>/neighbors`
@@ -63,8 +61,6 @@ This file documents the dashboard API found in Rita client.
 - Sample Call:
 
 `curl 127.0.0.1:4877/neighbors`
-
-
 
 ## /exits
 
@@ -117,11 +113,13 @@ This file documents the dashboard API found in Rita client.
   - Contents: `{}`
 - Error Response: `400 Bad Request`
 - Error Contents:
+
 ```json
 {
-    "error": "<description>",
+  "error": "<description>"
 }
 ```
+
 - Sample Call:
 
 `curl -XPOST 127.0.0.1:4877/exits/borked/reset`
@@ -138,11 +136,13 @@ This file documents the dashboard API found in Rita client.
   - Contents: `{}`
 - Error Response: `400 Bad Request`
 - Error Contents:
+
 ```json
 {
-    "error": "<description>",
+  "error": "<description>"
 }
 ```
+
 - Sample Call:
 
 `curl -XPOST 127.0.0.1:4877/exits/borked/reset`
@@ -159,12 +159,14 @@ This file documents the dashboard API found in Rita client.
   - Contents: `{}`
 - Error Response: `400 Bad Request`
 - Error Contents:
+
 ```json
 {
-    "error": "<description>",
-    "rust_error": "<stringified_rust_error>"
+  "error": "<description>",
+  "rust_error": "<stringified_rust_error>"
 }
 ```
+
 - Sample Call:
 
 `curl -XPOST 127.0.0.1:4877/exits/borked/register`
@@ -176,20 +178,22 @@ This file documents the dashboard API found in Rita client.
   `{nickname}` for verification using `{code}`
 - Method: `POST`
 - URL Params:
-    - `nickname`, string
-    - `code`, string
+  - `nickname`, string
+  - `code`, string
 - Data Params: `None`
 - Success Response:
   - Code: 200 OK
   - Contents: `{}`
 - Error Response: `400 Bad Request`
 - Error Contents:
+
 ```json
 {
-    "error": "<description>",
-    "rust_error": "<stringified_rust_error>"
+  "error": "<description>",
+  "rust_error": "<stringified_rust_error>"
 }
 ```
+
 - Sample Call:
 
 `curl -XPOST 127.0.0.1:4877/exits/borked/register`
@@ -315,8 +319,6 @@ This file documents the dashboard API found in Rita client.
 
 `curl 127.0.0.1:4877/settings`
 
-
-
 ## /settings
 
 - URL: `<rita ip>:<rita_dashboard_port>/settings`
@@ -340,8 +342,6 @@ This file documents the dashboard API found in Rita client.
 
 `curl -XPOST 127.0.0.1:<rita_dashboard_port>/settings -H 'Content-Type: application/json' -i -d '{"exit_client": {"current_exit": "SELECTEDEXIT"}}'`
 }
-
-
 
 ## /wifi_settings
 
@@ -400,8 +400,6 @@ This file documents the dashboard API found in Rita client.
 
 `curl 127.0.0.1:4877/wifi_settings`
 
-
-
 ## /wifi_settings
 
 - URL: `<rita ip>:<rita_dashboard_port>/wifi_settings`
@@ -444,7 +442,6 @@ This file documents the dashboard API found in Rita client.
 
 `curl -XPOST 127.0.0.1:<rita_dashboard_port>/wifi_settings/ssid -H 'Content-Type: application/json' -i -d '{"radio":"radio0", "ssid": "this is a freeform ssid"}'`
 
-
 ---
 
 ## /wifi_settings/pass
@@ -466,7 +463,6 @@ This file documents the dashboard API found in Rita client.
 - Sample Call:
 
 `curl -XPOST 127.0.0.1:<rita_dashboard_port>/wifi_settings/pass -H 'Content-Type: application/json' -i -d '{"radio":"radio0", "pass": "this is a freeform password"}'`
-
 
 ---
 
@@ -490,9 +486,7 @@ This file documents the dashboard API found in Rita client.
 
 `curl -XPOST 127.0.0.1:<rita_dashboard_port>/wifi_settings/mesh -H 'Content-Type: application/json' -i -d '{"radio":"radio0", "mesh": true}'`
 
-
 ---
-
 
 ## /wipe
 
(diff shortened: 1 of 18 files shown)


rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+            self.entries.insert(their_id.clone(), v);
+
+            send_membership_message(on_dao, their_id);
+        }
+    }
+}
+
+/// True if timestamp does not need to be updated
+fn timer_check(timestamp: Instant) -> bool {
+    if Instant::now() - timestamp < SETTING.get_dao().cache_timeout {
```
asoltys: You could drop the if/else and just return the result of the condition here

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -148,12 +153,29 @@ fn timer_check(timestamp: Instant) -> bool {
     }
 }
 
+/// Sends off a message to TunnelManager about the dao state
+fn send_membership_message(on_dao: bool, their_id: Identity) -> () {
```
mpapierski: I'd simplify it TBH - instead of `bool` arg I'd accept `TunnelAction` directly. This way we could save few branches of code.

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -97,44 +100,46 @@ impl Handler<CacheCallback> for DAOManager {
...
+                on_list: on_dao,
                 dao_address: dao_address,
                 id: their_id.clone(),
                 last_updated: Instant::now(),
             };
             let mut v = Vec::new();
             v.push(entry);
-            self.entries.insert(their_id, v);
-            // send appropriate dao message
+            self.entries.insert(their_id.clone(), v);
```
mpapierski: `vec![entry]`

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -169,13 +191,13 @@ fn check_cache(their_id: Identity, entries: &HashMap<Identity, Vec<DAOEntry>>) -
...
                     );
-                // send approval
+                    send_membership_message(true, their_id.clone());
                 } else if !timer_check(entry.last_updated) {
                     get_membership(entry.dao_address, entry.id.clone());
                 }
             }
             trace!("{:?} is not on any SubnetDAO", their_id);
-            // send rejection
+            send_membership_message(false, their_id);
```
mpapierski: Just noticed there is no `.clone()` in `their_id` as there is few lines above. Suspicious.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -667,8 +668,10 @@ impl Handler<TunnelStateChange> for TunnelManager {
                             info!("Membership for identity {:?} is expired", msg.identity);
                             match tunnel.state {
                                 TunnelState::Registered => {
-                                    tunnel.unmonitor(make_babel_stream()?)?;
-                                    tunnel.state = TunnelState::NotRegistered;
+                                    // TODO make this real after testing
+                                    trace!("Fake removing {:?} from peers", msg.identity);
+                                    //tunnel.unmonitor(make_babel_stream()?)?;
+                                    //tunnel.state = TunnelState::NotRegistered;
```
mpapierski: Hmm at least this state manipulation could be done to check state changes.

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    type Result = ();
+
+    fn handle(&mut self, msg: DAOCheck, _: &mut Context<Self>) -> Self::Result {
+        let their_id = msg.0;
+        check_cache(their_id, &self.entries);
+    }
+}
+
+/// Called by returning DAO requests, sends a message to TunnelManager
+pub struct CacheCallback((Identity, EthAddress, Web3Response));
```
mpapierski: Its probably my favorite nitpick but in non-trivial messages I'd rather see named structs instead of tuples. This way we can avoid mysterious calls to `(msg.0).2` etc.

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    fn handle(&mut self, msg: CacheCallback, _: &mut Context<Self>) -> Self::Result {
+        let their_id = (msg.0).0;
+        let dao_address = (msg.0).1;
+        let response = (msg.0).2;
+
+        let has_vec = self.entries.contains_key(&their_id);
+
+        if has_vec {
+            let entry_vec = self.entries.get_mut(&their_id).unwrap();
```
mpapierski: Those two lookups could be replaced with single `match entries.get(&their_id) { ... }`

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+            // No entry exists in the HashMap for this ID, create one
+            let entry = DAOEntry {
+                on_list: response.result == Uint256::zero(),
+                dao_address: dao_address,
+                id: their_id.clone(),
+                last_updated: Instant::now(),
+            };
+            let mut v = Vec::new();
+            v.push(entry);
+            self.entries.insert(their_id, v);
```
mpapierski: As alternative: `vec![entry]`

rita/src/rita_common/dao_manager/mod.rs
mpapierski: +1

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    };
+    let mut full_bytes: [u8; 32] = [0; 32];
+    let mut i = 0;
+    for byte in ip_bytes.iter() {
+        full_bytes[i] = byte.clone();
+        i = i + 1;
+    }
+
+    let call_args: Uint256 = full_bytes.into();
+    let func_call = format!("{{'jsonrpc':'2.0','method':'eth_call','params':[{{'to': '{:x}', 'data': '{:x}'}}, 'latest'],'id':1}}", dao_address, call_args);
```
mpapierski: Just for brevity we could serialize proper struct that imitates JSONRPC request.

drozdziak1: This doesn't feel right to me either.

rita/src/rita_common/dao_manager/mod.rs
jkilpatr: I think this is better since it encapsulates the TunnelManager state stuff, but I'm pretty neutral overall.

rita/src/rita_common/dao_manager/mod.rs
jkilpatr: you don't need it since the object is dropped shortly after, should I add it anyway?

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: good point

rita/src/rita_common/dao_manager/mod.rs
jkilpatr: I'm of the same opinion, I guess I'll make a cache callback object

rita/src/rita_common/dao_manager/mod.rs
jkilpatr: the real issue is the nesting required, the struct gets rather complicated looking. Not to mention how do I get 'latest' to serialize without a field name? If you can give me a struct that serializes this way I'll be happy but I don't know how to make one.

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl Handler<CacheCallback> for DAOManager {
+    type Result = ();
+
+    fn handle(&mut self, msg: CacheCallback, _: &mut Context<Self>) -> Self::Result {
+        let their_id = (msg.0).0;
+        let dao_address = (msg.0).1;
+        let response = (msg.0).2;
+
+        let has_vec = self.entries.contains_key(&their_id);
```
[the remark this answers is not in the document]
jkilpatr: lifetimes! By putting it into a variable the borrow on entries ends right there, but I use it at the top of a statement (or in a match) it doesn't and the borrow checker refuses to allow my mutable borrow.

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        let on_dao = response.result == Uint256::zero();
```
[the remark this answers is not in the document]
jkilpatr: good catch

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+impl Default for DAOManager {
+    fn default() -> DAOManager {
+        DAOManager::new()
+    }
+}
+
+impl DAOManager {
+    fn new() -> DAOManager {
+        DAOManager {
+            entries: HashMap::<Identity, Vec<DAOEntry>>::new(),
```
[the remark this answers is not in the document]
jkilpatr: done

rita/src/rita_common/dao_manager/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        let has_vec = self.entries.contains_key(&their_id);
+        let on_dao = response.result == Uint256::zero();
+
+        if has_vec {
+            let entry_vec = self.entries.get_mut(&their_id).unwrap();
+            let mut found_entry = false;
+            for entry in entry_vec.iter_mut() {
+                // There exists an entry, and we found it, so update it
+                if dao_address == entry.dao_address {
```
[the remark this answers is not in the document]
jkilpatr: good catch

jkilpatr:
Running into some TrustDNS problems testing this, as far as I can tell the future for DNS resolution gets issued and then hangs indefinitely, please let me know if you can or can not reproduce. 

Either way I don't think we want ot put the fix into this pull request, but that leaves the interesting question of how to test it.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -671,7 +671,7 @@ impl Handler<TunnelStateChange> for TunnelManager {
                                     // TODO make this real after testing
                                     trace!("Fake removing {:?} from peers", msg.identity);
                                     //tunnel.unmonitor(make_babel_stream()?)?;
```
mpapierski: Lets uncomment it before shipping this PR?

jkilpatr:
this fully works and needs final review

jkilpatr:
as I was concerned about this has introduced some interesting bugs, if you take a node and reboot it's peer it ends up with 2 ifidx0 tunnels. Some search parameter is slightly off.

jkilpatr:
as I was concerned about this has introduced some interesting bugs, if you take a node and reboot it's peer it ends up with 2 ifidx0 tunnels. Some search parameter is slightly off.

jkilpatr:
ha ha found it! Very tiny merge issue with big potential repercussions. Fixed now.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -675,9 +660,6 @@ impl TunnelManager {
                     );
                 }
 
-                // In the case that we have a tunnel and they don't we drop our existing one
-                // and agree on the new parameters in this message
-                self.tunnels.remove(&key);
```
mpapierski: This removal looks suspicious. Are you sure it should be this way?

jkilpatr: @mpapierski you remove the key further up right? Here we remove the identity when only a single tunnel may need to be recreated in this special case. This creates orphaned tunnels.

docs/api/router-dashboard.md
```diff
@@ -569,4 +563,81 @@ Format:
...
+Calling HTTP `GET` request on this endpoint returns a list of EthAddress' for a configured subnet dao
+
+- URL: `<rita ip>:<rita_dashboard_port>/dao_list`
+- Method: `GET`
+- URL Params: `None`
+- Data Params: `None`
+- Success Response:
+  - Code: 200 OK
+  - Contents: `JSON` structured message. See below for an example format.
+- Error Response: `500 Server Error`
```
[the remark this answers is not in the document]
jkilpatr: nope that's just an empty list.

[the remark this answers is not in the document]
jkilpatr: there noted that explicitly

docs/api/router-dashboard.md
```diff
@@ -569,4 +563,81 @@ Format:
...
+
+Calling HTTP `GET` request on this endpoint returns a list of EthAddress' for a configured subnet dao
+
+- URL: `<rita ip>:<rita_dashboard_port>/dao_list`
+- Method: `GET`
+- URL Params: `None`
+- Data Params: `None`
+- Success Response:
+  - Code: 200 OK
+  - Contents: `JSON` structured message. See below for an example format.
```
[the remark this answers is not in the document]
jkilpatr: well I think it would take up too much room there.

PR #237, opened 2018-08-27
Use 0.4.0 version of ethereum-types.

mpapierski:
See althea-mesh/guac_rs#6 for more details. This will pin the version to 0.4.x. This way in `guac_rs` should be improved to fix the build errors related to version mismatch.

Also there is a PR pending on guac side to remove `Cargo.lock` which also should help in dependency resolving.


mpapierski:
Depending on any code from Parity seems like a nightmare. There is an internal dependency somewhere in parity code for 0.3 that you can't swap, also parity-common has crate.io v0.4.0 BUT there is no tag so git pinning wouldn't work. I reported it, but still probably better to stick with 0.3.x.

PR #238, opened 2018-08-27
Input scrubbing for UCI values

jkilpatr:
Fixes things for people who want a wifi name or password with '


jkilpatr:
@mpapierski it's possible escaping is enough not sure how much we want to get into it.

Issue #236, opened 2018-08-27
Guac integration


mpapierski:
Integrate https://github.com/althea-mesh/guac_rs into the codebase.

# TODO

- [x] Initial PR of integration @kingoflolz #151 is out of date. Needs a rebase and make it healthy enough, so once we'll be able to provide our own implementation of tx signing it would be easy to merge.
- [x] ~Clarify where Guac will fits considering problems mentioned below. What would be its responsibility in relation of #232 which would bring ETH signing.~ Needs more research as part of #201  before deciding.

# Problems

- Guac won't work on MIPS architectures because of Parity dependencies.
- Interdependency on `ring` can be partially mitigated as it works on MIPS and big-endian (at least for most of the necessary code paths. There is still at least one test failure in `ring`, but tests with `parity-common` doesn't fail on crypto so we can probably delay fixing it).
- `parity-common` (one of internal deps) massively fails on big endian (see discussion in #201)

# Solution

I'm in contact (through PRs) with Parity team on the big endian problems, but initial research in #201 showed that probably writing our own ETH library could be easier approach as described in #232. 

@jkilpatr @jtremback

mpapierski:
So far I'd consider this issue closed. Guac repo is in healthy state and old pr in #151 is rebased and brought to a healthy state.

At least for now until new requirements are present when we made some progress with clarity.

Issue #25, opened 2018-01-25
num256 needs a math review


jtremback:

jtremback:
@mpapierski The only real purpose of this library is to panic on overflow. Do you not feel that this is important?

jtremback:
For instance, without overflow checking, I could cause my channel counterparty to believe that an update that will end up overflowing and throwing on the chain is valid. Perhaps they might then garbage collect their old updates. This would leave them unable to settle and vulnerable to an old update attack.

We could overflow check elsewhere, but why? Checking in the number lib seems much cleaner. Using this num256 library also gives us the flexibility to possibly switch to an optimized implementation on some platforms instead of just wrapping the bigint stuff.

mpapierski:
@jtremback You're right and I agree that we should do overflowing checks to keep it consistent with Ethereum's data types.

While playing with num256 I think I saw place where we mix Int256 and Uint256 - where normally rust forbids math on i32 and u32. Its potentially unsafe as there may be involved implicit conversion. I think a review of num256 usage would also improve stability.

jtremback:
@mpapierski good catch- do you remember where?

PR #240, opened 2018-08-28
Use ethereum-types 0.3

mpapierski:
This removes operators which are part of 0.4. In guac we ingest eth-types 0.4 too (althea_types) but also eth-types 0.3 (ethabi). This isn't easily replacable, so probably going back to 0.3 everywhere is safest thing to do.

I tried two approaches:

- Remove num256 and use BigInt/BigUint directly. Ran into custom serialization issue for backwards compatibility (default BigInt/BigUint serialization produces array of array bytes and we keep big nums as string, big unsinged nums as hexadecimal etc.). Kinda works, but needs more effort to figure out failing test cases.
- Make Uint256 a wrapper too, possibly implementing operators and make it look like a type (just like Int256). Probably the best way to go, just need some effort to write all traits to make Int256/Uint256 usable like normal numbers (and there are a lot of them)

This debt should be paid at some point, possibly could be done as part of #25.

And of course this will move forward #236

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,7 +130,7 @@ version = "0.1.0"
 dependencies = [
  "actix 0.7.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.9.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


jkilpatr:
looks good to me

PR #241, opened 2018-08-29
rita: Delete tunnels after a timeout

drozdziak1:
This commit adds a last_contact field to the Tunnel struct which keeps
track of the last point in time when we heard from or successfuly
hollered at a neighbor. On every such occasion we bump last_contact to
Instant::now().

Concordantly, the common rita_loop will send a new TriggerGC message to
TunnelManager containing the timeout to check open tunnels against. The
TriggerGC handler will iterate over all open tunnel entries and discard
those which exceed the passed timeout duration along with the underlying
wgN interfaces and ports.

rita/src/rita_common/rita_loop/mod.rs:
* Send a TriggerGC message to TunnelManager every tick

rita/src/rita_common/tunnel_manager/mod.rs:
* Add a last_contact field (std::time::Instant) to the Tunnel struct;
  Instant::now() is the default value.
* Bump the last_contact field in open_tunnel() whenever we have a
  tunnel; move some ifs around - the logic's implications stay the same.
* Add a TriggerGC message which goes over all tunnels and discards those
  which were last contacted longer ago than a passed std::time::Duration
  value.

### DIFF
--- a/rita/src/rita_common/rita_loop/mod.rs
+++ b/rita/src/rita_common/rita_loop/mod.rs
@@ -8,7 +8,7 @@ use actix_utils::ResolverWrapper;
 
 use KI;
 
-use rita_common::tunnel_manager::{GetNeighbors, TunnelManager};
+use rita_common::tunnel_manager::{GetNeighbors, TriggerGC, TunnelManager};
 
 use rita_common::traffic_watcher::{TrafficWatcher, Watch};
 
@@ -124,6 +124,24 @@ impl Handler<Tick> for RitaLoop {
                             actix::fut::ok(())
                         })
                 }),
+        );
+
+        let start = Instant::now();
+        Arbiter::spawn(
+            TunnelManager::from_registry()
+                .send(TriggerGC(Duration::from_secs(
+                    SETTING.get_network().tunnel_timeout_seconds,
+                )))
+                .then(move |res| {
+                    info!(
+                        "TunnelManager GC pass completed in {}s {}ms, with result {:?}",
+                        start.elapsed().as_secs(),
+                        start.elapsed().subsec_nanos() / 1000000,
+                        res
+                    );
+                    res
+                })
+                .then(|_| Ok(())),
         );
 
         let start = Instant::now();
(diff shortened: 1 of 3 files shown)


rita/src/rita_common/rita_loop/mod.rs
```diff
@@ -126,6 +126,22 @@ impl Handler<Tick> for RitaLoop {
                 }),
         );
 
+        let start = Instant::now();
+        Arbiter::spawn(
+            TunnelManager::from_registry()
+                .send(TriggerGC(Duration::from_secs(15)))
```
mpapierski: As mentioned few minutes (i.e. 15min) might be more user friendly. Or even a configuration setting with a default value.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -288,6 +291,49 @@ impl Handler<GetNeighbors> for TunnelManager {
     }
 }
 
+/// A message type for deleting all tunnels we haven't heard from for more than the duration.
+pub struct TriggerGC(pub Duration);
```
mpapierski: Obligatory tuple vs struct readability nitpick: lets add a name for duration?

rita/src/rita_common/rita_loop/mod.rs
jkilpatr: yes this needs to take a configured duration

rita/src/rita_common/tunnel_manager/mod.rs
jkilpatr: we only have one object here. I think that's too far.

settings/src/lib.rs
```diff
@@ -121,14 +125,17 @@ pub struct NetworkSettings {
...
+    pub tunnel_timeout_seconds: u64,
 }
 
 impl Default for NetworkSettings {
     fn default() -> Self {
         NetworkSettings {
             own_ip: "fd00::1".parse().unwrap(),
             bounty_ip: "fd00::3".parse().unwrap(),
-            discovery_ip: Ipv6Addr::new(0xff02, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x8),
+            discovery_ip: default_discovery_ip(),
```
jkilpatr: kinda curious how did you edit this but default _discovery_ip already existed? did I just miss it once?

PR #245, opened 2018-08-31
Divide the balance to ETH.

### LINKED ISSUE #244: Payments page crash
Important cause was already fixed in #188, but turns out there are still cases where balance page gets crash. Our priority is to not interrupt users, so this task is about reviewing and fixing the cases for that.

- [x] In https://github.com/althea-mesh/althea_rs/blob/d62fa3dc124cb4c3650e73c68554b6c51e3549a9/rita/src/rita_common/payment_controller/mod.rs#L92-L101 sounds like we're converting between `Int256` and `i64` which may fail when `balance > i64::max_value()`. Shouldn't be hard to hit this as ethereum is divisible up to 18 decimal places.

This seems like an easy thing to add unit test for to check those corner cases.

mpapierski:
Added a lot of safety i.e.  do a checked divide by 10^9 first, and later do a checked downcast. In case of problems it should result in HTTP 500 instead of a panic.

Caution: Balance less than 10^9 would display 0.

Closes #244


jkilpatr:
looks good to me.

PR #249, opened 2018-09-03
Doc Rita and settings

jkilpatr:
This commit adds documentaiton block paragraphs with the goal of describing
the functions of each file at the top of the file.

### DIFF
--- a/rita/src/exit.rs
+++ b/rita/src/exit.rs
@@ -1,3 +1,14 @@
+//! This is the main source file for the Rita exit binary, by 'exit' we mean 'a vpn server, not a
+//! mesh router out in the field'.
+//!
+//! All meshing and billing functionaltiy is contained in rita_common and is common to both rita and
+//! rita_exit. The major difference is billing and connection code for the 'exit', the mandatory
+//! vpn system integrated into the Althea network design, as well as API endpoints for a management
+//! dashboard of router functions like wifi, which the exit is not expected to have.
+//!
+//! This file initilizes the dashboard endpoints for the exit as well as the common and exit
+//! specific actors.
+
 #![cfg_attr(feature = "system_alloc", feature(alloc_system, allocator_api))]
 #![cfg_attr(feature = "clippy", feature(plugin))]
 #![cfg_attr(feature = "clippy", plugin(clippy))]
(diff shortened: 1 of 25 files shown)


rita/src/exit.rs
```diff
@@ -1,3 +1,14 @@
+//! This is the main source file for the Rita exit binary, by 'exit' we mean 'a vpn server, not a
+//! mesh router out in the field'.
+//!
+//! All meshing and billing functionaltiy is contained in rita_common and is common to both rita and
+//! rita_exit. The major difference is billing and connection code for the 'exit', the mandatory
```
kingoflolz: should be traffic counting that is different, to distinguish from "billing functionality" which you say is common

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -1,3 +1,11 @@
+//! Traffic watcher monitors system traffic by interfacing with KernelInterface to create and check
+//! ebtables counters on each per hop tunnel (the WireGuard tunnel between two devices). These counts
```
kingoflolz: we don't use ebtables anymore, we use iptables and ipset

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -1,3 +1,7 @@
+//! Traffic watcher monitors system traffic by interfacing with KernelInterface to create and check
+//! ebtables counters on each per hop tunnel (the WireGuard tunnel between two devices). These counts
```
kingoflolz: same here, iptables and ipset

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -1,3 +1,7 @@
+//! DB client is essentially a layer over Diesel which is itself a layer over a sqllite database
+//! at some point we will need to add multi-database support such that we can scale exits
+//! horizontally
+
```
kingoflolz: also contains a lot of the sign up logic and stuff (which it probs should not)

rita/src/rita_exit/network_endpoints/mod.rs
```diff
@@ -1,3 +1,6 @@
+//! Network endpoints for rita-exit there are not dashboard or local infromational endpoints
```
kingoflolz: "that" instead of "there"?

rita/src/rita_exit/traffic_watcher/mod.rs
```diff
@@ -1,3 +1,11 @@
+//! Traffic watcher monitors system traffic by interfacing with KernelInterface to create and check
+//! ebtables counters on each per hop tunnel (the WireGuard tunnel between two devices). These counts
```
kingoflolz: same here, iptables and ipset

rita/src/rita_client/traffic_watcher/mod.rs
jkilpatr: why switch? specifically the iptables lock is a scaling issue.

PR #260, opened 2018-09-13
Improve logging for exits

jkilpatr:
This adds some more info messages with better system summary data
such as total bytes in and out in a billing cycle.

It also demotes some warnings to trace messages as they dominated
total warnings counts without being all that important.

### DIFF
--- a/rita/src/rita_exit/traffic_watcher/mod.rs
+++ b/rita/src/rita_exit/traffic_watcher/mod.rs
@@ -109,7 +109,19 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, clients: Vec<Identity>) -> Re
     let output_counters = KI.read_exit_server_counters(&ExitFilterTarget::Output)?;
 
     trace!("input exit counters: {:?}", input_counters);
+    let mut total_in: u64 = 0;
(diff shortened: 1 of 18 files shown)


rita/src/rita_exit/traffic_watcher/mod.rs
```diff
@@ -109,7 +109,19 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, clients: Vec<Identity>) -> Re
...
     trace!("input exit counters: {:?}", input_counters);
+    let mut total_in: u64 = 0;
+    for entry in input_counters.iter() {
+        let input = entry.1;
+        total_in += input;
+    }
+    info!("Total input of {} bytes this round", total_in);
     trace!("output exit counters: {:?}", output_counters);
+    let mut total_out: u64 = 0;
+    for entry in output_counters.iter() {
```
mpapierski: Its repeating pattern here so you can consider something like: `let total_in = output_counters.iter().fold(0u64, |r,s| r + s.1)`

jkilpatr: I did consider a fancy iterator like that before deciding it didn't provide much gain except looking fancy.

PR #261, opened 2018-09-13
More defensive coding for TrafficWatcher

jkilpatr:
Today we had a crash in prod because of the TrafficWatcher debts
unwrap. As far as I can tell it's possible for this to happen
if we have a tunnel open for a peer but Babel has yet to insert
any kind of route for them.

This race condition required thousands of hours to reveal, hopefully
with more defensive handling and better logging we can identify rare
but possible conditions faster.

### DIFF
--- a/rita/src/rita_client/traffic_watcher/mod.rs
+++ b/rita/src/rita_client/traffic_watcher/mod.rs
@@ -86,7 +86,16 @@ pub fn watch<T: Read + Write>(
         }
     }
 
-    let (input, output) = KI.read_iface_counters("wg_exit")?;
+    let (input, output) = match KI.read_iface_counters("wg_exit") {
+        Ok(res) => res,
+        Err(e) => {
+            warn!(
+                "Error getting input output counters {:?} traffic has gone unaccounted!",
+                e
+            );
+            return Err(e);
+        }
+    };
 
     // account for wg packet overhead
     let input = input.total_bytes();
@@ -118,8 +127,7 @@ pub fn watch<T: Read + Write>(
                         ));
                     }
                 }
-            ))
-            .send()?
+            )).send()?
             .json()?;
         let client_rx = SystemTime::now();
 
--- a/rita/src/rita_common/traffic_watcher/mod.rs
+++ b/rita/src/rita_common/traffic_watcher/mod.rs
@@ -8,7 +8,7 @@ use rita_common::tunnel_manager::Neighbor;
 use althea_kernel_interface::FilterTarget;
 use KI;
 
-use althea_types::LocalIdentity;
+use althea_types::Identity;
 
 use babel_monitor::Babel;
 
@@ -101,13 +101,16 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
     let routes = babel.parse_routes()?;
     info!("Got routes: {:?}", routes);
 
-    let mut identities: HashMap<IpAddr, LocalIdentity> = HashMap::new();
-    let mut if_to_ip: HashMap<String, IpAddr> = HashMap::new();
+    let mut identities: HashMap<IpAddr, Identity> = HashMap::new();
+    let mut if_to_id: HashMap<String, Identity> = HashMap::new();
     let mut ip_to_if: HashMap<IpAddr, String> = HashMap::new();
-    for ident in neighbors {
-        identities.insert(ident.identity.global.mesh_ip, ident.identity.clone());
-        if_to_ip.insert(ident.iface_name.clone(), ident.identity.global.mesh_ip);
-        ip_to_if.insert(ident.identity.global.mesh_ip, ident.iface_name.clone());
+    for neigh in neighbors {
+        // provides a lookup from mesh ip to identity
+        identities.insert(neigh.identity.global.mesh_ip, neigh.identity.global.clone());
+        // provides a lookup from wireguard interface to mesh ip
+        if_to_id.insert(neigh.iface_name.clone(), neigh.identity.global.clone());
+        // provides a lookup from mesh ip to wireguard interface
+        ip_to_if.insert(neigh.identity.global.mesh_ip, neigh.iface_name.clone());
     }
 
     let mut destinations = HashMap::new();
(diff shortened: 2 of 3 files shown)


rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -175,28 +211,60 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
...
+        );
+        match state {
+            (Some(id), Some(_dest), Some(_ifip)) => match debts.get_mut(&id) {
+                Some(debt) => {
+                    *debt -= (destinations[&ip].clone()) * bytes;
+                }
+                // debts is generated from identities, this should be impossible
+                None => warn!("No debts entry for input entry id {:?}", id),
+            },
+            (Some(_id), Some(_), None) => warn!("This case should not be possible"),
```
kingoflolz: Should we group all the "This case should not be possible" cases at the end with a (_, _, _) ?

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -175,28 +211,60 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
...
-        } else {
-            warn!("destination not found {}, {}", ip, bytes);
+        let state = (
+            identities.get(&ip),
+            destinations.get(&ip),
+            if_to_ip.get(&interface),
+        );
+        match state {
+            (Some(id), Some(_dest), Some(_ifip)) => match debts.get_mut(&id) {
+                Some(debt) => {
```
kingoflolz: Same here

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -175,28 +211,60 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
...
-            warn!("flow destination not found {}, {}", ip, bytes);
+        let state = (
+            identities.get(&ip),
+            destinations.get(&ip),
+            if_to_ip.get(&interface),
+        );
+        match state {
+            (Some(id), Some(_dest), Some(_ifip)) => match debts.get_mut(&id) {
+                Some(debt) => {
+                    *debt -= (destinations[&ip].clone()) * bytes;
```
kingoflolz: why don't we use the _dest value instead of doing another lookup?

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -175,28 +211,60 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
...
-            warn!("destination not found {}, {}", ip, bytes);
+        let state = (
+            identities.get(&ip),
+            destinations.get(&ip),
+            if_to_ip.get(&interface),
+        );
+        match state {
+            (Some(id), Some(_dest), Some(_ifip)) => match debts.get_mut(&id) {
+                Some(debt) => {
+                    *debt += (destinations[&ip].clone() - local_price) * bytes;
```
kingoflolz: Same here

rita/src/rita_exit/traffic_watcher/mod.rs
```diff
@@ -121,22 +139,42 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, clients: Vec<Identity>) -> Re
...
-        if identities.contains_key(&ip) && destinations.contains_key(&ip) {
-            let id = identities[&ip].clone();
-            *debts.get_mut(&id).unwrap() -= (destinations[&ip].clone() + price) * bytes;
-        } else {
-            warn!("input sender not found {}, {}", ip, bytes);
+        let state = (identities.get(&ip), destinations.get(&ip));
+        match state {
+            (Some(id), Some(_)) => match debts.get_mut(&id) {
+                Some(debt) => {
+                    *debt -= (destinations[&ip].clone() + price) * bytes;
```
kingoflolz: Reuse result from destinations.get()

rita/src/rita_common/traffic_watcher/mod.rs
jkilpatr: I was thinking about that, but it made it hard for me to keep track of scenarios than this exhaustive binary countdown. don't really care much, I'll let others chime in and change it according to majority vote.

rita/src/rita_exit/traffic_watcher/mod.rs
jkilpatr: oh good point

jkilpatr:
this seems to break billing. I guess because of a reference issue?

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -126,16 +126,52 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
...
+        Ok(res) => res,
+        Err(e) => {
+            warn!(
+                "Error getting input counters {:?} traffic has gone unaccounted!",
+                e
+            );
+            return Err(e);
+        }
+    };
+    let fwd_output_counters = match KI.read_counters(&FilterTarget::ForwardOutput) {
```
mpapierski: Wouldn't `KI.read_counters(...).map_err(|e| { warn!("foo"); e })?` be shorter?

jkilpatr: well I didn't know about map err that's nice. But In general I don't put a very high value on brevity in code. I would rather make it twice as long to increase readability especially since the compiler is going to boil it all down to the same thing anyways.

rita/src/rita_common/traffic_watcher/mod.rs
```diff
@@ -175,28 +211,54 @@ pub fn watch<T: Read + Write>(mut babel: Babel<T>, neighbors: &Vec<Neighbor>) ->
     }
 
     for ((ip, interface), bytes) in total_input_counters {
-        if destinations.contains_key(&ip)
-            && if_to_ip.contains_key(&interface)
-            && identities.contains_key(&if_to_ip[&interface])
-        {
-            let id = identities[&if_to_ip[&interface]].clone();
```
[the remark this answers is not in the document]
jkilpatr: ok yes it does we're looking up the ID of the input rather than the destination IP

rita/src/rita_client/traffic_watcher/mod.rs
```diff
@@ -86,7 +86,16 @@ pub fn watch<T: Read + Write>(
         }
     }
 
-    let (input, output) = KI.read_iface_counters("wg_exit")?;
+    let (input, output) = match KI.read_iface_counters("wg_exit") {
+        Ok(res) => res,
+        Err(e) => {
+            warn!(
+                "Error getting input output counters {:?} traffic has gone unaccounted!",
```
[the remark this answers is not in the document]
jkilpatr: depends on how it failed, the error could in theory tell people that.

PR #263, opened 2018-09-14
[WIP] Port toggling

drozdziak1:
* [x] Retrieve all interfaces
* [ ] Make sure we only get the ones we want the user to control
* [ ] Reliable port mode detection
* [ ] Reliable port toggling strategy

### DIFF
--- a/rita/src/rita_client/dashboard/mod.rs
+++ b/rita/src/rita_client/dashboard/mod.rs
@@ -13,6 +13,7 @@ use serde_json;
 use serde_json::Value;
 use std::collections::HashMap;
 use std::net::{SocketAddr, TcpStream};
+use std::string::ToString;
 use std::{thread, time};
 
 use althea_types::ExitState;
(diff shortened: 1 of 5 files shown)


rita/src/rita_client/dashboard/mod.rs
```diff
@@ -485,3 +486,114 @@ impl Handler<SetWiFiMesh> for Dashboard {
...
+            error!("Mode setting entry {:?} not found", mode_entry_name);
+            bail!("Mode setting entry {:?} not found", mode_entry_name);
+        }
+    };
+
+    // Match mode
+    Ok(match mode_name.as_str() {
+        "adhoc" => InterfaceMode::Mesh,
+        "ap" => InterfaceMode::LAN,
+        "sta" => InterfaceMode::WAN,
```
jkilpatr: wifi as wan? not sure that's worth our time but it is interesting.

PR #268, opened 2018-09-15
Fix exit switching parsing

jkilpatr:
So during my TunnelManager refactor I made the ip parsing in a few
functions more restrictive. During this process I switched a function
used to detect the old exit ip on the exit tunnel from searching for ipaddr
in general to ipv6 addresses, meaning it would never identify ipv4 addresses
and fail to cleanup any leftovers from a previous tunnel.

This not only solves that problem but also goes a step further to prevent
the tunnel from being re-created every loop, reducing commands and possible
error.

### DIFF
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -2,7 +2,7 @@ use super::{KernelInterface, KernelInterfaceError};
 
 use failure::Error;
 
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
 impl KernelInterface {
     pub fn set_client_exit_tunnel_config(
@@ -58,7 +58,7 @@ impl KernelInterface {
             ],
         )?;
 
-        let prev_ip = self.get_global_device_ip("wg_exit");
+        let prev_ip: Result<Ipv4Addr, Error> = self.get_global_device_ip_v4("wg_exit");
 
         match prev_ip {
             Ok(prev_ip) => {
--- a/althea_kernel_interface/src/link_local_tools.rs
+++ b/althea_kernel_interface/src/link_local_tools.rs
@@ -1,6 +1,6 @@
 use super::{KernelInterface, KernelInterfaceError};
 
-use std::net::{IpAddr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use regex::Regex;
 
@@ -12,9 +12,13 @@ impl KernelInterface {
         let output = self.run_command("ip", &["addr", "show", "dev", dev, "scope", "link"])?;
         trace!("Got {:?} from `ip addr`", output);
 
-        let re = Regex::new(r"inet6 (\S*?)(/[0-9]+)? scope link").unwrap();
-        let str = String::from_utf8(output.stdout)?;
-        let cap = re.captures(&str);
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"inet6 (\S*?)(/[0-9]+)? scope link")
+                .expect("Unable to compile regular expression");
+        }
+
+        let cap_str = String::from_utf8(output.stdout)?;
+        let cap = RE.captures(&cap_str);
         if let Some(cap) = cap {
             trace!("got link local IP of {} from device {}", &cap[1], &dev);
             Ok(cap[1].parse::<Ipv6Addr>()?)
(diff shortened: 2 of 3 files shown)


althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -43,6 +43,23 @@ impl KernelInterface {
         }
     }
 
+    pub fn get_global_device_ip_v4(&self, dev: &str) -> Result<Ipv4Addr, Error> {
+        let output = self.run_command("ip", &["addr", "show", "dev", dev, "scope", "global"])?;
+        trace!("Got {:?} from `ip addr`", output);
+
+        let re = Regex::new(r"inet (\S*?)(/[0-9]+)? scope global").unwrap();
```
drozdziak1: Put this Regex into a `lazy_static` block so that it doesn't need to be recompiled on each call.

althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -43,6 +43,23 @@ impl KernelInterface {
...
+        let str = String::from_utf8(output.stdout)?;
+        let cap = re.captures(&str);
```
drozdziak1: Can't the RHS just be on the `if` line below?

althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -43,6 +43,23 @@ impl KernelInterface {
         }
     }
 
+    pub fn get_global_device_ip_v4(&self, dev: &str) -> Result<Ipv4Addr, Error> {
+        let output = self.run_command("ip", &["addr", "show", "dev", dev, "scope", "global"])?;
+        trace!("Got {:?} from `ip addr`", output);
+
+        let re = Regex::new(r"inet (\S*?)(/[0-9]+)? scope global").unwrap();
+        let str = String::from_utf8(output.stdout)?;
```
drozdziak1: Rust is clever enough to understand this, but using `str` is pretty risky. Even the syntax highlighting assumes it to be the keyword for the type and not a local var name you can use.

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -269,11 +272,23 @@ impl Handler<Tick> for ExitManager {
...
             if let Some(ref general_details) = exit.info.general_details() {
                 trace!("We have details for the selected exit!");
-                if let Some(_) = exit.info.our_details() {
+                // only run if we have our own details and we either have no setup exit or the chosen
+                // exit has changed
+                if exit.info.our_details().is_some()
+                    && !(self.last_exit.is_some() && self.last_exit.clone().unwrap() == exit)
+                {
+                    // fully cleanup exit tunnel before making a change
+                    match KI.del_interface("wg_exit") {
```
drozdziak1: If you mean for a full cleanup, don't you need to also remove the default route?

jkilpatr: true I suppose you could modfiy the gateway ip.

althea_kernel_interface/src/link_local_tools.rs
jkilpatr: RHS?

althea_kernel_interface/src/link_local_tools.rs
jkilpatr: yeha changed that.

althea_kernel_interface/src/link_local_tools.rs
mpapierski: +1

althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -43,6 +43,23 @@ impl KernelInterface {
...
+        let cap = re.captures(&str);
+        if let Some(cap) = cap {
+            trace!("got global IP of {} from device {}", &cap[1], &dev);
+            Ok(cap[1].parse::<Ipv4Addr>()?)
```
mpapierski: Rust should be able to infer type i.e `Ok(cap[1].parse()?)` should work AFAIK.

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -58,7 +58,7 @@ impl KernelInterface {
             ],
         )?;
 
-        let prev_ip = self.get_global_device_ip("wg_exit");
+        let prev_ip: Result<Ipv4Addr, Error> = self.get_global_device_ip_v4("wg_exit");
```
mpapierski: Rust should be able to infer type here too AFAIK.

jkilpatr: yes I did that for the sake of making sure I don't screw it up again.

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -269,11 +272,17 @@ impl Handler<Tick> for ExitManager {
...
-                if let Some(_) = exit.info.our_details() {
+                // only run if we have our own details and we either have no setup exit or the chosen
+                // exit has changed
+                if exit.info.our_details().is_some()
+                    && !(self.last_exit.is_some() && self.last_exit.clone().unwrap() == exit)
+                {
                     trace!("We are signed up for the selected exit!");
                     linux_setup_exit_tunnel().expect("failure setting up exit tunnel");
                     TrafficWatcher::from_registry()
                         .do_send(Watch(exit.id.clone(), general_details.exit_price));
```
kingoflolz: This should run every time, as traffic needs to be settled every 5 seconds, not only when exits change

jkilpatr:
@drozdziak1 fixed

PR #270, opened 2018-09-18
Port toggling

jkilpatr:
- [x] testing

- [x] documentation

- [ ] features to handle identifying device specific port orders


rita/src/rita_client/dashboard/interfaces.rs
```diff
@@ -0,0 +1,518 @@
...
+        ifname,
+        setting_name
+    );
+
+    // Match parent section name
+    Ok(match &setting_name.replace(".ifname", "") {
+        s if s.contains("rita_") => InterfaceMode::Mesh,
+        s if s.contains("lan") => InterfaceMode::LAN,
+        s if s.contains("backhaul") => InterfaceMode::WAN,
+        other => bail!(
```
drozdziak1: This is very inconsiderate, I must've left it in by accident. This should ideally work the same as `wlan2mode` -  log via `warn!` and return `InterfaceMode::Unknown`

rita/src/rita_client/dashboard/interfaces.rs
```diff
@@ -0,0 +1,518 @@
...
+            let ret = KI.set_uci_var(&format!("network.rita_{}.proto", ifname), "static");
+            return_codes.push(ret);
+            mesh_add = true;
+        }
+        InterfaceMode::Unknown => unimplemented!(),
+    }
+
+    // check all of our return codes in order to handle any possible issue
+    let mut error_occured = false;
+    for ret in return_codes {
```
mpapierski: I think it could be rewritten as something like `return_codes.map(|r| r.is_ok()).any()` or something similiar.

rita/src/rita_client/dashboard/nodeinfo.rs
```diff
@@ -0,0 +1,107 @@
...
+                .from_err()
+                .and_then(|res| {
+                    let stream = TcpStream::connect::<SocketAddr>(
+                        format!("[::1]:{}", SETTING.get_network().babel_port).parse()?,
+                    )?;
+                    let mut babel = Babel::new(stream);
+                    babel.start_connection()?;
+                    let route_table_sample = babel.parse_routes()?;
+
+                    let res = res?;
```
mpapierski: I'd consier doing `res?` at the beggining of the function to return early with potential error. This would save on doing TCP connection, and babel instance creation.

rita/src/rita_client/dashboard/nodeinfo.rs
```diff
@@ -0,0 +1,107 @@
...
+                                    nickname: serde_json::to_string(&identity.mesh_ip).unwrap(),
+                                    route_metric_to_exit: u16::max_value(),
+                                    total_payments: debt_info.total_payment_received.into(),
+                                    debt: debt_info.debt.clone().into(),
+                                    link_cost: u16::max_value(),
+                                    price_to_exit: u32::max_value(),
+                                });
+                                continue;
+                            }
+                            let route = maybe_route?;
```
mpapierski: At this point `maybe_route.unwrap()` is probably safe call since in case of error we're doing a continue, so at this point we're sure that `maybe_route().is_err() == false`

althea_kernel_interface/src/manipulate_uci.rs
```diff
@@ -47,19 +50,20 @@ impl KernelInterface {
...
-        let output = self.run_command("uci", &["show", &key])?;
+        let output = self.run_command("uci", &["get", &key])?;
         if !output.stderr.is_empty() {
             return Err(KernelInterfaceError::RuntimeError(format!(
                 "received error while getting UCI: {}",
                 String::from_utf8(output.stderr)?
             )).into());
         }
-        Ok(String::from_utf8(output.stdout)?)
+        let clean_string = String::from_utf8(output.stdout)?.replace("\n", "");
```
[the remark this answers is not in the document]
jkilpatr: I suppose we could have an encapsulated value and then strip internal newlines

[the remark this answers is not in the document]
jkilpatr: turns out rust has trim just for this. https://doc.rust-lang.org/std/string/struct.String.html#method.trim

althea_kernel_interface/src/manipulate_uci.rs
```diff
@@ -69,6 +73,18 @@ impl KernelInterface {
...
+    //Resets unsaved changes to UCI
+    pub fn uci_revert(&self, section: &str) -> Result<bool, Error> {
+        let output = self.run_command("uci", &["revert", section])?;
+        if !output.status.success() {
+            return Err(KernelInterfaceError::RuntimeError(format!(
+                "received error while reverting UCI: {}",
+                String::from_utf8(output.stderr)?
+            )).into());
+        }
+        Ok(true)
```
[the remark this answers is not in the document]
jkilpatr: yeha that's pretty old, but I'll fix it.

althea_kernel_interface/src/manipulate_uci.rs
```diff
@@ -80,4 +96,53 @@ impl KernelInterface {
...
+        }
+
+        let stdout = String::from_utf8(output.stdout)?;
+
+        let mut retval: HashMap<String, String> = HashMap::new();
+
+        for line in stdout.lines() {
+            let caps = match RE.captures(line) {
+                Some(c) => c,
+                None => bail!(
```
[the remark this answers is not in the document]
jkilpatr: https://docs.rs/failure/0.1.1/failure/macro.bail.html yes

rita/src/rita_client/dashboard/interfaces.rs
```diff
@@ -0,0 +1,518 @@
...
+    // we assume wlan0 => radio0 this is held true by our config
+    // modifications but is not generally true for OpenWRT
+    let radio = match ifname.chars().last() {
+        Some(character) => format!("radio{}", character),
+        None => bail!("Invalid interface name {:?}", ifname),
+    };
+    let network_section = format!("default_{}", radio);
+
+    match a {
+        InterfaceMode::WAN => unimplemented!(),
```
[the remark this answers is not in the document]
jkilpatr: we gate WAN and unknown at the start of the function so I guess I could just put {} there.

rita/src/rita_client/dashboard/nodeinfo.rs
jkilpatr: good point.

rita/src/rita_client/dashboard/interfaces.rs
jkilpatr: sure it could, but it would be harder to read and no more efficient.

PR #275, opened 2018-09-19
Deps update

jkilpatr:
Pretty aggressive update of all dependencies, things seem to still work but I would test more before merging.


mpapierski:
IIRC there was some strange problem with docopt last time we upgraded it. Our executable didn't run on exit nodes but was fine locally for some reason. (btw I learned that it doesn't matter if we'd put `foo = "1.0.42"` or `foo = "1.0"` - both would get resolved to the latest `1.0.X` version on next `cargo update`)

drozdziak1:
Looks good, but needs a rebase

jkilpatr:
@drozdziak1 @mpapierski I did have to rollback docopt, looks fine otherwise though, land it if the tests pass

PR #277, opened 2018-09-21
Add endpoints for changing the mesh IP and change own_ip to mesh_ip

drozdziak1:
Stuff that's still missing:
* [x] Resolve the nasty futures compilation error
* [x] Docs
* [x] Remove the mesh IP from actors that hold it in their internal state

### DIFF
--- a/clu/src/lib.rs
+++ b/clu/src/lib.rs
@@ -4,10 +4,13 @@ extern crate log;
 #[macro_use]
 extern crate failure;
 
+#[macro_use]
+extern crate lazy_static;
+
 use std::net::IpAddr;
 
 extern crate settings;
-use settings::{NetworkSettings, RitaCommonSettings};
+use settings::RitaCommonSettings;
 
 extern crate ipgen;
 extern crate rand;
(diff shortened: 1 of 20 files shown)


clu/src/lib.rs
```diff
@@ -34,45 +37,36 @@ pub enum CluError {
...
-    config.wg_public_key = wg_public_key.to_string();
-
-    Ok(())
-}
-
-pub fn linux_generate_mesh_ip(config: &mut NetworkSettings) -> Result<(), Error> {
+pub fn linux_generate_mesh_ip() -> Result<IpAddr, Error> {
     let seed: String = thread_rng().sample_iter(&Alphanumeric).take(50).collect();
-    let mesh_ip = ipgen::ip(&seed, "fd00::/8").unwrap();
+    let mesh_ip = match ipgen::ip(&seed, "fd00::/8") {
```
mpapierski: Alternatively:

```rust
use failure::err_msg;
let mesh_ip = ipgen::ip(…).map_err(err_msg)?;
```

clu/src/lib.rs
```diff
@@ -92,21 +86,65 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
...
+                    existing_own_ip
+                );
+                network_settings.mesh_ip = Some(existing_own_ip);
+            } else {
+                warn!(
+                    "Existing compat own_ip value {} is invalid, generating a new mesh IP and migrating to mesh_ip",
+                    existing_own_ip
+                    );
+                network_settings.mesh_ip =
+                    Some(linux_generate_mesh_ip().expect("failed to generate a new mesh IP"));
```
mpapierski: Wouldn't it make sense here to propagate possible error here (and in similiar places) instead of a panic?

rita/src/rita_common/network_endpoints/mod.rs
```diff
@@ -79,17 +80,22 @@ pub fn hello_response(
...
-            Ok(Json(LocalIdentity {
-                global: SETTING.get_identity(),
-                wg_port: tunnel.0.listen_port,
-                have_tunnel: Some(tunnel.1),
-            }))
-        }).from_err()
-        .responder()
+    Box::new(
+        TunnelManager::from_registry()
+            .send(IdentityCallback::new(their_id, peer, None))
```
jkilpatr: move the from_err() up here right after the send. The issue is that you have a mailbox error (the sort of thing you get from a failed send) and you need to  convert it to a failure error to be consistent with what you return on line 91

clu/src/lib.rs
drozdziak1: This is the only place where the IP is generated, and without said IP there's no point running the application.

clu/src/lib.rs
```diff
@@ -92,21 +86,65 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
     cleanup()?;
     KI.restore_default_route(&mut config.get_network_mut().default_route)?;
 
-    let privkey = config.get_network().wg_private_key.clone();
-    let pubkey = config.get_network().wg_public_key.clone();
-    let mesh_ip = config.get_network().own_ip.clone();
+    let mut network_settings = config.get_network_mut();
```
jkilpatr: holding a write reference for the duration of this function kinda makes me nervous, but at the same time it's pre-start

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -235,6 +237,69 @@ pub fn set_interfaces(
...
+        },
+        None => {
+            let error_msg = format!("set_mesh_ip: \"mesh_ip\" not found in supplied JSON");
+            info!("{}", error_msg);
+            ret.insert("error".to_owned(), error_msg);
+        }
+    }
+
+    if let Err(e) = KI.run_command("/etc/init.d/rita", &["restart"]) {
+        return Box::new(future::err(e));
```
jkilpatr: have you tested this? how did that go?

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -179,7 +179,14 @@ pub fn exit_setup_request(
     reg_details.email_code = code;
 
     let ident = ExitClientIdentity {
-        global: SETTING.get_identity(),
+        global: match SETTING.get_identity() {
```
jkilpatr: are these possible any more? or should we panic?

rita/src/rita_client/dashboard/network_endpoints.rs
drozdziak1: Good question, actually OpenWRT happily restarted Rita despite it being the calling process

clu/src/lib.rs
drozdziak1: So normally to me cases like this usually turn out to be a trade-off between atomicity of a bunch of code and the risk of hogging the lock unnecessarily. But here to me it makes sense to lean on the hogging side because the network settings section of the config is potentially useless without `clu` making sure everything's in place.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -477,7 +480,10 @@ fn contact_neighbor(peer: &Peer, our_port: u16) -> Result<(), Error> {
 
     let _res = HTTPClient::from_registry().do_send(Hello {
         my_id: LocalIdentity {
-            global: SETTING.get_identity(),
+            global: match SETTING.get_identity() {
```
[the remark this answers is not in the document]
drozdziak1: Nice, will do!

rita/src/rita_client/exit_manager/mod.rs
drozdziak1: Handled, went with `ok_or` as suggested by Michał

PR #280, opened 2018-09-24
Improve command-line UX

drozdziak1:
This commit removes the need to verbosely type out every option and
assumes sane defaults, including logging.

rita/src/client.rs:
rita/src/exit.rs:
* Give defaults to every option
* Clean up the docopt message
* Start on the info log level by default

### DIFF
--- a/rita/src/exit.rs
+++ b/rita/src/exit.rs
@@ -66,6 +66,7 @@ extern crate trust_dns_resolver;
 use settings::{RitaCommonSettings, RitaExitSettings, RitaExitSettingsStruct};
 
 use docopt::Docopt;
+use log::LevelFilter;
 #[cfg(not(test))]
 use settings::FileWrite;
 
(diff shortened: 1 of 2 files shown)


rita/src/exit.rs
```diff
@@ -166,7 +173,14 @@ fn main() {
     // On Linux static builds we need to probe ssl certs path to be able to
     // do TLS stuff.
     openssl_probe::init_ssl_cert_env_vars();
-    env_logger::init();
+
+    // Log at info level by default
+    match env::var("RUST_LOG") {
+        Ok(_) => env_logger::init(),
+        Err(_) => env_logger::Builder::new()
```
jkilpatr: no can do, start on ERR even warn fills router storage in minutes.

jkilpatr: I suppose we can change the setting on the routers by updating the startup script with the package, thoughts?

drozdziak1: @jkilpatr err is the default

drozdziak1:
It may be a good idea to **hold off on merging this** until #275 is in so that touching the CLI can also involve having the new version of `docopt` work with our code.

PR #282, opened 2018-09-24
Tunnel GC did not unmonitor Tunnels

jkilpatr:
Tunnel GC would leave junk in Babel about tunnels that no longer
exist. A new tunnel might then be added on the same wg id and
end up monitored when perhaps it should not be.


rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -345,6 +345,20 @@ impl Message for TriggerGC {
...
+            Ok(stream) => stream,
+            Err(e) => {
+                warn!("Tunnel GC failed to open babel stream with {:?}", e);
+                return Err(e);
+            }
+        };
+        let mut babel = Babel::new(stream);
+        let res = babel.start_connection();
+        if res.is_err() {
+            warn!("Failed start Babel RPC connection! {:?}", res);
```
drozdziak1: A gremlin stole the "to"

PR #283, opened 2018-09-24
Add wifi settings endpoint back

jkilpatr:
I should not have deleted GET wifi settings, which is actively used
along with POST wifi settings which was removed and no longer used.

### DIFF
--- a/rita/src/rita_client/dashboard/network_endpoints.rs
+++ b/rita/src/rita_client/dashboard/network_endpoints.rs
@@ -10,7 +10,7 @@ use althea_types::ExitState;
 use rita_client::dashboard::exitinfo::{ExitInfo, GetExitInfo};
 use rita_client::dashboard::interfaces::{GetInterfaces, InterfaceMode, InterfaceToSet};
 use rita_client::dashboard::nodeinfo::{GetNodeInfo, NodeInfo};
-use rita_client::dashboard::wifi::{WifiPass, WifiSSID};
+use rita_client::dashboard::wifi::{GetWifiConfig, WifiInterface, WifiPass, WifiSSID};
 use rita_client::exit_manager::exit_setup_request;
 use rita_common::dashboard::Dashboard;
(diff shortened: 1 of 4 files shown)


rita/src/rita_client/dashboard/wifi.rs
```diff
@@ -93,3 +96,51 @@ impl Handler<WifiPass> for Dashboard {
...
     }
 }
+
+pub struct GetWifiConfig;
+
+impl Message for GetWifiConfig {
+    type Result = Result<Vec<WifiInterface>, Error>;
+}
+
+impl Handler<GetWifiConfig> for Dashboard {
```
drozdziak1: Why exactly does this need to happen in the Dashboard actor? This message would generally clutter `Dashboard`'s mailbox while the data could be perfectly obtained from the request handler

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -256,3 +256,14 @@ fn validate_config_value(s: &str) -> Result<(), ValidationError> {
         Ok(())
     }
 }
+
+pub fn get_wifi_config(
```
drozdziak1: [OCD Tier] You're calling the endpoint `/wifi_settings`, why is it `[...]_wifi_config` all of the sudden? Consistency helps keep things organized, having redundant names for the same (or closely coupled) thing is avoidably adding to the complexity of our code.

rita/src/rita_client/dashboard/wifi.rs
jkilpatr: it's this or we put everything into the network endpoints receiver functions.  Might be better, but not by much.

rita/src/rita_client/dashboard/network_endpoints.rs
jkilpatr: this is how it was named before I ever touched it. Not sure it's worth trying to fix right now.

PR #287, opened 2018-09-26
Add timeouts to the main loop futures

jkilpatr:
This is pretty much the nuclear anti-lag option. Not sure if it's
unsafe in any way, but I have to guess it is in at least one way.

### DIFF
--- a/rita/src/rita_client/rita_loop/mod.rs
+++ b/rita/src/rita_client/rita_loop/mod.rs
@@ -42,6 +42,7 @@ impl Handler<Tick> for RitaLoop {
             ExitManager::from_registry()
                 .send(Tick {})
                 .into_actor(self)
+                .timeout(Duration::from_secs(5), MailboxError::Timeout)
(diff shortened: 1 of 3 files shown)


rita/src/rita_client/rita_loop/mod.rs
```diff
@@ -42,6 +42,7 @@ impl Handler<Tick> for RitaLoop {
             ExitManager::from_registry()
                 .send(Tick {})
                 .into_actor(self)
+                .timeout(Duration::from_secs(5), MailboxError::Timeout)
```
drozdziak1: This timeout should either be configurable or computed as e.g. 150% of the average time the Tick takes , the 5 seconds are as arbitrary as it gets, half a radix we daily, derived from the total finger count humans have.

jkilpatr: I tried an arbitrary value to test, but this plain doesn't work, I'll go looking at the locks.

PR #297, opened 2018-10-01
Set LAN interface even if none exist

asoltys:
Fixes an issue where if you had removed all LAN interfaces you wouldn't be
able to re-add one since the network.lan.ifname setting wouldn't exist and the
code would return an Err in that case.

Now if the setting doesn't exist, we go ahead and set it instead of returning 
an Err.

### DIFF
--- a/rita/src/rita_client/dashboard/interfaces.rs
+++ b/rita/src/rita_client/dashboard/interfaces.rs
@@ -268,8 +268,14 @@ pub fn ethernet_transform_mode(
                     return_codes.push(ret);
                 }
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


drozdziak1:
@asoltys Sounds good, I'd appreciate if you could adjust this PR for that. Thanks!

asoltys:
I moved the string parsing into the Kernel but decided that rather than adding a new error type to return it makes more sense to just return Ok() with an empty string if the uci setting wasn't found.

jkilpatr:
also why where there three commits that undo each other?  Just cherry pick the second one and lets go for it.

asoltys:
Ok, rebased

Issue #279, opened 2018-09-22
Exits should indicate if an email is or is not mandatory


jkilpatr:
Just need to fix tests for this one.

Right now it's possible to have exits that don't need emails, but the UI will always assume that one is needed because it can't be sure without trying. 
The exit should include in it's exit details struct if an email is or is not required. This will then be exposed from the exits endpoint to the UI which can take advantage of it. 

this is pretty low priority. Might be a good task for @asoltys rust wise if he has free time.

jkilpatr:
just add 'needs email' to the exit details struct.

drozdziak1:
Deadline estimated to **Thursday afternoon**

PR #298, opened 2018-10-02
[WIP]Add an enum for specifying the exit verification mode

drozdziak1:
* [x] - enum for specifying the verification mode
* [ ] - docs update
* [ ] - Make the enum value a global setting

### DIFF
--- a/althea_types/src/interop.rs
+++ b/althea_types/src/interop.rs
@@ -124,12 +124,20 @@ pub struct ExitClientIdentity {
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub enum ExitVerifMode {
+    Email,
+    Off,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct ExitDetails {
     pub server_internal_ip: IpAddr,
     pub netmask: u8,
     pub wg_exit_port: u16,
     pub exit_price: u64,
     pub description: String,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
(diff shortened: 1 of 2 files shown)


drozdziak1:
This solves #279

althea_types/src/interop.rs
```diff
@@ -123,13 +123,21 @@ pub struct ExitClientIdentity {
...
+
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct ExitDetails {
     pub server_internal_ip: IpAddr,
     pub netmask: u8,
     pub wg_exit_port: u16,
     pub exit_price: u64,
     pub description: String,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub verif_mode: Option<ExitVerifMode>,
```
mpapierski: Do you need Option here if it will be always `Some`? What would `None` mean? Or maybe there is no need to have `ExitVerifMode::Off` if `verif_mode.is_none()` means the verification is off?

jkilpatr: none is the current state (since we don't say right now) for backwards compat

rita/src/rita_exit/db_client/mod.rs
```diff
@@ -140,6 +140,7 @@ pub fn get_exit_info() -> ExitDetails {
         exit_price: SETTING.get_exit_network().exit_price,
         netmask: SETTING.get_exit_network().netmask,
         description: SETTING.get_description(),
+        verif_mode: Some(ExitVerifMode::Off),
```
jkilpatr: cool now we need to set this for when we send it out to the client.

PR #302, opened 2018-10-04
Report device name

asoltys:
Adds an optional string device field to the network settings in the config file if it doesn't already exist by reading the device name from /etc/althea-firmware-release

Report the device name in the common dashboard /info endpoint

### DIFF
--- a/clu/src/lib.rs
+++ b/clu/src/lib.rs
@@ -151,7 +151,12 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
                 if line.starts_with("device:") {
                     let mut array = line.split(" ");
                     array.next();
-                    network_settings.device = array.next().map(|s| s.to_string());
+                    let device = array.next();
+
+                    if device.is_some() {
+                        network_settings.device = Some(device.unwrap().to_string());
+                    }
+
                     break;
                 }
             }
--- a/clu/src/lib.rs
+++ b/clu/src/lib.rs
@@ -138,12 +138,13 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
            info!("Device name is {}", existing_device); 
         } 
         None => {
-            info!("No device name was found, reading from /etc/althea-firmware-release");
+            let release_file_path = "/etc/althea-firmware-release";
+            info!("No device name was found, reading from {}", release_file_path);
 
             let mut contents = String::new();
-            match File::open("/etc/althea-firmware-release") {
+            match File::open(release_file_path) {
                 Ok(mut f) => { f.read_to_string(&mut contents)?; },
-                Err(e) => warn!("Couldn't open /etc/althea-firmware-release: {}", e),
+                Err(e) => warn!("Couldn't open {}: {}", release_file_path, e),
             };
 
             for line in contents.lines() {
@@ -156,7 +157,7 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
             }
 
             if network_settings.device.is_none() {
-                warn!("Device name could not be read from /etc/althea-firmware-release");
+                warn!("Device name could not be read from {}", release_file_path);
             } 
         },
     }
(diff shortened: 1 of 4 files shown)


rita/src/rita_common/dashboard/mod.rs
```diff
@@ -69,10 +72,17 @@ impl Handler<GetOwnInfo> for Dashboard {
                         let balance = balance
                             .checked_div(&Int256::from(1_000_000_000i64))
                             .ok_or(OwnInfoError::RoundDownError(balance.clone()))?;
+ 
+                        let device = match SETTING.get_network().device.clone() {
```
mpapierski: Shorter: `let device = SETTING.get_network().device.clone().unwrap_or(String::new())` (I think if `device` is a `String` already you don't need to call .to_string()). Also, wouldn't it make more sense from the frontend side to send "null" value when its missing?

clu/src/lib.rs
```diff
@@ -130,6 +133,32 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
...
+            for line in contents.lines() {
+                if line.starts_with("device:") {
+                    let mut array = line.split(" ");
+                    array.next();
+                    match array.next() {
+                        Some(device) => {
+                            info!("Setting device name to {}", device);
+                            network_settings.device = Some(device.to_string());
+                        },
+                        None => warn!("Unable to get device from /etc/althea-firmware-release"),
```
mpapierski: Wouldn't it make sense to break from this loop in this case? Is this file guaranteed to be present, and when we can't parse it it is a serious error?

jkilpatr: this file is pretty much garunteed to be present, but not finding it should not be a serious error. Also yes we should break the for loop once we find it.

asoltys:
Ok, I made device an Option<String> so it will be null if it's not found and if the althea-firmware-release file is not found the error is handled and we get a warning rather than a crash. Also we break out of the loop as soon as the "device: xxxx" line is found now.

clu/src/lib.rs
```diff
@@ -139,23 +139,25 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
...
                     array.next();
-                    match array.next() {
-                        Some(device) => {
-                            info!("Setting device name to {}", device);
-                            network_settings.device = Some(device.to_string());
-                        },
-                        None => warn!("Unable to get device from /etc/althea-firmware-release"),
-                    }
+                    network_settings.device = array.next().map(|s| s.to_string());
+                    break;
```
jkilpatr: why not use a single to_string? map indicates there's more htan one object but there shouldn't be.

clu/src/lib.rs
```diff
@@ -130,6 +133,45 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
...
+            match File::open(release_file_path) {
+                Ok(mut f) => {
+                    f.read_to_string(&mut contents)?;
+                }
+                Err(e) => warn!("Couldn't open {}: {}", release_file_path, e),
+            };
+
+            for line in contents.lines() {
+                if line.starts_with("device:") {
+                    let mut array = line.split(" ");
```
drozdziak1: @asoltys (You'll want to ignore this and see the later nit instead, but it's good practice anyway) It's not an array, but an `Iterator`, this could get very confusing!

clu/src/lib.rs
```diff
@@ -130,6 +133,45 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
...
+                    array.next();
+                    let device = array.next();
```
drozdziak1: Looks like you just want the second element, in that case just go `let device = line.split(" ").nth(2).ok_or(format_err!("Could not obtain device name from line {:?}", line))?`. Here's what it does:
* `nth(arg)` does `next()` `arg+1` times and gives you the `Option` from the last `next()`
* `ok_or(arg)` converts the `Option` from `nth` to a `Result`. It'll return `Ok(val)` if the `Option` was `Some(whatever)` and `Err(arg)` otherwise. The `arg` is `ok_or`'s argument.
* `format_err!()` macro is a helper that'll build an error type from whatever formatted string you supply it with.
*  `?` after an expression does `return Err(something)` if the result of the expression was `Err(something)` and if it was an `Ok(something_else` it just unwraps it into just `something_else`. Analogically with `Option`'s but with `Some` and `None` in place of `Ok` and `Err`, respectively.

PR #304, opened 2018-10-05
Add an enum for specifying the exit verification mode

drozdziak1:
* [x] - The enum for choosing the verification mode
* [x] - Code for using the new enum
* [x] - Sane, extensible config scheme
* [x] - docs

### DIFF
--- a/clu/src/lib.rs
+++ b/clu/src/lib.rs
@@ -254,7 +254,9 @@ fn linux_exit_init(config: Arc<RwLock<settings::RitaExitSettingsStruct>>) -> Res
     drop(network_settings);
 
     // Migrate compat mailer settings. This is put in this particular spot so that the network
-    // settings lock can be dropped
+    // settings lock can be dropped beforehand.
+    //
+    // TODO: REMOVE IN ALPHA 13 FROM HERE TILL THE Ok(())
     let compat_mailer_settings = config.get_mailer().clone();
     let verif_settings = config.get_verif_settings().clone();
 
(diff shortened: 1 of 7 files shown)


jkilpatr:
* check if mailer is true so that the exit serves up the right value
* have clients generate with verification true in their config and we can update remotely
 
that's really all that needs to happen here, nothing fancy.

clu/src/lib.rs
```diff
@@ -251,6 +251,32 @@ fn linux_exit_init(config: Arc<RwLock<settings::RitaExitSettingsStruct>>) -> Res
         &network_settings.wg_private_key,
     )?;
 
+    drop(network_settings);
```
jkilpatr: we can do this? That's pretty great.

clu/src/lib.rs
```diff
@@ -251,6 +251,32 @@ fn linux_exit_init(config: Arc<RwLock<settings::RitaExitSettingsStruct>>) -> Res
...
+
+    match verif_settings.clone() {
+        Some(_settings) => match compat_mailer_settings {
+            Some(_) => {
+                info!("Both verif_settings and compat settings exist, removing compat settings.");
+                *config.get_mailer_mut() = None;
+            }
+            None => {}
+        },
+        None => match compat_mailer_settings {
```
jkilpatr: mark for removal in alpha 13

jkilpatr:
@drozdziak1  test failures?

PR #305, opened 2018-10-05
Add support for meshpoint in interfaces

jkilpatr:
This removes support for the legacy adhoc mode on wireless interfaces
and replaces it with dual purpose 'meshpoint' the reason it needs a
seperate mode from 'mesh' is that meshpoint is not exclusive the LAN
wifi will continue to operate while meshpoint is enabled.

This pull adds handling for 'Meshpoint' as well as removes support
for adhoc mode. Adhoc mode requires setting a significant number of
vars at runtime, while meshpoint can be configured entierly at firmware
build time and simply toggled with a much lower chance of failure.


jkilpatr:
this should wait on the next release. As adam will have to do work to support it.

PR #311, opened 2018-10-15
Exit addition

drozdziak1:
* [x] Local addition via POST `/exits`
* [x] Hauling an exit list from a remote HTTPS server
* [x] Docs

### DIFF
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -70,6 +70,75 @@ This file documents the dashboard API found in Rita client.
 ## /exits
 
 - URL: `<rita ip>:<rita_dashboard_port>/exits'
+- Comment: Merges a supplied exit list with the existing list; existing entries
+  are overwritten
+- Method: `POST`
+- URL Params: `None`
+- Data Params: A JSON object containing the exits we want to add, e.g.:
+```json
+{
+	"exit_other": {
+		"auto_register": false,
+		"description": "",
+		"general_details": {
+			"description": "EDITEDITjust a normal althea exit",
+			"exit_price": 50,
+			"netmask": 24,
+			"server_internal_ip": "172.168.1.254",
+			"wg_exit_port": 59999
+
+		},
+		"id": {
+			"eth_address": "0x0101010101010101010101010101010101010101",
+			"mesh_ip": "fd00::5",
+			"wg_public_key": "KaTbsJ0Hur4D7Tcb+nc8ofs7n8tKL+wWG3H38KFCwlE="
+
+		},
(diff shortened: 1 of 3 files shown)


rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -370,3 +372,102 @@ pub fn remote_logging_level(path: Path<String>) -> Box<Future<Item = HttpRespons
...
+}
+
+pub fn exits_sync(
+    list_url_json: Json<HashMap<String, String>>,
+) -> Box<Future<Item = HttpResponse, Error = Error>> {
+    debug!("/exits/sync hit with {:?}", list_url_json);
+
+    let list_url = match list_url_json.get("url") {
+        Some(url) if url.starts_with("https://") => url,
+        Some(_unsafe_url) => {
```
mpapierski: Wouldn't it make sense to match "http://", and another match that would say that the exit url is invalid?

```suggestion
        Some(url) if url.starts_with("http://") => {
```

rita/src/rita_client/dashboard/network_endpoints.rs
```diff
@@ -370,3 +372,102 @@ pub fn remote_logging_level(path: Path<String>) -> Box<Future<Item = HttpRespons
...
+
+                return Box::new(future::ok(
+                    HttpResponse::new(StatusCode::BAD_REQUEST)
+                        .into_builder()
+                        .json(ret),
+                ));
+            }
+        },
+        Err(e) => {
+            let mut ret = HashMap::<String, String>::new();
```
mpapierski: I think in this case rust can *probably* will infer type of HashMap, and other similiar cases.

```suggestion
            let mut ret = HashMap::new();
```

docs/api/router-dashboard.md
```diff
@@ -106,6 +175,45 @@ This file documents the dashboard API found in Rita client.
...
+  - `500 Internal Server Error` when the request fails
+```json
+{
+  "error": "<description>",
+  "rust_error": "<stringified_rust_error>"
+}```
+
+- Sample Call:
+
+`curl 127.0.0.1:4877/exits/borked/reset -H "Content-Type:
```
mpapierski: Accidentally wrapped in the middle of command?

drozdziak1: yep

rita/src/rita_client/dashboard/network_endpoints.rs
drozdziak1: That's true, needed this for a dummy version, but that's not the case anymore

PR #318, opened 2018-10-18
Run cross builds for supported architectures.

### LINKED ISSUE #316: Add cross arch tests to Rita
Rita needs the cross arch tests that are run on Clarity and Gauc to identify issues with the tests on mips and other platforms.

mpapierski:
This is mostly copy and paste of boilerplate code from clarity/guac/serde-rlp etc.

Closes #316

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,29 +7,40 @@ before_install:
   - sudo apt-get install -y libsqlite3-dev iperf3 python3-pip bridge-utils wireguard linux-source linux-headers-$(uname -r) curl git libssl-dev pkg-config build-essential ipset
   - which diesel || cargo install diesel_cli --no-default-features --features sqlite
   - sudo cp $(which diesel) /usr/bin
+  - command -v cross 1>/dev/null || cargo install cross
+
 env:
-  - TEST_COMMAND="rustup component add rustfmt-preview --toolchain stable && cargo fmt --all && git status && if ! git diff-index --quiet HEAD --; then exit 1; fi"
-  - TEST_COMMAND="cargo build --locked --verbose --all"
-  - TEST_COMMAND="cargo test --locked --verbose --all" RUST_TEST_THREADS=1
+  global:
+    - RUST_TEST_THREADS=1
+  matrix:
+    - TARGET="x86_64-unknown-linux-gnu"
+    - TARGET="mips-unknown-linux-gnu"
+    - TARGET="mipsel-unknown-linux-gnu"
+    - TARGET="mips64-unknown-linux-gnuabi64"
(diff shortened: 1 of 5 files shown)


clu/src/lib.rs
```diff
@@ -295,6 +295,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn test_generate_wg_key() {
```
mpapierski: Its not pretty but IMO `create_wg_keypair` doesn't go through the usual "mock" interface of KernelInterface like other methods. It just calls `Command::new`. It would make more sense to use the other interface to verify that `wg genkey` and `wg pubkey` is called.

.travis.yml
```diff
@@ -7,29 +7,38 @@ before_install:
   - sudo apt-get install -y libsqlite3-dev iperf3 python3-pip bridge-utils wireguard linux-source linux-headers-$(uname -r) curl git libssl-dev pkg-config build-essential ipset
   - which diesel || cargo install diesel_cli --no-default-features --features sqlite
   - sudo cp $(which diesel) /usr/bin
+  - command -v cross 1>/dev/null || cargo install cross
```
jkilpatr: so we're ignoring the stdout of this command? won't travis dot hat for us since it's part of the pre setup?

clu/src/lib.rs
jkilpatr: well I'm the guilty party that wrote that test, at the time I was interested to see if my stdin/out shenanigans worked for real and not just with the mocker.

rita/build.rs
```diff
@@ -1,9 +1,11 @@
 use std::process::Command;
+
 fn main() {
```
jkilpatr: why did this end up being changed?

.travis.yml
mpapierski: This is a trick to check if cross is already installed and if its not installed, it will be. Travis caches ~/.cargo dir, so it might happen that cross binary already exists

rita/build.rs
mpapierski: Docker images shipped with cross doesn't have git installed. I remember I reported it somewhere in cross repos, but can't find it now. This shouldn't be a problem as long as we're not doing deployments using cross. There is possiblity to specify own images for architectures so we could install binaries. Anyway I think we should be fine.

clu/src/lib.rs
mpapierski: Saved for posterity: https://github.com/althea-mesh/althea_rs/issues/320

mpapierski:
Oh well all architectures are working just that `cross-build.sh` fails with weird missing GCC error.

PR #323, opened 2018-10-22
Make Rita serve client dashboard on HTTPS

drozdziak1:
TODO:
* [x] Make key location configurable
* [x] Update curl calls in dashboard docs


settings/src/lib.rs
```diff
@@ -147,6 +147,10 @@ pub struct NetworkSettings {
     /// The name of the device or router model
     #[serde(skip_serializing_if = "Option::is_none")]
     pub device: Option<String>,
+    /// The DER-encoded key used for HTTPS
```
jkilpatr: do we want to skip if none here? There's no default value so it will break if not configured right now.

rita/src/middleware.rs
```diff
@@ -20,10 +26,27 @@ impl<S> Middleware<S> for Headers {
...
-            header::HeaderName::try_from("Access-Control-Allow-Origin").unwrap(),
-            header::HeaderValue::from_str(&format!("http://{}", url_no_port)).unwrap(),
-        );
+
+        let (key, cert) = {
+            let settings = SETTING.get_network();
+            (settings.key_path.clone(), settings.cert_path.clone())
+        };
+
+        if key.is_some()
```
jkilpatr: well this if statement will work by virtue of crashing if key or cert are none. Simple check to add

rita/src/client.rs
```diff
@@ -262,13 +264,74 @@ fn main() {
...
+                    key, cert
+                );
+                let mut builder = SslAcceptor::mozilla_intermediate(SslMethod::tls()).unwrap();
+                builder
+                    .set_private_key_file(
+                        SETTING
+                            .get_network()
+                            .key_path
+                            .clone()
+                            .unwrap_or("/etc/uhttpd.key".to_owned()),
```
jkilpatr: ok this is not a good way to do defaults. Just crash if the configured path is not good. We should separate bootstrapping from configuration concerns.

jkilpatr:
@drozdziak1 looks like integration tests are still failing.

PR #325, opened 2018-10-25
Fix exit tunnel setup cases

jkilpatr:
Previously we set the exit tunnel up every 5 seconds, I removed that and instead properly remembered system state, this adds back the minimum amount of checking and changes required to prevent problems with rouge dhcp reassignments and such.


jkilpatr:
resolves https://github.com/althea-mesh/althea_rs/issues/324

PR #329, opened 2018-10-26
Rita price-setting

drozdziak1:
This PR enables Rita to set the local fee and the metric factor at
startup using clu.

babel_monitor/src/lib.rs:
* Add methods for setting the local fee and metric factor in Babel

clu/src/lib.rs:
* Set local fee and metric factor on startup
* Issue warnings if the local fee or metric factor is 0 (fee=0 means
free bandwidth for everyone and factor=0 means we care solely about
price)

rita/src/rita_common/dashboard/mod.rs:
* Return local fee and metric factor in GetOwnInfo

settings/src/lib.rs:
* Add local_fee and metric_factor settings, getters, and defaults for
them


babel_monitor/src/lib.rs
```diff
@@ -81,7 +81,7 @@ pub struct Neighbor {
 }
 
 pub struct Babel<T: Read + Write> {
-    stream: BufStream<T>,
+    pub stream: BufStream<T>,
```
mpapierski: Is `pub` necessary here?

drozdziak1: Oh you're right I don't

clu/src/lib.rs
```diff
@@ -188,6 +188,13 @@ fn linux_init(config: Arc<RwLock<settings::RitaSettingsStruct>>) -> Result<(), E
         &network_settings.wg_private_key,
     )?;
 
+    // Yield the mut lock
+    drop(network_settings);
```
jkilpatr: it's going to get dropped in like 5 lines and there's nothing that may stop it. Are you just trying to be explicit here?

rita/src/rita_common/dashboard/mod.rs
```diff
@@ -77,6 +78,7 @@ impl Handler<GetOwnInfo> for Dashboard {
                             balance: balance
                                 .to_i64()
                                 .ok_or(OwnInfoError::DownCastError(balance))?,
+                            local_fee: 0,
```
jkilpatr: why bother putting this here is it's going to be hardcoded?

jkilpatr:
oh yeha @drozdziak1  we need a way to adjust the adjustment factor.

rita/src/rita_common/dashboard/mod.rs
drozdziak1: Yup, left it by accident, thanks

settings/src/lib.rs
```diff
@@ -562,6 +588,14 @@ impl RitaCommonSettings<RitaSettingsStruct> for Arc<RwLock<RitaSettingsStruct>>
     fn set_future(&self, future: bool) {
         self.write().unwrap().future = future
     }
+
+    fn get_local_fee(&self) -> u32 {
+        self.read().unwrap().local_fee
```
jkilpatr: why are these here? we just check values directly in other cases

settings/src/lib.rs
```diff
@@ -86,6 +86,14 @@ fn default_tunnel_timeout() -> u64 {
     900 // 15 minutes
 }
 
+fn default_local_fee() -> u32 {
+    0
+}
+
+fn default_metric_factor() -> u32 {
+    1000u32
```
jkilpatr: um souldn't the factor be a float now that we're doing logs and float operations in babel?

settings/src/lib.rs
```diff
@@ -86,6 +86,14 @@ fn default_tunnel_timeout() -> u64 {
     900 // 15 minutes
 }
 
+fn default_local_fee() -> u32 {
+    0
```
jkilpatr: default fee 500k wei/byte

clu/src/lib.rs
jkilpatr: hmm typically I try and access/clone rather than doing a drop but I guess it's the same.

settings/src/lib.rs
drozdziak1: Currently the value corresponds to thousandths of the factor. I'm not using a `f64` so that the `Eq` derive still works (not defined for `f64` so that  people only compare them consciously I presume). I agree that what I did feels kinda complicated and I'm more than open to alternatives

Issue #177, opened 2018-08-07
In TunnelManager query the operating system for open ports rather than keeping track of them ourselves


jkilpatr:
would be faster, less error prone, and use less memory than our own setup.

PR #335, opened 2018-11-01
Query OS directly for free unused ports rather than directly managing them.

donkey-hotei:
This is my first go at contributing to Althea. Here's an attempt at solving #177 . Here I took the route of using a newer crate called [`netstat`](https://github.com/ivxvm/netstat-rs) to query the OS directly for unused UDP ports.

### DIFF
--- a/althea_kernel_interface/src/udp_socket_table.rs
+++ b/althea_kernel_interface/src/udp_socket_table.rs
@@ -1,0 +1,86 @@
+use super::KernelInterface;
+use std::fs::File;
+use std::io::prelude::*;
+use std::u16;
+
+use super::KernelInterfaceError;
+use failure::Error;
+
+/// Returns a kernel interface runtime error with the given message.
+fn runtime_error<T>(msg: &str) -> Result<T, Error> {
+    Err(KernelInterfaceError::RuntimeError(msg.to_string()).into())
+}
+
+/// Helper function for parsing out port number from local_address column
+fn parse_local_port(s: &str) -> Result<u16, Error> {
+    // second column in table contains local_address
+    let local_addr = match s.split_whitespace().nth(1) {
+        Some(addr) => addr,
+        None => return runtime_error("Error parsing local_address column!"),
+    };
+    // having a format like "00000000:14E9"
+    let port = match local_addr.split(":").nth(1) {
+        Some(port) => port,
+        None => return runtime_error("Error parsing local_address column!"),
+    };
+
+    match u16::from_str_radix(port, 16) {
+        Ok(port) => Ok(port),
+        Err(_) => runtime_error("Error parsing port from local_address column!"),
+    }
+}
+
+impl KernelInterface {
+    fn read_udp_socket_table(&self) -> Result<String, Error> {
+        let mut f = File::open("/proc/net/udp")?;
+        let mut contents = String::new();
+
+        f.read_to_string(&mut contents)?;
+
+        Ok(contents)
+    }
+
+    /// Returns list of ports in use as seen in the UDP socket table (/proc/net/udp)
+    pub fn used_ports(&self) -> Result<Vec<u16>, Error> {
+        let udp_sockets_table = self.read_udp_socket_table()?;
+        let mut lines = udp_sockets_table.split("\n");
+
+        lines.next(); // advance iterator to skip header
+
+        let ports: Vec<u16> = lines
+            .take_while(|line| line.len() > 0)  // until end of the table is reached,
+            .map(|line| parse_local_port(line)) // parse each udp port,
+            .filter_map(Result::ok) // only taking those which parsed successfully
+            .collect();
+
+        Ok(ports)
+    }
+}
+
+#[test]
+pub fn test_parse_local_port_on_valid_string_successful() {
+    let line = "1228: 00000000:14E9 00000000:0000 07 00000000:00000000 00:00000000 00000000  1000        0 34229668 2 ffff88007cb08800 0";
+
+    assert_eq!(parse_local_port(line).unwrap(), 5353)
+}
+
+#[test]
+pub fn test_parse_local_port_failure_on_totally_malformed_string() {
+    let line = "abcdefg";
+
+    assert!(parse_local_port(line).is_err())
+}
+
+#[test]
+pub fn test_parse_local_port_failure_on_malformed_local_address_column() {
+    let line = "1228: 00000000_14E9 00000000:0000 07 00000000:00000000 00:00000000 00000000  1000        0 34229668 2 ffff88007cb08800 0";
+
+    assert!(parse_local_port(line).is_err())
+}
+
+#[test]
+pub fn test_parse_local_port_failure_on_invalid_hex() {
+    let line = "1228: 00000000:FFFG 00000000:0000 07 00000000:00000000 00:00000000 00000000  1000        0 34229668 2 ffff88007cb08800 0";
+
+    assert!(parse_local_port(line).is_err())
+}
--- a/rita/src/rita_common/tunnel_manager/mod.rs
+++ b/rita/src/rita_common/tunnel_manager/mod.rs
@@ -220,7 +220,7 @@ impl Handler<IdentityCallback> for TunnelManager {
     fn handle(&mut self, msg: IdentityCallback, _: &mut Context<Self>) -> Self::Result {
         let our_port = match msg.our_port {
             Some(port) => port,
-            _ => match self.free_ports.pop() {
+            _ => match self.get_port() {
                 Some(p) => p,
                 None => {
                     warn!("Failed to allocate tunnel port! All tunnel opening will fail");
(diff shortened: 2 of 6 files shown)


rita/Cargo.toml
```diff
@@ -63,3 +63,4 @@ handlebars = "1.0.3"
 byteorder = { version = "1.2.6", features = ["i128"] }
 openssl-probe = "0.1.2"
 num-traits="0.2"
+netstat="0.7.0"
```
jkilpatr: looks like this requires some sort of native dependency to link against. I'll have to do more research about exactly what.

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+
+        let free_ports: Vec<u16> = (0..65535).filter(|s| !used_ports.contains(s)).collect();
+
+        PortEmissary { free_ports }
+    }
+}
+
+pub fn free_ports() -> Vec<u16> {
```
jkilpatr: ok, so every time we need a port we run this and grab it from the list?

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+use rita_common::port_emissary::netstat::*;
+
+struct PortEmissary {
+    free_ports: Vec<u16>,
+}
+
+impl PortEmissary {
+    fn new() -> PortEmissary {
+        let af_flags = AddressFamilyFlags::IPV4 | AddressFamilyFlags::IPV6;
+        let sockets_info = get_sockets_info(af_flags, ProtocolFlags::UDP).unwrap();
```
jkilpatr: if we're running this more than once at startup this unwrap needs to be handled.

jkilpatr:
actually on second glance it looks like a var issue not a deps one

jkilpatr:
error[E0658]: use of unstable library feature 'iterator_flatten' (see issue #48213)
  --> /home/justin/.cargo/registry/src/github.com-1ecc6299db9ec823/netstat-0.7.0/src/integrations/linux/api.rs:34:42
   |
34 |     Ok(attach_pids(iterators.into_iter().flatten()))
   |

donkey-hotei:
What version of Rust are you running? As of version 1.29 the `iterator_flatten` language feature was [stabilized](https://github.com/rust-lang/rust/blob/a7a60dc7f8e9a53feb70ee7aa05485345a045afb/src/libcore/iter/mod.rs#L2578).

rita/Cargo.toml
jkilpatr: is this still needed now?

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,45 @@
+use std::fs::File;
```
jkilpatr: I'd argue this goes as a Kernel interface module, as we (try to) set it up as a pluggable interface for different operating systems we may need to work with.

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::io::prelude::*;
+use std::u16;
+
+/// Helper function for parsing out port number from local_address column
+fn take_local_port(s: &str) -> u16 {
+    // second column in table contains local_address
+    let local_addr = s.split_whitespace().nth(1).unwrap();
```
jkilpatr: too many unwraps in this, what if you get invalid input? needs error handling and to return a result type so that we can do something other than crash the program when it goes wrong.

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    // having a format like "00000000:14E9"
+    let port = local_addr.split(":").nth(1).unwrap();
+
+    u16::from_str_radix(port, 16).unwrap()
+}
+
+/// Returns list of ports in use as seen in the UDP socket table (/proc/net/udp)
+fn used_ports() -> Vec<u16> {
```
jkilpatr: the result from take ports will probably filter up through here and then through tunnel manager until we gracefully print and drop it at the handler or Rita loop level. Crashes in Rita are a very big deal as we don't often have a large amount of environmental control or an easy way to restart Rita remotely. (actually a cron job to restart rita all the time if it's not running might be a good use of time...)

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    let mut udp_sockets_table = String::new();
+
+    f.read_to_string(&mut udp_sockets_table)
+        .expect("Error reading UDP socket table!");
+
+    let mut lines = udp_sockets_table.split("\n");
+
+    lines.next(); // advance iterator to skip header
+
+    let ports: Vec<u16> = lines
```
jkilpatr: very functional style here, why take while as opposed to filter? I think it's pretty much the same though.

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -3,22 +3,35 @@ use std::io::prelude::*;
 use std::u16;
 
 /// Helper function for parsing out port number from local_address column
-fn take_local_port(s: &str) -> u16 {
+fn take_local_port(s: &str) -> Result<u16, &'static str> {
```
jkilpatr: if you look in kernel interface you'll see our error enum that we use to wrap things like this, also this file should be moved over to kernel interface.

rita/src/rita_common/port_emissary/mod.rs
```diff
@@ -3,22 +3,35 @@ use std::io::prelude::*;
...
+    }
 }
 
 /// Returns list of ports in use as seen in the UDP socket table (/proc/net/udp)
-fn used_ports() -> Vec<u16> {
-    let mut f = File::open("/proc/net/udp").expect("UDP socket table not found!");
+fn used_ports() -> Result<Vec<u16>, &'static str> {
+    let mut f = match File::open("/proc/net/udp") {
+        Ok(file) => file,
+        Err(err) => return Err("UDP socket table not found!"),
```
jkilpatr: you could return err and pass through the error? also the err must be giving you an unused warning right now.

rita/src/rita_common/port_emissary/mod.rs
jkilpatr: ok that seems to make sense.

scripts/.git-hooks/pre-commit
```diff
@@ -3,18 +3,18 @@
...
-    if [ ${file: -3} == ".rs" ]; then
-        rustfmt --check $file 1>/dev/null;
+    if [ "${file: -3}" == ".rs" ]; then
+        rustfmt --check "$file" 1>/dev/null;
         if [ $? != 0 ]; then
             inconsistent_files+=($file);
         fi
     fi
-done < <(git diff --name-only --cached)
+done < <(git diff --name-only --cached --diff-filter=ACMRTUXB)
```
jkilpatr: don't think you intended to commit this.

althea_kernel_interface/src/udp_socket_table.rs
```diff
@@ -0,0 +1,75 @@
...
+        .take_while(|line| line.len() > 0)
+        .map(|line| parse_local_port(line))
+        .filter_map(Result::ok)
+        .collect();
+
+    Ok(ports)
+}
+
+/// Returns a list of all those ports not found in the UDP socket table.
+pub fn free_ports() -> Result<Vec<u16>, Error> {
```
jkilpatr: a thought, right now we will hand out a port then wait for a http response, during that time we may hand out another port, during this time the port is in the hello 'state' but is not bound, so what happens? I presume two callers will get the same port and then cause conflicts.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -513,11 +517,31 @@ fn contact_neighbor(peer: &Peer, our_port: u16) -> Result<(), Error> {
...
-            free_ports: ports,
-            tunnels: HashMap::new(),
+        let udp_table = match KI.used_ports() {
+            Ok(used_ports) => used_ports,
+            Err(_) => vec![],
+        };
+
+        let ports: HashMap<u16, bool> = (start..65535)
+            .into_iter()
+            .map(|port| (port, !udp_table.contains(&port)))
```
donkey-hotei: Since `udp_table` could be blank, this could be interpreted as _all ports being open_ which, if `used_ports()` failed then it could certainly not be the case.

rita/src/rita_common/tunnel_manager/mod.rs
```diff
@@ -526,8 +550,11 @@ impl TunnelManager {
     pub fn neighbor_inquiry_hostname(&mut self, their_hostname: String) -> Result<(), Error> {
         trace!("Getting tunnel, inq");
 
-        let our_port = match self.free_ports.pop() {
-            Some(p) => p,
+        let our_port = match self.port_query() {
+            Some(port) => {
+                self.ports.insert(port, false);
```
jkilpatr: why no have port_query mark the port it gives you as not free?

donkey-hotei: Borrowing `self.ports` as mutable would conflict with `is_free`'s immutability (as a `bool`). Then, even if I make a mutable reference when borrowing, there can only be one mutable reference to `self.ports` so the borrow checker would be violated if I `self.ports.insert(port, false)` in the same scope. Though, I'm still learning Rust - there's probably a way.

donkey-hotei:
Cleaned up in another PR with some git-fu and a cleaner definition of `get_ports` by @jkilpatr !

PR #359, opened 2018-11-18
The + is a valid char in wgkey parsing

jkilpatr:
This is causing errors in counting exit balances, it only seems to cause issues
with some keys so it's easy to see how it got missed during testing.


jkilpatr:
fixes https://github.com/althea-mesh/althea_rs/issues/354

PR #368, opened 2018-11-24
Introduce and Integrate WgKey type

jkilpatr:
This adds a validating WgKey type that prevents mistakes using keys
and generally makes things more programmer friendly and less prone
to bad strings sneaking in as WgKeys


jkilpatr:
So the only failures here are backwards compat tests which happen because they use the old config with "" for a wg key, this is no longer valid. Upgrades should work because a valid key is already on disk and is then parsed. 

In the case that a device is using an invalid key right now, well it can't open an exit tunnel, so this should be fine.

PR #369, opened 2018-11-26
Generate and use real Eth addresses

jkilpatr:
Needs Key serialization and deserialization in Clarity


clu/src/lib.rs
```diff
@@ -57,6 +61,26 @@ pub fn validate_mesh_ip(ip: &IpAddr) -> bool {
...
+            .unwrap(),
+        "0x0101010101010101010101010101010101010101"
+            .parse()
+            .unwrap(),
+    ];
+    if list_of_junk_addresses.contains(address) {
+        return false;
+    }
+    true
+}
```
drozdziak1: I don't think there's a way to really validate anything in this respect, all of those are valid Ethereum addresses that have a private key somewhere.

settings/src/lib.rs
```diff
@@ -239,8 +242,10 @@ pub struct PaymentSettings {
     pub close_fraction: Int256,
     /// The amount of billing cycles a node can fall behind without being subjected to the threshold
     pub buffer_period: u32,
-    /// Our own eth address
-    pub eth_address: EthAddress,
+    /// Our own eth address, derived from eth_private_key and stored for ease of access
+    pub eth_address: Option<Address>,
```
drozdziak1: We don't need the address. It's another thing for the user to worry about. Any address is deterministically derived off the private key (last 20 bytes of the Keccak hash of the public key which in itself also has a deterministic relationship with the private key)

clu/src/lib.rs
jkilpatr: this is mostly to remove the junk addresses we commonly use, there's no real validation possible otherwise, you're right about that.

settings/src/lib.rs
jkilpatr: I was on the fence about that. You can see in clu that on startup we check that address matches the private key and fix it if it doesn't. We could totally ignore storing it but in doing so we'd have to compute our own address quite a bit, I'm not really sure which way is better.

PR #378, opened 2018-12-04
Validate that transactions are actually real

jkilpatr:
This adds transaction validation to the payments system, when we
get a payment from a peer we will go and look it up on our own
full node and check that it satisfies all of the properties
required to make it a valid payment to us for the amount expected.

The biggest issue with this commit is that it adds more failure
conditions to a payment being accepted but there's still no resolution
to how to re-converge on lost payments or otherwise 'try again' and
make sure things went through.

I think a good solution might be some sort of watcher that will retry
failed txid validations a few times in the case of some spurious failure
that should reduce things to malicious behavior and uncommon enough edgecases
as to not be an issue.

### DIFF
--- a/rita/src/rita_common/payment_controller/mod.rs
+++ b/rita/src/rita_common/payment_controller/mod.rs
@@ -97,7 +97,7 @@ impl PaymentController {
 
     /// This is called by the other modules in Rita to make payments. It sends a
     /// PaymentTx to the `mesh_ip` in its `to` field.
-    pub fn make_payment(&mut self, pmt: PaymentTx) -> Result<(), Error> {
+    pub fn make_payment(&mut self, mut pmt: PaymentTx) -> Result<(), Error> {
         let payment_settings = SETTING.get_payment();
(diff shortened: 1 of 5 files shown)


rita/src/rita_common/payment_controller/mod.rs
```diff
@@ -153,7 +153,10 @@ impl PaymentController {
 
         let futures_chain = Box::new(transaction_status.then(move |outcome| {
             match outcome {
-                Ok(_tx_id) => {
+                Ok(tx_id) => {
+                    // add published txid to submission
+                    let mut pmt = pmt;
```
mpapierski: This looks weird - isnt it redundant?

jkilpatr: the binding isn't mutable, we could rebind it as mutable at the top of the function but I did it down here instead

jkilpatr: fixed

PR #383, opened 2018-12-07
Update guac_rs and num256

mpapierski:
This brings guac_rs and num256 to updated versions (respectively recent `master` and 0.2).

Needs some math review + adjust for new more explicit API of num256. Work in progress.


rita/src/rita_client/dashboard/neighbors.rs
```diff
@@ -1,11 +1,12 @@
...
+use num256::Uint256;
 
 #[derive(Serialize)]
 pub struct NodeInfo {
     pub nickname: String,
     pub route_metric_to_exit: u16,
-    pub total_payments: Int256,
-    pub debt: i128,
+    pub total_payments: Uint256,
+    pub debt: Int256,
```
jkilpatr: is this properly displayed such that it's not going to have hex values?

mpapierski: Good idea. No reason to keep current behavior. I'll address it here: althea-mesh/num256_rs#10

mpapierski: Solved in num256_rs 0.2.4

PR #390, opened 2018-12-20
Basic integration of compressed_log crate.

mpapierski:
This replaces syslog with [compressed_log](https://crates.io/crates/compressed_log) at least in this POC. Needs more improvements, but basically it works as intended.

- There is problem with TRACE level with `Arbiter::spawn` and I'm 99% sure its because during `Arbiter::spawn` to send a message there is a `trace!` log issued, but we're already holding a lock. Recursive mutex might solve this issue, though. WIP.
- URLs are hardcoded because configuration expects IP address. This could be better configured with an URL instead of hardcoding it now.

### DIFF
--- a/rita/src/rita_client/exit_manager/mod.rs
+++ b/rita/src/rita_client/exit_manager/mod.rs
@@ -36,9 +36,9 @@ use futures::Future;
 
 use tokio::net::TcpStream as TokioTcpStream;
 
+use compressed_log::builder::LoggerBuilder;
+use compressed_log::lz4::Compression;
 use log::LevelFilter;
-use syslog::Error as LogError;
-use syslog::{init_udp, Facility};
 
 use crate::KI;
 use failure::Error;
@@ -46,7 +46,7 @@ use std::net::SocketAddr;
 use std::time::Duration;
 
 /// enables remote logging if the user has configured it
-fn enable_remote_logging(server_internal_ip: IpAddr) -> Result<(), LogError> {
+fn enable_remote_logging(server_internal_ip: IpAddr) -> Result<(), Error> {
     // now that the exit tunnel is up we can start logging over it
     let log = SETTING.get_log();
     trace!("About to enable remote logging");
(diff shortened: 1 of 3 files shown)


rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -36,41 +36,42 @@ use futures::Future;
...
-use syslog::Error as LogError;
-use syslog::{init_udp, Facility};
 
 use crate::KI;
 use failure::Error;
 use std::net::SocketAddr;
 use std::time::Duration;
 
 /// enables remote logging if the user has configured it
-fn enable_remote_logging(server_internal_ip: IpAddr) -> Result<(), LogError> {
```
jkilpatr: you can go ahead and configure this to work using a url.

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -36,41 +36,42 @@ use futures::Future;
...
-    info!("Remote logging enabled with {:?}", res);
-    return res;
+
+    let logger = LoggerBuilder::new()
+        .set_level(
+            level
+                .to_level()
+                .ok_or(format_err!("Unable to convert level filter to a level"))?,
+        )
+        .set_compression_level(Compression::Fast)
```
jkilpatr: any idea what sort of compression ratio we get from fast?

rita/src/rita_client/exit_manager/mod.rs
```diff
@@ -36,41 +36,42 @@ use futures::Future;
...
+        // TODO: It would good idea to have this configurable
+        .set_sink_url(&format!("http://{}:9999/sink/", server_internal_ip))
+        .set_threshold(16384)
```
jkilpatr: is this in bytes or kb or what?

mpapierski: Bytes

PR #420, opened 2019-02-20
API Endpoints for Getting and Setting the ETH Wallet Private Key

asoltys:


jkilpatr:
This needs rustfmt you can run

cargo +stable fmt

rita/src/rita_client/dashboard/eth_private_key.rs
```diff
@@ -0,0 +1,52 @@
+use super::*;
+use clarity::PrivateKey;
+
+pub fn get_eth_private_key(_req: HttpRequest) -> Box<dyn Future<Item = HttpResponse, Error = Error>> {
```
jkilpatr: this can just be Result<httpresponse, error> you're turning your synchronous result on line 20 into a future to match this return type and invoking a bunch of code you don't need to. But we still have lots of endpoints that work this way from back when Actix was too stubborn to take non-future types so it's no a huge deal. We're just trying to move away from it to simplify type signatures and such.

rita/src/rita_client/dashboard/eth_private_key.rs
```diff
@@ -0,0 +1,52 @@
...
+            info!("{}", error_msg);
+            ret.insert("error".to_owned(), error_msg.to_string());
+        }
+    }
+
+    // try and save the config and fail if we can't
+    if let Err(e) = SETTING.write().unwrap().write(&ARGS.flag_config) {
+        return Err(e);
+    }
+    // it's now safe to restart the process, return an error if that fails somehow
```
jkilpatr: good call we do need to do this.

rita/src/rita_client/dashboard/eth_private_key.rs
```diff
@@ -0,0 +1,52 @@
...
+            let error_msg = "No mesh IP configured yet";
+            warn!("{}", error_msg);
+            ret.insert("error".to_owned(), error_msg.to_owned());
+        }
+    }
+
+    Box::new(future::ok(HttpResponse::Ok().json(ret)))
+}
+
+pub fn set_eth_private_key(eth_private_key_data: Json<HashMap<String, String>>) -> Result<HttpResponse, Error> {
```
jkilpatr: Why post a hashmap object? so that we can set dozens of keys at once?

If you want a post format like {"eth_key": "actual key data I'm too lazy to type"} then you need an object like this

#[derive(Deserialize)]
struct KeyItem {
eth_key: PrivateKey
}

The Json fromat of the struct keyitem would then be exactly as above, you would derive deserialize so that the struct could be read from it's json format. 

The alternative is that you just have it take a Json<Privatekey>

which could get kinda strange but if it works right would be {"private key data I don't want to type"} in the post request. 

The finally option is a path variable which is how you get things like 

POST /endpoint/<actual key value goes here>

and that looks like Path<PrivateKey> or somthing like that. In theory we can use it here because no spaces, but it's also not something we want to screw up by involving it in strange url limitations so I don't think we should go this way. 

KeyItem is probably the way to go.

asoltys:
I think I've addressed your comments now.

Issue #440, opened 2019-03-25
Self enforcement doesn't work


jkilpatr:
In https://github.com/althea-mesh/althea_rs/commit/e3bd4dcbb22e0ed580b2d13d86d892febc0cc84a I tried to have devices limit themselves to the free tier when they had a low balance. 

I thought limiting the wg_exit tunnel with tbf would be sufficient but it turns out that doesn't do what I expect and somehow traffic can traverse that tunnel without being subject to that rule.

jkilpatr:
I've realized that my approach here is entirely wrong thanks to @jtremback  I've realized that we can't rate limit incoming (or download) traffic using our current method and tc

jkilpatr:
after further discussion with @jtremback  it seems the best approach is to limit upload on the local side and use the low balance warning parameter to tell the exit to limit download speed exit side. This will limit usage during low balance periods voluntarily and achieve our most important goal of allowing for clients downstream of relays to be immune to a relay that's both not depositing and heavily using themselves.

PR #445, opened 2019-03-26
WIP: Fix local enforcement

jkilpatr:
Doesn't work at the moment, I'm trying to use the same strategies
as exit enforcement but it seems I'll have to acutally break out
my linux kernel routing diagram and figure out where I'm going wrong


jkilpatr:
this approach can't work, discussed one that will with @jtremback  see issue #440

PR #451, opened 2019-04-04
WifiSettings combined endpoint

jkilpatr:
Most users try and change the wifi settings while on wifi themselves
this causes a bunch of problems when they attempt to change several
settings at different endpoints but their connection is disrupted by
each individual change.

From the users perspective only some changes get applied when they
hit save.

This commit updates the /wifi_settings endpoint to interpret the
same json objects previously posted to individual endpoints as an array
of tokens all of which are iterated over and applied as one operation

### DIFF
--- a/docs/api/router-dashboard.md
+++ b/docs/api/router-dashboard.md
@@ -484,6 +484,9 @@ application/json" -d '\{"url": "https://somewhere.safe"\}'
 
 ## /wifi_settings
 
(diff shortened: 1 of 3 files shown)


asoltys:
Tested and updated dashboard in https://github.com/althea-mesh/althea-dash/commit/4ac1f3bead0c20e56c64090b5d61845d1d11898c

PR #452, opened 2019-04-04
Add dashboard authentication

jkilpatr:
This adds http auth to the Rita dashboard endpoints, before any value
is set no password is requested, once a new value is set it will be
used from there on out.


jkilpatr:
setting the local password doesn't work and the sha digest doesn't seem to match sha3sum on the console (possible assci versus utf encoding issue?)

jkilpatr:
the issue was not using fixed length sha3 encoding and such. Also upgraded to 512

jkilpatr:
@asoltys  so we probably want to use a repo on the althea github, also your commit message for your unauthorized changed is really bad and should probably be expanded. Once we've got this setup to operate from the althea github and all that we should be good.

asoltys:
Ok I forked the repo and also rebased on master a little more cleanly so this should be ready to merge now. Just waiting on Travis.

PR #473, opened 2019-04-30
Exclude wg-keys from identity hashing

jkilpatr:
This is a somewhat risky change, on one hand we have an issue where
exits (with different peer to peer and full path keys) are not being
properly attributed in billing but on the ohter hand we rely on identity
hashes quite often.


jkilpatr:
this seems to have caused intermediaries to lose track of who was owed what and not update debts values. Bizzare, can't track down the actual error path.

PR #511, opened 2019-07-03
Add /router/reboot endpoint

asoltys:


jkilpatr:
closing in favour of 

https://github.com/althea-net/althea_rs/pull/512

which has a different version of the same commit.

PR #512, opened 2019-07-08
Reset exit state after private key import

asoltys:


rita/src/rita_client/dashboard/eth_private_key.rs
```diff
@@ -39,6 +41,11 @@ pub fn set_eth_private_key(data: Json<EthPrivateKey>) -> Result<HttpResponse, Er
     SETTING.get_payment_mut().eth_private_key = Some(pk);
```
jkilpatr: this is a deadlock causign conditons, you have a write lock on the global SETTING object for get_payment_mut then you try and get another write lock to get_exits_mut so what you should do is go 

let payment_settings = SETTING.get_payment_mut().eth_private_key
.... 
// the drop here will release the write lock from get_payment_mut 
drop payment_settings;


SETTING.get_exits_mut()

PR #526, opened 2019-08-04
Ignore negative debts from the exit

jkilpatr:


jkilpatr:
this is wrong somehow, but I'm not sure how.

jkilpatr:
this is right it's the gateway client corner case that's unreliable/wrong

PR #527, opened 2019-08-05
Add a new setting and endpoints to track if keys have been backed up

asoltys:


jkilpatr:
Needs documentation in the docs folder, otherwise good.

PR #536, opened 2019-08-23
Multi lan

jkilpatr:
fixes multiple lan support, adds support for static WAN interfaces


jkilpatr:
@asoltys  note the new type of WAN interface StaticWAN, I think a modal to set it when we go to make a WAN port would work out well.

thoughts @jtremback ?

jkilpatr:
Also this is a Beta 9 feature, do not merge until after the Beta 8 window

PR #545, opened 2019-09-10
cake as leaf qdisc

jkilpatr:
This adds codel as the leaf qdisc to the exit htb setup. I don't
think this is actually doing much over pfifo-fast in it's current form
though


jkilpatr:
So this is actually an effective patch that I would like to merge, sadly on Centos 7 we run into this kernel bug. 

https://bugzilla.kernel.org/show_bug.cgi?id=109581

Which doesn't seem fatal but may be enough reason to leave it aside until Centos 8 comes out later this year.

PR #551, opened 2019-09-25
Add latency and packet loss stats to neighbors endpoint

asoltys:


rita/src/rita_client/dashboard/neighbors.rs
```diff
@@ -133,6 +150,7 @@ fn generate_neighbors_list(
                 debt: debt_info.debt.clone(),
                 link_cost: exit_route.refmetric,
                 price_to_exit: exit_route.price,
+                stats: *(stats.get(&neigh_route.iface).unwrap()),
```
jkilpatr: To clean this up you would want to wrap this block in an if let

if let Some(stats_entry) = stats.get(&neigh_route.iface) {
output.push(... // but this time stats_entry is safely handled and not unwrapped) 
}
else {
// nonviable node stuff
}

jkilpatr: another possibility is to make the if let on line 106 into a match statement then match on both of these things. see the match statement chapter in the rust book.

rita/src/rita_common/network_monitor/mod.rs
```diff
@@ -180,6 +184,77 @@ impl Default for NetworkMonitor {
     }
 }
 
+pub struct GetNetworkMonitor;
```
jkilpatr: I don't think this is used anymore, if it's not remove it.

rita/src/rita_common/network_monitor/mod.rs
```diff
@@ -180,6 +184,77 @@ impl Default for NetworkMonitor {
...
+pub type Stats = HashMap<String, IfaceStats>;
+
+impl Handler<GetStats> for NetworkMonitor {
+    type Result = Result<Stats, Error>;
+
+    fn handle(&mut self, msg: GetStats, _ctx: &mut Context<Self>) -> Self::Result {
+        let mut stats = Stats::new();
+
+        for route in msg.routes.iter() {
+            let latency_stats = self.latency_history.get_mut(&route.iface).unwrap();
```
jkilpatr: you shouldn't need to get these as mutable, you're not trying to change the history? You probably added mut here to avoid a .clone() but in this case you need one.

rita/src/rita_common/network_monitor/mod.rs
```diff
@@ -180,6 +184,63 @@ impl Default for NetworkMonitor {
...
+    std_dev: Option<f32>,
+}
+
+#[derive(Default, Serialize, Copy, Clone)]
+pub struct PacketLossStats {
+    avg: Option<f32>,
+    five_min_avg: f32,
+}
+
+pub struct GetStats {
```
jkilpatr: why exactly are we asking from the neighbours endpoint end to get a specific list of stats? If we just send what we have then the list should be the same and we don't risk crashing with .unwrap() if it's not.

rita/src/rita_common/network_monitor/mod.rs
```diff
@@ -180,6 +184,63 @@ impl Default for NetworkMonitor {
...
+
+pub type Stats = HashMap<String, IfaceStats>;
+
+impl Handler<GetStats> for NetworkMonitor {
+    type Result = Result<Stats, Error>;
+
+    fn handle(&mut self, msg: GetStats, _ctx: &mut Context<Self>) -> Self::Result {
+        let mut stats = Stats::new();
+
+        for route in msg.routes.iter() {
```
jkilpatr: Somehow you want to combine these two lists without unwrap (because unwrap crashes, what if someone views neighbors while and we have a latency stats entry for an interface but not a packet loss stats entry? In theory this should never happen because of the way actix works, but the best way to write code that works is to assume someone's going to go around breaking things like that. 

for (iface, latency_stats) self.latency_history.iter() {
if let Some(packet_stats) = self.packet_loss_history.get(&iface) {
stats.insert(...
}
else {
error!("Found entry in one that's not in the other ") // do this error message better but you get the idea
}
}

this way we have no unwraps, and a simpler request from the other side. 

https://users.rust-lang.org/t/how-to-iterate-through-two-arrays-at-once/783/3

This could be used to zip the iterators, or you could iter over one and if match the other, or a match statement where each one uses .iter().next() (to get the next entry). But this is syntactically complicated although pretty if you can do it. Match statements are really awesome. They take complicated if/else trees and turn them into giant pattern matching self checking systems. Kinda like switch statements but more awesome.

rita/src/rita_client/dashboard/neighbors.rs
```diff
@@ -123,17 +140,20 @@ fn generate_neighbors_list(
...
-                ip: identity.mesh_ip.to_string(),
-                id: *identity,
-                route_metric_to_exit: exit_route.metric,
-                route_metric: neigh_route.metric,
-                total_payments: debt_info.total_payment_received.clone(),
-                debt: debt_info.debt.clone(),
-                link_cost: exit_route.refmetric,
-                price_to_exit: exit_route.price,
-            })
+            if let Some(stats_entry) = stats.get(&neigh_route.iface) {
```
jkilpatr: Right now you won't print an entry at all for a peer if you don't get a stats entry for it. You can either add an else statement or do a nice trick on line 123

let tup = (current_exit, stats.get(&neigh_route.iface))
if let (Some(current_exit), Some(stats_entry) = tup) {
...
}
else {
 ....
}

What this does is let you reuse the else state by pattern matching on a tuple, in the cases where one is some and the other is None it will go to else. Anyway it's slightly cleaner with less code duplication really a nit.

PR #558, opened 2019-10-19
Add /routes endpoint

asoltys:


babel_monitor/Cargo.toml
```diff
@@ -12,3 +12,4 @@ ipnetwork = "0.14"
 log = "0.4"
 futures = "0.1"
 tokio = "0.1"
+serde = "1.0.101"
```
jkilpatr: this should be just 1.0 so that the minor revision updates on it's own. As you can tell it gets iterated quite often and it's annoying to keep track of.

PR #566, opened 2019-11-20
Read Wyre account_id from oracle updates

asoltys:


settings/src/localization.rs
```diff
@@ -8,4 +8,7 @@ pub struct LocalizationSettings {
     // cryptocurrency through Wyre as part of the funding flow.
     #[serde(default = "default_wyre_enabled")]
     pub wyre_enabled: bool,
+    // Wyre account_id used to associate transactions with a specific Wyre account
+    #[serde(default)]
```
jkilpatr: this is the only thing we actually need, because we put the struct itself on the wire in the endpoint we don't need any mods there and becuase we have the arbitrary settings modifying block we don't need any special code. 

We could remove the blockchain settings in the oracle since we could merge them through settings changes. Except gateway client prices where we actually need to know what type of node we are. I've left them broken out for easier modification (And becuase we change them often) for this no need to do anything.

PR #568, opened 2019-12-02
Add UDP heartbeat monitor to fast loop

asoltys:


rita/src/rita_client/rita_loop/mod.rs
```diff
@@ -76,6 +78,24 @@ impl Handler<Tick> for RitaLoop {
 
         Arbiter::spawn(check_for_gateway_client_billing_corner_case());
 
+        match UdpSocket::bind("0.0.0.0:33333") {
```
jkilpatr: move this out into a function please.

jkilpatr: it should also check if the user has metrics enabled in the settings.

PR #569, opened 2019-12-09
Back to the future

jkilpatr:
This adds in compat with std futures this will allow us to modify crates one by one under Rita to use std-futures and then modify rita itself into executing only async code. It also lets us use mixed async and old style futures in Rita for the time being.

PR #570, opened 2019-12-09
UDP Heartbeat

asoltys:
This branch is based off the back-to-the-future branch #569


asoltys:
Rebased off master instead of the futures branch now

PR #582, opened 2021-11-30
Pranay devenv

ptulugu:


rita_client/src/heartbeat/mod.rs
```diff
@@ -285,20 +291,31 @@ fn send_udp_heartbeat_packet(
...
-    let remote_ip = dns_socket.ip();
     let remote_port = dns_socket.port();
     let remote = dns_socket;
 
+    // When in op tools development mode, since we use the same container for rita and optools, we cant
+    // use port 33333 to both listen and send packets, so we change the sender port in this mode
+    #[cfg(not(feature = "dev_env"))]
     let local_socketaddr = SocketAddr::from(([0, 0, 0, 0], remote_port));
+    #[cfg(feature = "dev_env")]
+    let local_socketaddr = SocketAddr::from(([0, 0, 0, 0], remote_port + 2));
```
jkilpatr, line 302: this feature flag isn't needed, we can just use remote_port + 2 all the time. Specifically a constant local port that's just remote port iterated.

I appreciate trying to keep non-testing environment changes to a minimum, but in this case I think flattening it is worth more since the send port doesn't matter.

jkilpatr (changes requested):
one minor change, looks good otherwise.

PR #593, opened 2022-01-10
Added handler for SIGTERM

ptulugu:


rita_common/src/debt_keeper/mod.rs
```diff
@@ -606,6 +606,21 @@ impl DebtKeeper {
...
 
+/// On an interupt (SIGTERM), saving debtkeeper before exiting
+pub fn save_debt_on_shutdown() {
+    let mut dk = DEBT_DATA.write().unwrap();
+    let previous_save = dk.last_save;
+    dk.last_save = Some(Instant::now());
+
+    if let Err(e) = dk.save() {
+        error!("Failed to save debts {:?}", e);
+        dk.last_save = previous_save;
```
jkilpatr, line 617: (A) this can be an if/else and avoid the return statement, B what's up with the last save logic? I don't think it does anything give that the program is exiting

jkilpatr (changes requested):
This looks great, very compact and ctrlc is a nice package I didn't know about before. No new trasitive dependencies either which is always nice.

Commit message should detail more of the 'why' for this change, which is to prevent losing data from any of these data stores.

jkilpatr (approved):

PR #596, opened 2022-01-21
migrating network_monitor to async/await

ch-iara:
Converted network_monitor to async/await: actix impls removed and replaced with lazy_static to store data for cross thread getters

### DIFF
--- a/rita_client/src/dashboard/neighbors.rs
+++ b/rita_client/src/dashboard/neighbors.rs
@@ -10,7 +10,7 @@ use babel_monitor_legacy::start_connection_legacy;
 use futures01::Future;
 use num256::{Int256, Uint256};
 use rita_common::debt_keeper::{dump, NodeDebtData};
-use rita_common::network_monitor::{GetStats, IfaceStats, NetworkMonitor, Stats};
+use rita_common::network_monitor::{get_stats, IfaceStats, Stats};
 use rita_common::tunnel_manager::{GetNeighbors, Neighbor, TunnelManager};
 use std::collections::HashMap;
 
@@ -82,20 +82,13 @@ pub fn get_neighbor_info(
                                     .from_err()
                                     .and_then(|(_stream, routes)| {
                                         let route_table_sample = routes;
-
-                                        NetworkMonitor::from_registry()
-                                            .send(GetStats {})
-                                            .from_err()
-                                            .and_then(|stats| {
-                                                let stats = stats.unwrap();
-                                                let output = generate_neighbors_list(
-                                                    stats,
-                                                    route_table_sample,
-                                                    combined_list,
-                                                );
-
-                                                Ok(Json(output))
-                                            })
+                                        let stats = get_stats();
+                                        let output = generate_neighbors_list(
+                                            stats,
+                                            route_table_sample,
+                                            combined_list,
+                                        );
+                                        Ok(Json(output))
                                     })
                                     .responder()
                             })
--- a/rita_client/src/heartbeat/mod.rs
+++ b/rita_client/src/heartbeat/mod.rs
@@ -22,8 +22,8 @@ use babel_monitor::BabelMonitorError;
 
 use dummy::dummy_selected_exit_details;
 
+use rita_common::network_monitor::get_network_info;
 use rita_common::network_monitor::GetNetworkInfo;
-use rita_common::network_monitor::NetworkMonitor;
 use rita_common::tunnel_manager::Neighbor as RitaNeighbor;
 
 use actix::actors::resolver;
@@ -95,9 +95,6 @@ pub fn send_udp_heartbeat() {
     let dns_request = Resolver::from_registry()
         .send(resolver::Resolve::host(heartbeat_url.to_string()))
         .timeout(CLIENT_LOOP_TIMEOUT);
-    let network_info = NetworkMonitor::from_registry()
-        .send(GetNetworkInfo {})
-        .timeout(CLIENT_LOOP_TIMEOUT);
 
     // Check for the basics first, before doing any of the hard futures work
     #[allow(unused_assignments, unused_mut)]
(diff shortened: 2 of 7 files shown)


rita_client/src/dashboard/neighbors.rs
```diff
@@ -76,29 +76,18 @@ pub fn get_neighbor_info(
...
-                                    })
-                                    .responder()
-                            })
+                        .from_err()
+                        .and_then(move |stream| {
+                            parse_routes_legacy(stream)
+                                .from_err()
+                                .and_then(|(_stream, routes)| {
+                                    let route_table_sample = routes;
+                                    let stats = get_stats(GetStats {});
```
jkilpatr, line 85: we can remove the GetStats{} argument here, it's no longer doing anything now that we don't need to attach an actix message trait implementation to it.

rita_client/src/heartbeat/mod.rs
```diff
@@ -130,12 +128,22 @@ pub fn send_udp_heartbeat() {
 
     trace!("we have heartbeat basic info");
 
-    let res = dns_request.join(network_info).then(move |res| {
+    //let networkInfoVal = network_info.expect("Failed to get network info");
```
jkilpatr, line 131: remove this commented out line.

rita_client/src/heartbeat/mod.rs
```diff
@@ -95,9 +95,7 @@ pub fn send_udp_heartbeat() {
     let dns_request = Resolver::from_registry()
         .send(resolver::Resolve::host(heartbeat_url.to_string()))
         .timeout(CLIENT_LOOP_TIMEOUT);
-    let network_info = NetworkMonitor::from_registry()
-        .send(GetNetworkInfo {})
-        .timeout(CLIENT_LOOP_TIMEOUT);
+    let _network_info = get_network_info(GetNetworkInfo);
```
jkilpatr, line 98: since this is not used (indicated by the _) it should be removed.

rita_client/src/heartbeat/mod.rs
```diff
@@ -187,12 +195,12 @@ pub fn send_udp_heartbeat() {
...
-            Ok((Err(e), _)) => {
+            Ok(Err(e)) => {
                 warn!("DNS resolution failed with {:?}", e);
             }
-            Ok((_, Err(e))) => {
-                warn!("Could not get network info with {:?}", e);
-            }
+            // Ok((_, Err(e))) => {
+            //     warn!("Could not get network info with {:?}", e);
+            // }
```
jkilpatr, line 203: commented out code to remove, but the warn message should go on line 136 so that we can log the error.

rita_common/src/network_monitor/mod.rs
```diff
@@ -102,84 +83,62 @@ pub struct IfaceStats {
...
-                        },
-                        packet_loss: PacketLossStats {
-                            avg: packet_loss_stats.get_avg(),
-                            five_min_avg: packet_loss_stats.get_five_min_average(),
-                        },
+pub fn get_stats(_msg: GetStats) -> Stats {
+    let mut stats = Stats::new();
+
+    for (iface, latency_stats) in (*(NETWORK_MONITOR.write().unwrap())).latency_history.iter() {
+        if let Some(packet_loss_stats) = (*(NETWORK_MONITOR.write().unwrap())).packet_loss_history.get(iface) {
```
jkilpatr, line 92: here's probably your deadlock, get the data from network monitor once at the top, then use that value twice here. Each time you call 'write' you're grabbing a write lock, a write lock is exclusive so right now you are grabbing one once at the start of the for loop and again during each iteration, resulting in a deadlock. 

These can be a lot more subdle, hopefully this is it and we have an easy time of it.

rita_common/src/network_monitor/mod.rs
```diff
@@ -102,84 +83,62 @@ pub struct IfaceStats {
...
-
-    fn handle(&mut self, _msg: GetNetworkInfo, _ctx: &mut Context<Self>) -> Self::Result {
-        match self.last_babel_dump.clone() {
-            Some(dump) => Ok(dump),
-            None => Err(RitaCommonError::MiscStringError(
-                "No babel info ready!".to_string(),
-            )),
-        }
+pub fn get_network_info(_msg: GetNetworkInfo) -> Result<NetworkInfo, RitaCommonError> {
+    match (*(NETWORK_MONITOR.write().unwrap())).last_babel_dump.clone() {
```
jkilpatr, line 116: this is a read only operation, it does not need a write lock, only a read lock. It's much better to use a read lock than a write lock when possible becuase there can be an unlimited number of parallel readers but only one writer.

rita_common/src/network_monitor/mod.rs
```diff
@@ -102,84 +83,62 @@ pub struct IfaceStats {
...
-    }
+/// updates babel neighbors, babel routes, and rita neighbors for a NetworkInfo
+pub fn update_network_info(msg: NetworkInfo) {
+    let babel_neighbors = &msg.babel_neighbors;
+    let babel_routes = &msg.babel_routes;
+    let rita_neighbors = &msg.rita_neighbors;
+    observe_network(
+        babel_neighbors,
+        rita_neighbors,
+        &mut (*(NETWORK_MONITOR.write().unwrap())).latency_history,
```
jkilpatr, line 137: good job figuring out how to deal with the references here, but as noted previously it's better to get one mutable object and pass it, that way we only need to get the lock once.

let mut network_monitor = NETWORK_MONITOR.write().unwrap();

this actaully frees you from doing a lot of reference operations becuase .latency_history will still 'just work' this is becuase the '.' statement in Rust can do a ton of lifting under the hood to remove referencing/dereferencing etc so long as the compiler can prove to itself that it's safe.

rita_common/src/network_monitor/mod.rs
```diff
@@ -287,10 +252,9 @@ fn observe_network(
     // shape the misbehaving tunnels, we do this all at once for the sake
     // of efficiency as lots of do_sends have a high chance of getting lost
     // also there's nontrivial overhead
-    TunnelManager::from_registry().do_send(ShapeMany { to_shape });
-
+    set_to_shape(to_shape);
     // update NeighborStatus info for operator tools
-    TunnelManager::from_registry().do_send(UpdateNeighborStatus);
+    //TunnelManager::from_registry().do_send(UpdateNeighborStatus);
```
jkilpatr, line 257: commented out code to remove.

rita_common/src/tunnel_manager/neighbor_status.rs
```diff
@@ -18,46 +17,38 @@ pub fn get_neighbor_status() -> HashMap<Identity, NeighborStatus> {
...
-            let mut enforced = false;
-            for tunnel in tunnel_list.iter() {
-                match (tunnel.speed_limit, lowest_shaper_speed) {
-                    (Some(new), Some(current)) => {
-                        if new < current {
-                            lowest_shaper_speed = Some(new);
-                        }
+/// Handles updates to neighbor status with lazy static lock
+pub fn update_neighbor_status(tunnel_manager: &mut TunnelManager) {
+    let mut external_list = NEIGHBOR_STATUS.write().unwrap();
```
jkilpatr, line 22: see this as an example for the scheme to avoid multiple .write() calls in the same function.

jkilpatr (changes requested):
This looks good! It's a big change and the substance of it is all correct.

There are a few issues involving locks that I've flagged. I think these will be our deadlock, which is great because deadlock bugs can be a lot harder to deal with.

jkilpatr (approved):

jkilpatr (approved):

PR #597, opened 2022-01-25
Fix add_iptables_rule usage

thosmos:
add_iptables_rule expects that a rule with -I will also have an insertion location included.  Leaving it out will only impact if the rule is run multiple times, in which case the check to see if it already exists will fail resulting in an additional rule being added.

### DIFF
--- a/althea_kernel_interface/src/iptables.rs
+++ b/althea_kernel_interface/src/iptables.rs
@@ -2,6 +2,10 @@ use crate::KernelInterface;
 use crate::KernelInterfaceError;
 
 impl dyn KernelInterface {
+    /// -I and -A will be checked to see if they already exist before adding.
+    /// Expects that a rule with -I will also have an insertion location included.
+    /// Leaving it out will only impact if the rule is run multiple times,
+    /// in which case the check to see if it already exists will fail resulting in an additional rule being added.
     pub fn add_iptables_rule(
         &self,
         command: &str,
--- a/althea_kernel_interface/src/iptables.rs
+++ b/althea_kernel_interface/src/iptables.rs
@@ -3,9 +3,6 @@ use crate::KernelInterfaceError;
 
 impl dyn KernelInterface {
     /// -I and -A will be checked to see if they already exist before adding.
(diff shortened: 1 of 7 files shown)


althea_kernel_interface/src/iptables.rs
```diff
@@ -2,6 +2,10 @@ use crate::KernelInterface;
 use crate::KernelInterfaceError;
 
 impl dyn KernelInterface {
+    /// -I and -A will be checked to see if they already exist before adding.
+    /// Expects that a rule with -I will also have an insertion location included.
+    /// Leaving it out will only impact if the rule is run multiple times,
+    /// in which case the check to see if it already exists will fail resulting in an additional rule being added.
     pub fn add_iptables_rule(
         &self,
         command: &str,
```
jkilpatr, line 11: this is really generic, I've always thought we should modify it to take an enum and build a command that way. I'm trying to consider if we want to do that here or what.

jkilpatr:
My clippy lints didn't catch this locally either, had to run rustup-update, I've ammeneded the comit with the fixes so that it's ready to merge.

althea_kernel_interface/src/iptables.rs
thosmos: It's a good question.  I missed it previously because there was no comment about it.  But it would be even better if the compiler would require an insertion position.  Also, there should probably be a separate fn for deleting.  I think the rest of the rule should stay an open array for now, given the massive set of possibilities.

thosmos: I'll take a crack at adding an enum

thosmos: Actually, what about this simpler safety fix:

```rust
            if *rul == "-I" {
                // if there's an insertion index, then skip it for the check statement
                if rule[i + 2].parse::<u32>().is_ok() {
                    i_pos_skip = Some(i + 2);
                }
                new_command.push("-C");
            }
```

thosmos:
Made a fix to first check if the caller included an insertion index, and only does a skip for the check then.  This avoids breaking any external code that depends on this function.

PR #602, opened 2022-02-03
Add the LTE port type and network interface

thosmos:
This adds a new  LTE port and lte network interface that is basically a DHCP WAN port, but does not have the gateway characteristics, and therefore no static routes to exits nor gateway pricing.  This expects the /etc/config/firewall to include the 'lte' interface in the backhaul firewall zone, but this does not make sure it's added.

### DIFF
--- a/rita_client/src/dashboard/interfaces.rs
+++ b/rita_client/src/dashboard/interfaces.rs
@@ -165,14 +165,13 @@ fn set_interface_mode(iface_name: &str, mode: InterfaceMode) -> Result<(), RitaC
         return Err(RitaClientError::InterfaceModeError(
             "Attempted to configure non-existant or unavailable interface!".to_string(),
         ));
-    } else if target_mode == InterfaceMode::Wan || target_mode == InterfaceMode::LTE {
+    } else if target_mode == InterfaceMode::Wan {
         // we can only have one WAN interface, check for others
         // StaticWAN entries are not identified seperately but if they ever are
         // you'll have to handle them here
         for entry in interfaces {
-            let iface = entry.0;
             let mode = entry.1;
-            if iface != iface_name && (mode == InterfaceMode::Wan || mode == InterfaceMode::LTE) {
+            if mode == InterfaceMode::Wan {
                 return Err(RitaClientError::InterfaceModeError(
                     "There can only be one WAN interface!".to_string(),
                 ));
--- a/rita_client/src/dashboard/interfaces.rs
+++ b/rita_client/src/dashboard/interfaces.rs
@@ -39,6 +39,8 @@ pub enum InterfaceMode {
         ipaddr: Ipv4Addr,
         gateway: Ipv4Addr,
     },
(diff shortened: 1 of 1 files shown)


rita_client/src/dashboard/interfaces.rs
```diff
@@ -161,13 +165,14 @@ fn set_interface_mode(iface_name: &str, mode: InterfaceMode) -> Result<(), RitaC
         return Err(RitaClientError::InterfaceModeError(
             "Attempted to configure non-existant or unavailable interface!".to_string(),
         ));
-    } else if target_mode == InterfaceMode::Wan {
+    } else if target_mode == InterfaceMode::Wan || target_mode == InterfaceMode::LTE {
         // we can only have one WAN interface, check for others
-        // StaticWAN entires are not identified seperately but if they ever are
```
thosmos, line 166: I took a quick pass at StaticWan but it behaves differently due to being a struct enum type.  It would be nice to check, but it's not essential.

rita_client/src/dashboard/interfaces.rs
```diff
@@ -161,13 +165,14 @@ fn set_interface_mode(iface_name: &str, mode: InterfaceMode) -> Result<(), RitaC
...
+        // StaticWAN entries are not identified seperately but if they ever are
         // you'll have to handle them here
         for entry in interfaces {
+            let iface = entry.0;
             let mode = entry.1;
-            if mode == InterfaceMode::Wan {
+            if iface != iface_name && (mode == InterfaceMode::Wan || mode == InterfaceMode::LTE) {
```
thosmos, line 175: This avoids an error when changing from WAN to LTE

jkilpatr: adding that as a comment could be valuable.

rita_client/src/dashboard/interfaces.rs
```diff
@@ -39,6 +39,8 @@ pub enum InterfaceMode {
         ipaddr: Ipv4Addr,
         gateway: Ipv4Addr,
     },
+    /// Similar to WAN, but not a gateway, so no static DNS routes and only IP peers (10.45.0.1)
+    LTE,
```
jkilpatr, line 43: I notice nothing is being done with manual peers here. That's probably ok.

rita_client/src/dashboard/interfaces.rs
jkilpatr: so this makes wan and LTE exclusive? is that actually required? Since we don't have external nic or any gateway stuff I don't think so.

jkilpatr (approved):
Just added some thoughts otherwise good, bump for beta 19 before doing a release.

rita_client/src/dashboard/interfaces.rs
thosmos: I began with the assumption they would be exclusive, but as I got into it, I think you're right that they could be compatible.

thosmos: Removed the exclusion check

PR #603, opened 2022-02-03
Modernizing usage tracker, replacing serde_json with bincode

ch-iara:
- load_from_disk() now looks for the updated .bincode usage tracker file to read from, and if only the old .json is found reads from that then resaves with bincode and deletes the .json file
- save() now uses bincode serialization to a .bincode file
- added unit tests

### DIFF
--- a/rita_common/src/usage_tracker/mod.rs
+++ b/rita_common/src/usage_tracker/mod.rs
@@ -7,21 +7,26 @@
 use actix::Message;
 use althea_types::Identity;
 use althea_types::PaymentTx;
+use bincode::Error as BincodeError;
 use flate2::read::ZlibDecoder;
 use flate2::write::ZlibEncoder;
 use flate2::Compression;
 use num256::Uint256;
 use serde::{Deserialize, Serialize};
-use serde_json::Error as SerdeError;
+use serde_json::Error as JsonError;
+use settings::set_rita_common;
 use std::collections::VecDeque;
 use std::fs::File;
 use std::io::Error as IOError;
 use std::io::Read;
 use std::io::Write;
+use std::path::Path;
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
+use std::usize;
 
 use crate::RitaCommonError;
 
@@ -52,7 +57,7 @@ pub enum UsageType {
 
 /// A struct for tracking each hour of usage, indexed by time in hours since
 /// the unix epoch
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
 pub struct UsageHour {
     index: u64,
     up: u64,
@@ -92,7 +97,7 @@ fn to_formatted_payment_tx(input: PaymentTx) -> FormattedPaymentTx {
 }
 
 /// A struct for tracking each hours of payments indexed in hours since unix epoch
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct PaymentHour {
     index: u64,
     payments: Vec<FormattedPaymentTx>,
@@ -100,7 +105,8 @@ pub struct PaymentHour {
 
 /// The main actor that holds the usage state for the duration of operations
 /// at some point loading and saving will be defined in service started
-#[derive(Clone, Debug, Serialize, Deserialize)]
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct UsageTracker {
     last_save_hour: u64,
     // at least one of these will be left unused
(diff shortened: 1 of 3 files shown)


rita_common/src/usage_tracker/mod.rs
```diff
@@ -112,18 +116,41 @@ pub struct UsageTracker {
...
+
+        let mut newsize = MAX_TX_ENTRIES;
+        loop {
+            match file.write_all(&compressed_bytes) {
+                Ok(save) => {
+                    return Ok(save);
+                }
+                Err(e) => {
+                    // 500 tx min
+                    if newsize >= 1000 {
```
jkilpatr, line 136: we should add a comment here talking about how payments are larger than usage data.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +160,120 @@ impl UsageTracker {
...
+            // if we get here, the bincode file does not exist yet
+            let file = File::open(settings::get_rita_common().network.usage_tracker_file);
+
+            match file {
+                Ok(mut file) => {
+                    let mut byte_contents = Vec::new();
+                    // try compressed
+                    match file.read_to_end(&mut byte_contents) {
+                        Ok(_bytes_read) => {
+                            let decoder = ZlibDecoder::new(&byte_contents[..]);
```
jkilpatr, line 216: can we de-duplicate this with the try compressed one on line 174, it might have to be broken out into it's own function.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +160,120 @@ impl UsageTracker {
...
+                    blank_usage_tracker
+                }
             }
         }
     }
 }
 
+/// Compresses serialized data
+fn compress_serialized(serialized: Vec<u8>) -> Result<Vec<u8>, IOError> {
+    let buffer: Vec<u8> = Vec::new();
```
jkilpatr, line 270: 👍🏻

rita_common/src/usage_tracker/mod.rs
```diff
@@ -313,6 +422,102 @@ pub fn save_usage_on_shutdown() {
 
     let history = &mut USAGE_TRACKER.write().unwrap();
     history.last_save_hour = current_hour;
+    while history.payments.len() > MAX_TX_ENTRIES {
+        let _discarded_entry = history.payments.pop_back();
```
jkilpatr, line 426: why not use the trim_payments function here?

rita_common/test.toml
```diff
@@ -0,0 +1,36 @@
+[payment]
+pay_threshold = "10000000000000"
```
jkilpatr, line 2: this file was moved here from settings, change the path on line 488 to ../../settings/test.toml (or maybe only one ../)  and delete this one to avoid duplication.

jkilpatr (changes requested):
This is looking great! I especially like the unit tests. 

Just needs a little attention on de-duplication

rita_common/test.toml
ch-iara: fixed

rita_common/src/usage_tracker/mod.rs
ch-iara: fixed

rita_common/src/usage_tracker/mod.rs
ch-iara: sort of fixed... not broken out into its own function, but nested match statements have been broken up. Trying to wrap serdeError and bincode Error into ritacommonerror was being a pain

rita_common/src/usage_tracker/mod.rs
ch-iara: fixed

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +160,107 @@ impl UsageTracker {
...
+            let mut fileopen = match file {
+                Ok(file) => file,
+                Err(e) => {
+                    error!("Failed to open usage tracker file! {:?}", e);
+                    return blank_usage_tracker;
+                }
+            };
+            let mut byte_contents = Vec::new();
+            // try compressed
+            match fileopen.read_to_end(&mut byte_contents) {
```
jkilpatr, line 179: The first two steps here, load file, zlib unzip can be combined into a function that takes file paths. Or in this case since both the if and else condition read the same variable for file path the entire operation can be done once. 

What you might find useful is the idea of doing this with a match statement rather than an if/else tree. It's easier to understand IMO.

Matching on a tuple like this allows you to fold a lot of logic into a singe line.

let file_exists = exits check
let file_is_bincode = bincode check
let grabbed_bytes_and_zlip = function that reads the file and unzlib's the bytes
match (file_exists, file_is_bincode, grabbed_bytes_and_zlip_decompressed) 
{
   (true, true, Ok(bytes)) => bincode deserialization,
  (true, false, Ok(bytes)) => serde deserialization
  ...
}

The great thing about a multi-variable match is that it will tell you all the cases you are not handling and make you consider them, something that's very difficult in a deeply nested if else tree.

jkilpatr (changes requested):
Noticed a few logical problems on a second pass this morning. Left comments about how to resolve them.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +162,94 @@ impl UsageTracker {
...
+                    Err(e) => {
+                        error!(
+                            "Failed to deserialize bytes in compressed bw history {:?}",
+                            e
+                        );
+                        blank_usage_tracker
+                    }
+                };
+                let mut rita_common = get_rita_common();
+                rita_common.network.usage_tracker_file =
```
jkilpatr, line 224: comments here about what's being done, we're changing the file name

also we should change only the filename on the end of the string, what if the user had provided /my/strange/filepath/tracker.json, we just want to take everything before the dot (their filename) and append .bincode as the new suffix

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +162,94 @@ impl UsageTracker {
...
+        }
+        let file = File::open(settings::get_rita_common().network.usage_tracker_file);
+        let fileopen = match file {
+            Ok(file) => file,
+            Err(e) => {
+                error!("Failed to open usage tracker file! {:?}", e);
+                return blank_usage_tracker;
+            }
+        };
+        match (file_exists, file_is_bincode, decompressed(fileopen)) {
```
jkilpatr, line 195: nice! I like the match, please comment each of these cases with a quick plain English summary 

// New file exists, no need for conversion
// New file does not exist, old file does, we must load using the old format and update settings 

etc etc

rita_common/src/usage_tracker/mod.rs
```diff
@@ -133,44 +162,94 @@ impl UsageTracker {
...
+                        blank_usage_tracker
+                    }
+                };
+                let mut rita_common = get_rita_common();
+                rita_common.network.usage_tracker_file =
+                    "/etc/rita-usage-tracker.bincode".to_string();
+                set_rita_common(rita_common);
+                match res.save() {
+                    Ok(()) => {
+                        let _r = std::fs::remove_file("/etc/rita-usage-tracker.json");
```
jkilpatr, line 229: This should remove the old file path variable, not the exact file path.

jkilpatr (changes requested):
This is good! 

The match statement is much cleaner and handles all conditions. I've got one final nit as far as handling of the file before we wrap up and a few requests for comments. 

The commit message should also include what you changed and a short plain-english description of how it works.

rita_common/src/usage_tracker/mod.rs
ch-iara: fixed

rita_common/src/usage_tracker/mod.rs
ch-iara: fixed, added a check for in case the old name is the same as the new

jkilpatr (approved):

PR #605, opened 2022-02-04
Pranay Migrate client and common network enpoints

ptulugu:


rita_client/src/rita_loop/mod.rs
```diff
@@ -161,17 +83,36 @@ pub fn start_rita_loop() {
...
                     check_for_gateway_client_billing_corner_case();
                     // update the client exit manager, which handles exit registrations
                     // and manages the exit state machine in general. This includes
                     // updates to the local ip and description from the exit side
                     exit_manager_tick().await;
                     // sends an operator payment if enough time has elapsed
                     tick_operator_payments().await;
+
+                    if metrics_permitted() {
+                        send_udp_heartbeat();
```
jkilpatr, line 103: I don't see any migration to AWC for heartbeats here, so I don't believe this will work. Heartbeats will simply not be sent.

rita_common/src/network_endpoints/mod.rs
```diff
@@ -19,13 +21,13 @@ pub struct JsonStatusResponse {
 impl JsonStatusResponse {
```
jkilpatr, line 21: are we sure we still need this? I think this sort of conversion is done automatically by actix web now.

rita_common/src/network_endpoints/mod.rs
```diff
@@ -66,21 +65,24 @@ pub fn make_payments(pmt: (Json<PaymentTx>, HttpRequest)) -> HttpResponse {
     HttpResponse::Ok().json("Payment Received!")
 }
 
-pub fn hello_response(
-    req: (Json<LocalIdentity>, HttpRequest),
-) -> Result<Json<LocalIdentity>, RitaCommonError> {
+pub fn hello_response(req: (Json<LocalIdentity>, HttpRequest)) -> HttpResponse {
```
jkilpatr, line 68: keeping the result return type here is important. What about the error handling presented a problem with keeping it that way?

jkilpatr (changes requested):
Looks good overall. My biggest statement is that the commit message should indicate that LightClientManager was migrated and generally include more details about what was done.

In general keeping the return type of the endpoints as a Result<Json<whatever>> is by far the best. You can use .responder() to perform a converstion to http response that will be more accurate than trying to do it yourself. 

The big issue remaining is that the send_udp_heartbeat will not work the way it's been modified here. An old actix context is required for it to send it's dns requests. This change may have to wait on heartbeats being migrated before it can be merged.

rita_common/src/network_endpoints/mod.rs
ptulugu: fixed

PR #607, opened 2022-02-09
Migrated Exit endpoints to async/await

ptulugu:
Also used return type HttpResponse here. All return types need to be of type impl Responder


jkilpatr (approved):
Nice job here! Looks like moving the database request from a future to sync really let the rest of the logic fold out nicely.

PR #608, opened 2022-02-10
Pranay migrate dashboard endpoints to async await

ptulugu:
Middleware needs to be updated

### DIFF
--- a/rita_client/src/dashboard/mod.rs
+++ b/rita_client/src/dashboard/mod.rs
@@ -25,6 +25,8 @@ pub mod system_chain;
 pub mod usage;
 pub mod wifi;
 
+use std::thread;
+
 use crate::dashboard::auth::*;
 use crate::dashboard::backup_created::*;
 use crate::dashboard::bandwidth_limit::*;
@@ -45,8 +47,8 @@ use crate::dashboard::router::*;
 use crate::dashboard::system_chain::*;
 use crate::dashboard::usage::*;
 use crate::dashboard::wifi::*;
-use actix_web::http::Method;
-use actix_web::{server, App};
+use actix_async::System;
+use actix_web_async::{web, App, HttpServer};
 use rita_common::dashboard::babel::*;
 use rita_common::dashboard::debts::*;
 use rita_common::dashboard::development::*;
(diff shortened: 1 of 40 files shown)


rita_client/src/dashboard/mod.rs
```diff
@@ -62,160 +64,156 @@ use rita_common::network_endpoints::*;
 
 pub fn start_client_dashboard(rita_dashboard_password: u16) {
```
jkilpatr, line 65: this variable should be called 'rita_dashboard_port' not password. This isn't your doing just noticed during review.

rita_client/src/exit_manager/mod.rs
```diff
@@ -255,46 +251,51 @@ fn decrypt_exit_state(
...
+async fn send_exit_setup_request(
     exit_pubkey: WgKey,
     to: SocketAddr,
     ident: ExitClientIdentity,
-) -> impl Future<Item = ExitState, Error = RitaClientError> {
+) -> Result<ExitState, RitaClientError> {
     let endpoint = format!("http://[{}]:{}/secure_setup", to.ip(), to.port());
+    info!("Into this func 1");
     let ident = encrypt_exit_client_id(&exit_pubkey.into(), ident);
+    info!("Into this func 2");
```
jkilpatr, line 262: can we remove this log, or make it more verbose if it will be good to have in the future.

rita_common/src/middleware.rs
```diff
@@ -1,26 +1,58 @@
...
+{
+    type Response = ServiceResponse<B>;
+    type Error = Error;
+    type InitError = ();
+    type Transform = HeadersMiddleware<S>;
+    type Future = Ready<Result<Self::Transform, Self::InitError>>;
 
-impl<S> Middleware<S> for Headers {
-    fn start(&self, _req: &HttpRequest<S>) -> Result<Started> {
-        Ok(Started::Done)
```
jkilpatr, line 15: for this custom headers handler work could you add a comment describing where you got the example?

rita_common/src/middleware.rs
```diff
@@ -33,55 +65,105 @@ impl<S> Middleware<S> for Headers {
...
-        if password.is_none() || req.request().path() == "/exits" {
-            return Ok(Started::Done);
-        }
+        let req_path = req.path().to_string();
+
+        let auth = Authorization::<Basic>::parse(&req);
+
+        let fut = self.service.call(req);
+
+        async move {
```
jkilpatr, line 142: please preserve the comment explaining why the exits path is exempt from authentication.

jkilpatr (changes requested):
This is great! There are a lot of changes but as far as I can tell everything has been migrated correctly. I've made a few comments with nits

A change that is not in this PR right now is to remove the old http_auth import, which seems to have been left in. 

Note that we no longer have our custom http-auth patch and we will have to see if this causes issues when tested on an actual router

rita_client/src/exit_manager/mod.rs
ptulugu: Fixed, left these in by mistake

rita_client/src/dashboard/mod.rs
ptulugu: fixed

rita_common/src/middleware.rs
ptulugu: fixed

rita_common/src/middleware.rs
ptulugu: fixed, added a comment to the top of document

jkilpatr (approved):

PR #609, opened 2022-02-15
migrating heartbeat sender to async/await

ch-iara:
https://trello.com/c/EerVFBwy

removed actix resolver in favor of to_socket_addrs and put the heartbeat sending into a loop

### DIFF
--- a/rita_bin/Cargo.toml
+++ b/rita_bin/Cargo.toml
@@ -52,6 +52,7 @@ jemallocator = {version = "0.3", optional = true}
 openssl = {version = "0.10", features = ["vendored"], optional = true}
 
 [features]
+default = ["bundle_openssl"]
 jemalloc = ["jemallocator"]
 bundle_openssl = ["openssl"]
 # Features for big iron devices with more ram
--- a/rita_bin/Cargo.toml
+++ b/rita_bin/Cargo.toml
@@ -52,7 +52,6 @@ jemallocator = {version = "0.3", optional = true}
 openssl = {version = "0.10", features = ["vendored"], optional = true}
 
 [features]
-default = ["bundle_openssl"]
 jemalloc = ["jemallocator"]
 bundle_openssl = ["openssl"]
 # Features for big iron devices with more ram
--- a/rita_client/src/heartbeat/mod.rs
+++ b/rita_client/src/heartbeat/mod.rs
@@ -13,7 +13,6 @@
 //! This packet is encrypted using the usual LibSodium box construction and sent to the heartbeat server in the following format
 //! WgKey, Nonce, Ciphertext for the HeartBeatMessage. This consumes 32 bytes, 24 bytes, and to the end of the message
 
-use crate::rita_loop::CLIENT_LOOP_TIMEOUT;
 use althea_types::ExitDetails;
 
 use babel_monitor::get_installed_route;
@@ -26,17 +25,19 @@ use rita_common::network_monitor::get_network_info;
 use rita_common::network_monitor::GetNetworkInfo;
 use rita_common::tunnel_manager::Neighbor as RitaNeighbor;
 
-use actix::actors::resolver;
-use actix::{Arbiter, SystemService};
+use actix::System;
 use althea_types::HeartbeatMessage;
 use althea_types::Identity;
 use althea_types::WgKey;
 use babel_monitor::Neighbor as NeighborLegacy;
 use babel_monitor::Route as RouteLegacy;
-use futures01::future::Future;
 use settings::client::ExitServer;
 use sodiumoxide::crypto::box_;
 use std::collections::VecDeque;
+use std::iter::FromIterator;
+use std::net::ToSocketAddrs;
+use std::thread;
+use std::time::Instant;
 
 #[allow(unused_imports)]
 use dummy::dummy_neigh_babel;
@@ -51,7 +52,7 @@ use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
 
-type Resolver = resolver::Resolver;
+pub const HEARTBEAT_LOOP_SPEED: u64 = 5;
 
 mod dummy;
 pub struct HeartbeatCache {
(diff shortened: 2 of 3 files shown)


rita_client/src/heartbeat/mod.rs
```diff
@@ -92,9 +136,10 @@ pub fn send_udp_heartbeat() {
     }
 
     trace!("attempting to send heartbeat");
-    let dns_request = Resolver::from_registry()
-        .send(resolver::Resolve::host(heartbeat_url.to_string()))
-        .timeout(CLIENT_LOOP_TIMEOUT);
+    // let dns_request = Resolver::from_registry()
+    //     .send(resolver::Resolve::host(heartbeat_url.to_string()))
+    //     .timeout(CLIENT_LOOP_TIMEOUT);
+    let _dns_request = heartbeat_url.to_socket_addrs();
```
jkilpatr, line 142: this looks like a throwaway assignment?

rita_client/src/rita_loop/mod.rs
```diff
@@ -115,7 +115,7 @@ impl Handler<Tick> for RitaLoop {
         trace!("Client Tick!");
 
         if metrics_permitted() {
-            send_udp_heartbeat();
+            send_heartbeat_loop();
```
jkilpatr, line 118: This is going to spawn a new thread with every iteration of the loop, put it next to where we call start_rita_loop (use the new rita loop below as an example).

jkilpatr (changes requested):
This is looking good! I've left a few nits about code cleanup and one major change which is to move where we spawn the heartbeats loop to a place where it will only be called once. Currently we would spawn a new thread every 5 seconds forever running this code.

rita_client/src/heartbeat/mod.rs
```diff
@@ -78,7 +80,49 @@ lazy_static! {
         Arc::new(RwLock::new(None));
 }
 
-pub fn send_udp_heartbeat() {
+pub fn send_heartbeat_loop() {
+    let mut last_restart = Instant::now();
+    // this is a reference to the non-async actix system since this can bring down the whole process
+    let system = System::current();
```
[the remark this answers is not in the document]
ch-iara: this system is used in the error handle at the bottom of this function ( system.stop_with_code(121) ), though both of these were copied straight over from the loop I modeled this off. I'll take it off if we don't want the stop_with_code as well

rita_client/src/heartbeat/mod.rs
ch-iara: changed from a throwaway to be used in _res further down

rita_bin/Cargo.toml
```diff
@@ -52,6 +52,7 @@ jemallocator = {version = "0.3", optional = true}
 openssl = {version = "0.10", features = ["vendored"], optional = true}
 
 [features]
+default = ["bundle_openssl"]
```
jkilpatr, line 55: this looks to have snuck in from your testing and should be removed. It will bloat the size of Rita on all of our routers by move than 1mb which is quite a bit considering some devices have only 8mb of storage total, which we have to share with the operating system.

rita_client/src/heartbeat/mod.rs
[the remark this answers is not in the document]
jkilpatr: That's actually a hard call, the router is still functioning if heartbeats are not being sent, so it's not a failure condition. But the router will show as down to us even if not the user, lets leave it in and see how often it gets triggered. My guess is extremely rarely if ever.

rita_client/src/heartbeat/mod.rs
```diff
@@ -92,9 +137,7 @@ pub fn send_udp_heartbeat() {
     }
 
     trace!("attempting to send heartbeat");
-    let dns_request = Resolver::from_registry()
-        .send(resolver::Resolve::host(heartbeat_url.to_string()))
-        .timeout(CLIENT_LOOP_TIMEOUT);
+    let dns_request = heartbeat_url.to_socket_addrs();
 
     // Check for the basics first, before doing any of the hard futures work
     #[allow(unused_assignments, unused_mut)]
```
jkilpatr, line 143: this flag for unused assignments and mut should be possible to remove now without a warning. is it?

rita_client/src/heartbeat/mod.rs
```diff
@@ -106,8 +149,7 @@ pub fn send_udp_heartbeat() {
...
     let mut selected_exit_details: ExitDetails = dummy_selected_exit_details();
 
-    #[cfg(not(feature = "operator_debug"))]
-    {
+    if !cfg!(feature = "operator_debug") {
```
jkilpatr, line 152: 💯

rita_client/src/heartbeat/mod.rs
```diff
@@ -135,28 +177,43 @@ pub fn send_udp_heartbeat() {
         }
     };
 
-    let res = dns_request.then(move |res| {
+    let _res: std::result::Result<(), RitaClientError> = {
```
jkilpatr, line 180: I think I've commented this before but this block can be removed, we don't need a closure here now that it's not a future. Is it giving you trouble?

rita_client/src/heartbeat/mod.rs
```diff
@@ -135,28 +177,43 @@ pub fn send_udp_heartbeat() {
...
-                        let neigh_option =
-                            get_neigh_given_route(&route, &network_info.babel_neighbors);
-
-                        #[cfg(not(feature = "operator_debug"))]
-                        let neigh_option =
-                            get_rita_neigh_option(neigh_option, &network_info.rita_neighbors);
-                        #[cfg(feature = "operator_debug")]
-                        let neigh_option = Some((dummy_neigh_babel(), dummy_neigh_tunnel()));
+                        // #[allow(unused_variables)]
+                        // let neigh_option =
```
jkilpatr, line 201: commented out code to remove.

rita_client/src/heartbeat/mod.rs
```diff
@@ -135,28 +177,43 @@ pub fn send_udp_heartbeat() {
...
+                        } else {
+                            let neigh_option1 =
+                                get_neigh_given_route(&route, &network_info.babel_neighbors);
+                            neigh_option =
+                                get_rita_neigh_option(neigh_option1, &network_info.rita_neighbors);
+                        }
+                        // #[cfg(not(feature = "operator_debug"))]
+                        // let neigh_option =
+                        //     get_rita_neigh_option(neigh_option, &network_info.rita_neighbors);
+                        // #[cfg(feature = "operator_debug")]
```
jkilpatr, line 215: commented out code to remove.

rita_client/src/rita_loop/mod.rs
```diff
@@ -195,6 +191,9 @@ pub fn start_rita_loop() {
 
 pub fn check_rita_client_actors() {
     assert!(crate::rita_loop::RitaLoop::from_registry().connected());
+    if metrics_permitted() {
+        send_heartbeat_loop();
```
jkilpatr, line 195: 💯

jkilpatr (changes requested):
Awesome.

The main functoinality of this PR is ready to go, just a couple of final nits on the conditional compilation stuff and some commented out lines of code.

rita_bin/Cargo.toml
ch-iara: fixed

rita_client/src/heartbeat/mod.rs
ch-iara: removed

rita_client/src/heartbeat/mod.rs
ch-iara: removed

rita_client/src/heartbeat/mod.rs
ch-iara: removed

rita_client/src/heartbeat/mod.rs
ch-iara: removed

jkilpatr (approved):

PR #614, opened 2022-02-25
Migrated Hello Handler to asnyc/await

ptulugu:


rita_common/src/hello_handler/mod.rs
```diff
@@ -1,119 +1,65 @@
...
     pub to: Peer,
 }
 
-impl Message for Hello {
-    type Result = Result<(), RitaCommonError>;
-}
-
 /// Handler for sending hello messages, it's important that any path by which this handler
 /// may crash is handled such that ports are returned to tunnel manager, otherwise we end
 /// up with a port leak which will eventually crash the program
```
jkilpatr, line 27: the part of this comment talking about ports can be removed, we garbage collect now

rita_common/src/tunnel_manager/mod.rs
```diff
@@ -357,7 +345,7 @@ impl PeersToContact {
 
 /// Takes a list of peers to contact and dispatches requests if you have a WAN connection
 /// it will also dispatch neighbor requests to manual peersINGS MENTIONED
```
jkilpatr, line 347: this comment seems to be some sort of type, should be fixed

jkilpatr (approved):
This looks good to me, I've flagged two comments that need correction. 

Really great work and a very complex move made concise.

I'd appreciate if you discussed in your commit message how hello handler and peer listener are releated for the reader, as the commit title only talks about hello handler.

PR #615, opened 2022-02-28
Migrated exit dbclient and traffic watcher to async/await

ptulugu:
Once All these commits are merged, rita-bin can be migrated and additional crates and files like babel-monitor-legacy and WaitTimeout can be removed


jkilpatr (approved):
Awesome! Good to have the exit completely migrated.

PR #616, opened 2022-02-28
Completely Migrated Rita Client to asnyc await. Removed the following…

ptulugu:
Once All these commits are merged, rita-bin can be migrated and additional crates and files like babel-monitor-legacy and WaitTimeout can be removed

### DIFF
--- a/rita_client/src/heartbeat/mod.rs
+++ b/rita_client/src/heartbeat/mod.rs
@@ -25,7 +25,6 @@ use rita_common::network_monitor::get_network_info;
 use rita_common::network_monitor::GetNetworkInfo;
 use rita_common::tunnel_manager::Neighbor as RitaNeighbor;
 
-use actix::System;
 use althea_types::HeartbeatMessage;
 use althea_types::Identity;
 use althea_types::WgKey;
@@ -81,8 +80,6 @@ lazy_static! {
 
 pub fn send_heartbeat_loop() {
     let mut last_restart = Instant::now();
-    // this is a reference to the non-async actix system since this can bring down the whole process
-    let system = System::current();
 
     // outer thread is a watchdog inner thread is the runner
     thread::spawn(move || {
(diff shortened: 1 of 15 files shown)


rita_client/src/heartbeat/mod.rs
```diff
@@ -25,7 +25,7 @@ use rita_common::network_monitor::get_network_info;
 use rita_common::network_monitor::GetNetworkInfo;
 use rita_common::tunnel_manager::Neighbor as RitaNeighbor;
 
-use actix::System;
+// use actix::System;
```
jkilpatr, line 28: nit: this can be removed.

rita_client/src/heartbeat/mod.rs
```diff
@@ -114,7 +112,6 @@ pub fn send_heartbeat_loop() {
             error!("Heartbeat loop thread panicked! Respawning {:?}", e);
             if Instant::now() - last_restart < Duration::from_secs(60) {
                 error!("Restarting too quickly, leaving it to auto rescue!");
-                system.stop_with_code(121)
```
jkilpatr, line 117: Without this we won't stop the whole application at this point. The new solution will be somewhat complicated since I think we'll need thread handles for the other loops. Just modify the messages here to indicate that we simply log this error. After all I can't find any instances of this log in graylog so it's hard to even know if this is ever being called anyways.

rita_common/src/payment_validator/mod.rs
```diff
@@ -197,11 +196,7 @@ pub async fn validate() {
                 error!("{}", msg);
                 // drop the lock to prevent poisoning if we don't manage to crash
                 drop(history);
-                // get the non-async actix system and try to shut down the whole process
```
jkilpatr, line 200: Modify this to indicate that we are not crashing

jkilpatr (changes requested):
This looks really great! Excited to finally remote all these dependencies and move on. 

I've left a couple of nits, revolving around the way we previously used actix to stop the entire application in case of an issue where we wanted to crash inentionally. This no longer seems easily possible and restoring this functionality will be difficult. On the other hand I can't tell if it was ever actually used in the first place, so lets try to just remove it for now.

PR #618, opened 2022-02-28
Fix to rita exit preventing possible infinite payments from client to…

kennethAltheaSystems:
… exit

There is an issue within rita client that packet loss could result in an excess
payment since rita client assumes there is no packet loss and will always
pay the amount the exit gives even if the local amount is different.

This combined with how the payments could take a while to verify means that the
client will continually send payments over to the exit which piles ontop
of itself. This code will attempt to subtract the excess and handle this.


jkilpatr (approved):
This looks good, would appreciate some changes to the commit message with more detail about how the feedback loop works

exit is slow verifying payments -> rotuer is running at normal speed, checks in again and sees that it's got a balance (because the payment it sent last round has not been applied -> pays again.

PR #620, opened 2022-03-01
Additional async/await cleanup

ptulugu:
Completely Migrated rita bin to async/await
Removed babel_monitor_legacy crate
Removed wait_timeout module
Removed crash_actor endpoints and handlers


rita_bin/Cargo.toml
```diff
@@ -20,9 +20,6 @@ clu = { path = "../clu" }
 actix-web-httpauth = {git = "https://github.com/althea-net/actix-web-httpauth"}
```
jkilpatr, line 20: isn't this still the old version? I think it can be removed.

jkilpatr (approved):
Awesome! super glad to have this

PR #623, opened 2022-03-03
Creating stripped binary by adding strip to Cargo.toml

kennethAltheaSystems:
Simple addition to Cargo.toml to strip unnecessary information like debuginfo
from binaries to make it smaller.

### DIFF
--- a/rita_bin/Cargo.toml
+++ b/rita_bin/Cargo.toml
@@ -5,6 +5,9 @@ edition = "2018"
 license = "Apache-2.0"
 build = "build.rs"
 
+[profile.release]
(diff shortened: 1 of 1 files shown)


jkilpatr (changes requested):
Looks like github actions has not yet updated their rust, they may skip this release entierly due to a perf regression in the compiler. So this may have to sit for a while.

the desired stripping is 'all' not none.

kennethAltheaSystems:
Fixed now

jkilpatr:
looks like this just needs a clippy fix (or a rebase to get it from master)

PR #624, opened 2022-03-03
Removed migration code and old exit member struct and instances

kennethAltheaSystems:
The config has not been updated in the firmware repo and does this
automatically upon first boot. This means the old exit code is no
longer needed. It's not necessary to run every time on first boot,
so deleting this code is useful.

### DIFF
--- a/settings/src/client.rs
+++ b/settings/src/client.rs
@@ -3,9 +3,7 @@ use crate::logging::LoggingSettings;
 use crate::network::NetworkSettings;
 use crate::operator::OperatorSettings;
 use crate::payment::PaymentSettings;
-use crate::{
-    json_merge, set_rita_client, spawn_watch_thread_client, update_config, SettingsError, SUBNET,
-};
+use crate::{json_merge, set_rita_client, spawn_watch_thread_client, SettingsError};
 use althea_types::wg_key::WgKey;
 use althea_types::{ContactStorage, ExitState, Identity};
 use clarity::Address;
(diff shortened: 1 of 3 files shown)


settings/src/client.rs
```diff
@@ -123,10 +105,8 @@ fn default_balance_notification() -> bool {
 /// to a exit and to setup the exit tunnel
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
 pub struct ExitClientSettings {
-    #[serde(rename = "exits", default)]
-    pub old_exits: HashMap<String, OldExitServer>,
     /// This stores a mapping between an identifier (any string) to exits
-    #[serde(rename = "new_exits", default)]
+    #[serde(default)]
```
jkilpatr, line 109: since we have routers in the field with `old_exits` as the name for this field in the config we'll need to use the serde [alias](https://serde.rs/field-attrs.html#alias) attribute to rename it. So put `#[serde(alias = "new_exits", default)]` here.

settings/src/lib.rs
```diff
@@ -488,11 +393,12 @@ mod tests {
...
     }
 
-    #[test]
-    fn test_settings_example() {
-        let settings = RitaClientSettings::new("old_example.toml").unwrap();
-        assert!(!settings.exit_client.exits.is_empty())
-    }
+    // Deprecated test on old rita exits
+    // #[test]
+    // fn test_settings_example() {
```
jkilpatr, line 398: remove commented out code.

jkilpatr (changes requested):
This looks good! We just need to get that pr up on althea-firmware to perform the migration to the default config there. 

I think you've mixed up the commit message, it should say the config has been migrated since we're just about to do that. 

left a few other comments.

jkilpatr (approved):
Cool this looks good to me, waiting on @ptulugu  to review when he's back in.

ptulugu (approved):

PR #625, opened 2022-03-03
Restart Rita and babeld after an opkg update

ptulugu:
Currently, we are required to reboot the entire router after an optools opkg update
Restarting rita makes the update quicker

### DIFF
--- a/rita_client/src/operator_update/updater.rs
+++ b/rita_client/src/operator_update/updater.rs
@@ -15,15 +15,24 @@ pub fn update_rita(instruction: UpdateType) -> Result<Output, KernelInterfaceErr
                 info!("Set new release feed for opkg");
 
                 //opkg commands
-                let res = Err(KernelInterfaceError::RuntimeError(
+                let mut res = Err(KernelInterfaceError::RuntimeError(
                     "No commands given for opkg".to_string(),
                 ));
                 for cmd in commands.command_list {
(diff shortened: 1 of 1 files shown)


rita_client/src/operator_update/updater.rs
```diff
@@ -15,16 +15,24 @@ pub fn update_rita(instruction: UpdateType) -> Result<Output, KernelInterfaceErr
...
                     }
                 }
+
+                // Restart rita after opkg
+                let args = vec!["restart"];
+                let _res_restart = KI.run_command("/etc/init.d/babeld", &args);
+                let args = vec!["restart"];
+                let res_restart = KI.run_command("/etc/init.d/rita", &args);
+
+                info!("Restarted has error? : {:?}", res_restart);
```
jkilpatr, line 35: can we log the babeld error as well?

Also I think you can reuse the args struct. 

Awesome we got the init scripts to work!

jkilpatr (changes requested):
Awesome to see this working.

Please not in your commit message that the old update.ash rebooted the router after updating packages, the beta 18 opkg update did not, resulting in a router that was updated, but still running old software until an operator sent a manual reboot command.

This change restarts just Rita and babel, skipping the update which is what I've been doing by hand.

jkilpatr (approved):

PR #626, opened 2022-03-04
Extract wifi info

ptulugu:


althea_kernel_interface/src/hardware_info.rs
```diff
@@ -313,6 +323,74 @@ fn get_ethernet_stats() -> Option<Vec<EthernetStats>> {
...
+        .output();
+
+    if res.is_err() {
+        error!("Unable to run station dump {:?}", res);
+        return Vec::new();
+    }
+
+    let res = String::from_utf8(res.unwrap().stdout).unwrap();
+    extract_wifi_station_data(&res)
+}
```
ptulugu, line 392: These function still need to be tested

althea_types/src/wifi_info.rs
```diff
@@ -0,0 +1,471 @@
...
+    pub station: String,
+    pub inactive_time_ms: u64,
+    pub rx_bytes: u64,
+    pub rx_packets: u64,
+    pub tx_bytes: u64,
+    pub tx_packets: u64,
+    pub tx_retries: u16,
+    pub tx_failed: u16,
+    pub rx_drop_misc: u16,
+    pub signal_dbm: String,
```
jkilpatr, line 27: can't we used a signed integer here?

jkilpatr, line 27: can't we used a signed integer here?

jkilpatr:
```
root@OpenWrt:~# iw wlan1 station dump
Station <mac removed> (on wlan1)
	inactive time:	10 ms
	rx bytes:	141627376
	rx packets:	329289
	tx bytes:	1106901
	tx packets:	6172
	tx retries:	0
	tx failed:	0
	rx drop misc:	322643
	signal:  	-73 dBm
	signal avg:	-73 dBm
	tx bitrate:	6.0 MBit/s 40MHz
	rx bitrate:	54.0 MBit/s
	rx duration:	0 us
	authorized:	yes
	authenticated:	yes
	associated:	yes
	preamble:	short
	WMM/WME:	yes
	MFP:		no
	TDLS peer:	no
	DTIM period:	2
	beacon interval:100
	short preamble:	yes
	short slot time:yes
	connected time:	26651 seconds
```

Just grabbed this sample from an older router, it looks to me like some of these fields (airtime_weight, associated_at, current_time) may need to be options.

althea_types/src/wifi_info.rs
```diff
@@ -0,0 +1,471 @@
...
+}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct WifiDevice {
+    pub name: String,
+    pub survey_data: Vec<WifiSurveyData>,
+    pub station_data: Vec<WifiStationData>,
+}
+
+impl From<Vec<&str>> for WifiSurveyData {
```
jkilpatr, line 57: this is clever!

jkilpatr, line 57: this is clever!

althea_types/src/wifi_info.rs
```diff
@@ -0,0 +1,471 @@
...
+    pub short_slot_time: bool,
+    pub connected_time_sec: u16,
+    pub associated_at_boottime_sec: String,
+    pub associated_at_ms: u64,
+    pub current_time_ms: u64,
+}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct WifiDevice {
+    pub name: String,
```
jkilpatr, line 52: 👍🏻

jkilpatr, line 52: 👍🏻

althea_types/src/wifi_info.rs
```diff
@@ -0,0 +1,471 @@
...
+        let mut ret = WifiSurveyData::default();
+        populate_struct_sur(&mut ret, survey);
+        ret
+    }
+}
+
+impl From<Vec<&str>> for WifiStationData {
+    fn from(station: Vec<&str>) -> Self {
+        let mut ret = WifiStationData::default();
+        populate_struct_stat(&mut ret, station);
```
jkilpatr, line 68: can we have this function either be an associated function of WifiStationData so that it's &mut self or just return WifiStationData. Either option seems better than having to take a mutable reference.

In general I'd prefer if the function created the struct from scratch, reduces the chance of someone using an half initialized wifi data at some point in the future.

jkilpatr, line 68: can we have this function either be an associated function of WifiStationData so that it's &mut self or just return WifiStationData. Either option seems better than having to take a mutable reference.

In general I'd prefer if the function created the struct from scratch, reduces the chance of someone using an half initialized wifi data at some point in the future.

althea_types/src/interop.rs
```diff
@@ -572,6 +573,8 @@ pub struct HardwareInfo {
     pub entire_system_kernel_version: String,
     /// Vector of eth data i.e. whether a link is up and if so what the link speed is
     pub ethernet_stats: Option<Vec<EthernetStats>>,
+    // Vector of wifi devices on router with staion and survey data for each
+    pub wifi_devices: Option<Vec<WifiDevice>>,
```
jkilpatr, line 577: I believe I used an option here for the ethernet stats vec for backwards compatibility with routers that are not yet updated to send this data. In terms of actual representation an empty vec is just as good as a 'None' and requires less nesting to process.

So just flagging that this is the right call for now, but we should probably remove the option at some point (that point may be now for the ethernet stats, but that's a matter for another pr.

jkilpatr, line 577: I believe I used an option here for the ethernet stats vec for backwards compatibility with routers that are not yet updated to send this data. In terms of actual representation an empty vec is just as good as a 'None' and requires less nesting to process.

So just flagging that this is the right call for now, but we should probably remove the option at some point (that point may be now for the ethernet stats, but that's a matter for another pr.

jkilpatr (changes requested):
This is some really clever parsing logic. 

I'm very impressed at how compact this ended up but I have some concerns about the amount of unwrap. I've provided in the comments an example of a station dump from an older router that should be added as another test case. I also suggest passing some complete junk data (empty etc etc) and making sure that the parsing code doe snot panic and instead logs an error and returns an empty value.

althea_types/src/wifi_info.rs
ptulugu: signal:      -59 [-68, -59, -100, -100] dBm
    signal avg:    -58 [-67, -58, -100, -100] dBm

althea_types/src/wifi_info.rs
ptulugu: fixed

ptulugu: fixed

althea_types/src/interop.rs
ptulugu: fixed

ptulugu: fixed

jkilpatr (approved):

PR #627, opened 2022-03-07
moving contact info to lazy_static

ch-iara:
Part of a larger update to move contact info to ops tools so that it can be remotely edited.

https://trello.com/c/FgThnqNW/384-store-contact-info-in-ops-tools

### DIFF
--- a/rita_client/src/operator_update/mod.rs
+++ b/rita_client/src/operator_update/mod.rs
@@ -7,10 +7,11 @@ use crate::rita_loop::is_gateway_client;
 use crate::rita_loop::CLIENT_LOOP_TIMEOUT;
 use crate::set_router_update_instruction;
 use althea_kernel_interface::opkg_feeds::CUSTOMFEEDS;
+use althea_types::ContactDetails;
+use althea_types::ContactType;
 use rita_common::rita_loop::is_gateway;
 use rita_common::tunnel_manager::neighbor_status::get_neighbor_status;
 use rita_common::tunnel_manager::shaping::flag_reset_shaper;
-use rita_common::utils::option_convert;
 use updater::update_rita;
 
 use althea_kernel_interface::hardware_info::get_hardware_info;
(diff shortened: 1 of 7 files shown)


rita_client/src/operator_update/mod.rs
```diff
@@ -70,6 +71,11 @@ lazy_static! {
         Arc::new(RwLock::new(UptimeStruct::new()));
     static ref OPERATOR_UPDATE: Arc<RwLock<OperatorUpdate>> =
         Arc::new(RwLock::new(OperatorUpdate::default()));
+    pub static ref CONTACT_INFO: Arc<RwLock<ContactDetails>> =
```
jkilpatr, line 74: this needs a doc-comment describing the workflow, we set the contact info when the user registers to an exit or otherwise sets their contract info. At that point it's sent up to ops tools and then pulled back down from ops tools on next run.

settings/src/client.rs
```diff
@@ -151,7 +151,7 @@ impl Default for ExitClientSettings {
             exits: HashMap::new(),
             current_exit: None,
             wg_listen_port: 59999,
-            contact_info: None,
+            //contact_info: None,
```
jkilpatr, line 154: remove commented out lines

jkilpatr (changes requested):
This is a good start, we need to modify the the operator update and check in message type in althea types to include contact info.

We also need to store a sequence number in our lazy static for contact info, and add a signature (just a byte array) to the checkin message.

These are required by a mix of challenges. We need to get the latest contact info either locally or from the server, So when the user registers to an exit or sets their contact info locally we should increment the sequence number on our local contact info, that way when we send it up to operator tools ops tools can say 'oh hey this is newer' and use it.

The other way around is also true, ops tools will send us contact info with every checkin, but how do we know what's the latest? By checking the sequence number.

Finally there's the signature component, this is going to use libsodium which is also used in the exit manager, we want to take our operator update message and produce a signature and then verify it on ops tools. This lets us verify clients on ops tools and make sure we don't serve up contact info to anyone impersonating their router. The opearator update struct should have a 'create and sign checksum' function as well as a 'verify checksum' function that uses the wg key to do so.

jkilpatr:
closing as discussed on our call. We don't want to remove the local contact info capabilities, that's a bad design in the ticket on my part.

PR #628, opened 2022-03-09
Adding sequence number to contact info to enable editing from ops tools

ch-iara:
This sequence number is keeping track of the most recently updated version
of a router's user contact info. Editing it either from the router or from
ops tools will increment it and ops tools will display the most recent edit.

### DIFF
--- a/althea_types/src/contact_info.rs
+++ b/althea_types/src/contact_info.rs
@@ -17,21 +17,32 @@ pub struct ContactDetails {
 impl From<ContactType> for ContactDetails {
     fn from(val: ContactType) -> Self {
         match val {
-            ContactType::Phone { number } => ContactDetails {
+            ContactType::Phone {
+                number,
+                sequence_number: _,
+            } => ContactDetails {
                 phone: Some(number.to_string()),
                 email: None,
             },
-            ContactType::Email { email } => ContactDetails {
+            ContactType::Email {
+                email,
+                sequence_number: _,
+            } => ContactDetails {
                 phone: None,
                 email: Some(email.to_string()),
             },
-            ContactType::Both { email, number } => ContactDetails {
+            ContactType::Both {
+                email,
+                number,
+                sequence_number: _,
+            } => ContactDetails {
                 phone: Some(number.to_string()),
                 email: Some(email.to_string()),
             },
             ContactType::Bad {
                 invalid_email,
                 invalid_number,
+                sequence_number: _,
             } => ContactDetails {
                 phone: invalid_number,
                 email: invalid_email,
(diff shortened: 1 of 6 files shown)


althea_types/src/contact_info.rs
```diff
@@ -12,29 +12,45 @@ use phonenumber::PhoneNumber;
 pub struct ContactDetails {
     pub phone: Option<String>,
     pub email: Option<String>,
+    sequence_number: Option<u32>,
```
jkilpatr, line 15: so this is specifically for the exits, we don't need sequence number here because the exit always takes the latest on checkin. We'll get more into the details of this when we start to upgrade the exit format.

althea_types/src/contact_info.rs
```diff
@@ -95,37 +117,53 @@ pub struct ContactStorage {
     email: Option<EmailAddress>,
     invalid_number: Option<String>,
     invalid_email: Option<String>,
+    sequence_number: u32,
```
jkilpatr, line 120: 👍🏻  

we'll need to add serde default as a tag to this so that it starts at zero.

Serde default means that if serde does not find the field it populates it with that value, this will 'add it in' to all the existing routers configs. For ContactType we'll still use an option so ops tools can identify the older routers that are not sending a sequence number.

althea_types/src/contact_info.rs
```diff
@@ -95,37 +117,53 @@ pub struct ContactStorage {
...
+            ContactType::Both {
+                number,
+                email,
+                sequence_number,
+            } => ContactStorage {
                 number: Some(number),
                 email: Some(email),
                 invalid_number: None,
                 invalid_email: None,
+                sequence_number: sequence_number.unwrap_or(0),
```
jkilpatr, line 135: 👍🏻

rita_client/src/dashboard/contact_info.rs
```diff
@@ -14,10 +14,17 @@ fn clean_quotes(val: &str) -> String {
     val.trim().trim_matches('"').trim_matches('\\').to_string()
 }
 
+fn add_to_sequence(sequence: Option<u32>) -> u32 {
```
jkilpatr, line 17: I like this design.

rita_client/src/operator_update/mod.rs
```diff
@@ -209,6 +209,10 @@ async fn checkin() {
 
     let mut rita_client = settings::get_rita_client();
 
+    //match new_settings.contact_info {
+    //Some(sequence) =>
```
jkilpatr, line 213: here we should take and store the new contact info if it has the higher sequence.

jkilpatr (changes requested):
This is very much on the right track. I like the rust syntax you've used here, clean and making good use of matching.

The logic to actually update the contact info in the operator update needs to be filled out, so that will require adding a contactType struct to the operator update message.

althea_types/src/contact_info.rs
ch-iara: This was mostly because there's a conversion function going from ContactDetails to ContactType. I've just changed the conversion function to take in sequence_number as an argument instead.

jkilpatr (approved):
This looks great! I really like the operator update cleanup as well.

PR #630, opened 2022-03-18
Assign global ipv6 address to clients who connect

ptulugu:
Basic PR that takes subnet from config, generates unique ip and adds to database
1.) Assumes there is a subnet specified in rita exit config
2.) Assumes all instances share a singular database, or if mulitple, all the data is the same
3.) Uses ipgen library for generating random ip
4.) Possibility of a race condition? (Two instances with exact same subnet assign exact same ip to two clients at the same time and write same ip to the database. Seems unlikely but would this be an issue?)


exit_db/src/models.rs
```diff
@@ -8,6 +8,7 @@ pub struct Client {
     pub wg_port: i32,
     pub eth_address: String,
     pub internal_ip: String,
+    pub global_ipv6: String,
```
jkilpatr, line 11: maybe internet_ipv6?  or just a comment indicating it's an internet routable ipv6 address assigned to this router.

jkilpatr: also this is a subnet, so it should include client subnet size as part of the string encoding. This makes the 'pick a non-overlapping subnet to assign to a new client' problem extremely hard becuase what if subnet assignments changed since first startup? We could maybe ignore that case and just say 'all client subnet sizes are always the same all the time' and put a value in the config.

rita_exit/src/database/database_tools.rs
```diff
@@ -15,6 +17,9 @@ use exit_db::{models, schema};
 use std::net::IpAddr;
 use std::net::Ipv4Addr;
 
+/// Max number of time we loop to find an available ip before we quit
+const MAX_DEPTH: u8 = 25;
```
jkilpatr, line 21: we should avoid recursion in production, it's unpredictable, the algorithm should assign Subnets to clients in a predictible way and allow them to be returned to the pool for later allocation.

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +345,95 @@ pub fn create_or_update_user_record(
...
 
         info!("Inserting new client {}", client.global.wg_public_key);
         diesel::insert_into(clients).values(&c).execute(conn)?;
 
         Ok(c)
     }
 }
+
+/// Out of the subnet provided to the exit, choose an ip that is not in use
+/// Randomly select ips in the subnet until we get a unique one that is not in the database
```
jkilpatr, line 360: 1) recursion is probably not the best way to do this
2) we need to make sure subnets don't overlap (so a random IP could be within another clients subnet)

jkilpatr (changes requested):
1.) 👍🏻 
2.) 👍🏻 
3.) Like I say in the comments I don't think random is an appropriate way to handle this.
4.) This shouldn't happen as the client will select only one exit at a time and therefore only talk to one at a time.

This is a great start for the database parts. The IP address and subnet assignment components are going to require significantly more logic and unit tests to work well. Remember we're assigning a client *subnet* so we need to deal with logic around their subnet size and overlapping subnets.

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +379,216 @@ pub fn create_or_update_user_record(
...
+        let addr = addr.unwrap();
+        if !ip_list.contains(&addr.to_string()) {
+            info!("Reclaimed an ip subnet! {:?}", addr);
+            return Ok(addr);
+        }
+    }
+
+    // Next get iterative index to generate next subnet. The reason we loop here is that there is a possiblity that a different exit instance has
+    // the same subnet as us and has already assigned some iterative ip addrs that we have not
+    for _ in 0..LOOP_THRESH {
```
jkilpatr, line 430: There's no database access in this loop, so I'm not sure how it prevents the case described in the comment. There's still a race condition that could occur while this loop is running so we would want to do a database check before and after allocation to prevent collision.

But like I said before I don't think we need to worry about this too much.

jkilpatr (changes requested):
This is nice, compact and clean. I like the comments and low-touch changes.

I do think the assignment process can have the lazy static removed, we already check the db on every allocation so I'm not sure what the advantage of having local state is. Compared to checking the DB iterating over subnets and determining what is available directly from the database state is an extremely fast operation on the local CPU, there's no need to try and optimize it. 

I also think the loop to avoid ip assignment collisions needs to be thought through more carefully. I don't expect collisions to be much of an issue, but this design wouldn't really prevent them either. We could do a 'proper' job of this by having for example an integer field for the last IP assigned, increment it atomicly and take the number returned by postgres as the subnet to assign the user. But I'm frankly very weak when it comes to SQL so my advice may not be the best.

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +379,216 @@ pub fn create_or_update_user_record(
...
         diesel::insert_into(clients).values(&c).execute(conn)?;
 
         Ok(c)
     }
 }
+
+/// This function finds an available ipv6 subnet for a client that connects. It works as follows:
+/// 1.) Find assigned subnet and client configured subnet length (currently hardcoded to CLIENT_SUBNET_LENGTH)
+/// 2.) Retreive all active subnets from database
+/// 3.) Look at lazy static for any subnets in stack (This stack get populated when a client get deleted)
```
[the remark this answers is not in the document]
ptulugu: fixed

rita_exit/src/database/database_tools.rs
```diff
@@ -270,6 +296,19 @@ pub fn delete_client(client: ExitClient, connection: &PgConnection) -> Result<()
...
+
+    // Add the reclaimed subnet to available subnets
+    let filtered_list = clients
+        .select(internet_ipv6)
+        .filter(mesh_ip.eq(&mesh_ip_string));
+    let mut subnet = filtered_list.load::<String>(connection)?;
+    if let Some(subnet) = subnet.pop() {
+        let client_ip_tracker = &mut *CLIENT_IP_TRACKER.write().unwrap();
+        client_ip_tracker
+            .available_subnets
```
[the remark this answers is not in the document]
ptulugu: Looking at database now

rita_exit/src/database/database_tools.rs
ptulugu: Changed the logic to only pull from a new database table, removing the need for iteration anywhere.
1.) Retrieve subnet struct from database 
2.) Use reclaimed index from deleted client, else choose iterative index
3.) Use index to generate a subnet and update all info in databases

One thing I havent tested is whether my database queries are valid or not

exit_db/src/models.rs
```diff
@@ -19,3 +21,17 @@ pub struct Client {
     pub last_seen: i64,
     pub last_balance_warning_time: i64,
 }
+
+/// This struct holds information about the ipv6 subnets being assigned to clients who connect.
+/// The vector available subnets is a stack that has a list of available subnets to use. This stack gets populated whenever
```
jkilpatr, line 26: really like this comment 💯

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +410,366 @@ pub fn create_or_update_user_record(
...
+
+/// This function creates an entry for the given subnet in the assgined_ips table if doesnt exist
+fn initialize_subnet_datastore(
+    sub: IpNetwork,
+    conn: &PgConnection,
+) -> Result<models::AssignedIps, RitaExitError> {
+    use self::schema::assigned_ips::dsl::{assigned_ips, subnet};
+    let filtered_list = assigned_ips.filter(subnet.eq(sub.to_string()));
+    match filtered_list.load::<models::AssignedIps>(conn) {
+        Err(_) => {
```
jkilpatr, line 432: comment thse cases otherwise logic looks good

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +410,366 @@ pub fn create_or_update_user_record(
...
+                iterative_index: 0,
+            };
+            diesel::insert_into(assigned_ips)
+                .values(&record)
+                .execute(conn)?;
+            Ok(record)
+        }
+        Ok(mut a) => {
+            if a.len() > 1 {
+                error!("More than one entry for singular subnet in database, please fix");
```
jkilpatr, line 445: if this is a failure condition should we maybe panic instead of just logging?

rita_exit/src/database/database_tools.rs
```diff
@@ -348,11 +410,366 @@ pub fn create_or_update_user_record(
...
+    let mut net = network.network();
+    net.prefix = net.prefix.from(subprefix).unwrap();
+
+    if subprefix < network.prefix.num {
+        return Err(RitaExitError::MiscStringError(
+            "Client subnet larger than exit subnet".to_string(),
+        ));
+    }
+
+    if ind < (1 << (subprefix - network.prefix.num)) {
```
jkilpatr, line 590: can you comment what the bitshifting is doing?

jkilpatr (approved):
Just a few comment requests, this logic is looking solid.

rita_exit/src/database/database_tools.rs
ptulugu: fixed

rita_exit/src/database/database_tools.rs
ptulugu: This wont break the functinoality of the program. There should be only one entry, but there will be two entries if there is an 
unlikely race condition, the code will still work as normal, except it will miss some available subnets. Should I still panic in this case?

PR #631, opened 2022-03-21
Optimize writes to disk

kennethAltheaSystems:


rita_common/src/rita_loop/fast_loop.rs
```diff
@@ -55,30 +53,30 @@ pub fn start_rita_fast_loop() {
...
-                            );
-                        }
-                    }
+                    // let neighbors = res;
+                    // let neigh = Instant::now();
+                    // info!(
+                    //     "GetNeighbors completed in {}s {}ms",
+                    //     start.elapsed().as_secs(),
+                    //     start.elapsed().subsec_millis()
+                    // );
```
jkilpatr, line 66: this all needs to be uncommented.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,121 @@
...
+                info!(
+                    "TrafficWatcher completed in {}s {}ms",
+                    neigh.elapsed().as_secs(),
+                    neigh.elapsed().subsec_millis()
+                );
+            }
+        }
+    });
+}
+pub fn is_router_memory_small(router_model: &str) -> bool {
```
jkilpatr, line 107: this function seems to be duplicated from rita_bin, we should call this one from rita_bin (the other review comment about adding a doc comment and renaming it still applies)

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,121 @@
...
+                }
+                match new_settings.network.device {
+                    Some(val) => router_mem_small = is_router_memory_small(&val),
+                    None => router_mem_small = false,
+                }
+            }
+        }
+
+        // debt keeper
+        if !router_mem_small {
```
jkilpatr, line 69: this should document why we disable debt keeper on small storage devices, also should be router_storage_small not mem.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -362,9 +371,9 @@ impl DebtKeeper {
 
     fn save(&mut self) -> Result<(), IOError> {
         // convert to the serializeable format and dump to the disk
-        let serialized = serde_json::to_string(&debt_data_to_ser(self.debt_data.clone()))?;
+        let serialized = bincode::serialize(&debt_data_to_ser(self.debt_data.clone())).unwrap();
         let mut file = File::create(settings::get_rita_common().payment.debts_file)?;
-        file.write_all(serialized.as_bytes())
```
jkilpatr, line 367: you seem to have modified only teh save routine, how will we load bincode encoded data if we're still trying to decode it using json? Also we have to take into account that what's on the disk is currently json and migrate it.

So there needs to be some logic that 

1) tries to deserialize with both json and bincode, takes what's successful (the json path goes away next release once every is migrated)
2) renames the file in the settings to .bincode rather than .json

Also save if needed should be called from the saving loop now so that we don't need to put the save frequency constants in this file and we can keep them in one place.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,121 @@
...
+            runner.block_on(async move {
+                if let Err(e) = send_debt_update() {
+                    warn!("Debt keeper update failed! {:?}", e);
+                }
+                info!("Config/settings is being saved to");
+            });
+        }
+
+        //usage tracker
+        let babel_port = settings::get_rita_common().network.babel_port;
```
jkilpatr, line 80: nothing below this line seems to actually call usage tracker?

rita_client/src/lib.rs
```diff
@@ -69,10 +71,11 @@ impl Default for Args {
 /// and does not need to be specified.
 pub fn get_client_usage(version: &str, git_hash: &str) -> String {
     format!(
-        "Usage: {} [--config=<settings>] [--platform=<platform>] [--future]
+        "Usage: {} [--config=<settings>] [--platform=<platform>] [--save_interval=<time>] [--future]
```
jkilpatr, line 74: arguments are pretty hard for us to modify on the routers, much better to remove this and use a config value.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,121 @@
...
+use settings::{
+    client::RitaClientSettings, exit::RitaExitSettingsStruct, get_rita_client, get_rita_exit,
+    FileWrite,
+};
+
+pub const FAST_LOOP_TIMEOUT: Duration = Duration::from_secs(4);
+// Save duration for all writes to disk in order to reduce write operations
+// pub const SAVING_TO_DISK_FREQUENCY: Duration = Duration::from_secs(600);
+#[derive(Clone)]
+pub enum OldSettingsType {
```
jkilpatr, line 20: In this case you're using 'old' to mean 'the settings on the disk' this took a couple of minutes to figure out. So the name/comment needs some improvement to indicate that maybe SettingsOnDisk?

Also I don't think we should pass the settings on the disk as an argument, nor the save interval. Only the file path since it's an argument.

Right now this loop never updates the old settings so once it' saves once it will keep saving again every loop iteration forever, you should declare a variable with the config on the disk at the top of the loop and update it every time you save. Save interval should come from the config.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -116,8 +117,47 @@ pub struct UsageTracker {
...
+        b && c
+    }
+}
+
 impl UsageTracker {
     fn save(&mut self) -> Result<(), RitaCommonError> {
+        let minimum_number_transactions = 75;
+        if self.payments.len() < minimum_number_transactions
+            || at_least_two(
+                check_usage_hour(&mut self.client_bandwidth, self.last_save_hour),
```
jkilpatr, line 150: this should also be called from the saving loop and the number of tx's before a save should be an argument there.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -130,6 +170,7 @@ impl UsageTracker {
         loop {
             match file.write_all(&compressed_bytes) {
                 Ok(save) => {
+                    info!("Saved to disk for usage tracker");
```
jkilpatr, line 173: add the number of bytes compressed_bytes.len() to this message. as well as all saving log messages

settings/src/lib.rs
```diff
@@ -297,6 +295,16 @@ pub fn get_rita_exit() -> RitaExitSettingsStruct {
     }
 }
 
+/// Retrieve exit settings from memory
```
jkilpatr, line 298: this is what you made the function for, not what the function does.

settings/src/lib.rs
```diff
@@ -315,50 +323,26 @@ pub fn json_merge(a: &mut Value, b: &Value) {
...
-                trace!("writing updated config: {:?}", new_settings);
-                if let Err(e) = new_settings.write(&file_path) {
-                    warn!("writing updated config failed {:?}", e);
-                }
-                old_settings = new_settings.clone();
-            }
-        }
-    });
-}
+// fn spawn_watch_thread_exit(settings: RitaExitSettingsStruct, file_path: &str) {
```
jkilpatr, line 326: this commented out code can be removed.

jkilpatr (changes requested):
This is a good start, but it needs significant design work before it's really ready to move forward. We're trying to pull together a lot of differnet types and methods of saving from across rita and make them a look consistent and be called from a single location, in addition to the individual tunings we are making.

rita_common/src/usage_tracker/mod.rs
kennethAltheaSystems: The saving loop is really far abstracted from actually touching the transactions. There wasn't a clean way to count the total transactions and prevent bad saves.

rita_common/src/rita_loop/write_to_disk.rs
kennethAltheaSystems: The issue was that calling this loop function needs the ability to discriminate between if the device was a client or an exit. I thought passing in the enum was the best way since if you just pass the file path I wasn't sure how to tell if the device was a client or exit. I think if you call settings::get_rita_common it isn't exactly clear how to tell what it is.

jkilpatr: I think in this case the best option may be to add a function in settings that returns an enum. I agree the enum is needed, but I don't think it needs to be an argument to the saving loop.

rita_common/src/usage_tracker/mod.rs
jkilpatr: the transaction counting logic is fine, but this save function should be called in the saving loop.

jkilpatr (changes requested):
Updating with more comments.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub const FAST_LOOP_TIMEOUT: Duration = Duration::from_secs(4);
+// Save duration for all writes to disk in order to reduce write operations
+// pub const SAVING_TO_DISK_FREQUENCY: Duration = Duration::from_secs(600);
+#[derive(Clone)]
+pub enum SettingsOnDisk {
+    RitaClientSettings(RitaClientSettings),
+    RitaExitSettingsStruct(RitaExitSettingsStruct),
+}
+pub fn save_to_disk_loop(mut old_settings: SettingsOnDisk, file_path: &str) {
```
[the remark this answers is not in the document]
kennethAltheaSystems: The reason it was commented was because I moved the logic into write_disk_loop

settings/src/client.rs
```diff
@@ -192,6 +194,8 @@ pub struct RitaClientSettings {
     pub future: bool,
     #[serde(default = "default_app_name")]
     pub app_name: String,
+    #[serde(default = "default_save_interval")]
+    pub save_interval: u64,
```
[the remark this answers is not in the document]
kennethAltheaSystems: The save function is indirectly invoked in the watch function and inside the validate.await.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,138 @@
...
+        match old_settings.clone() {
+            SettingsOnDisk::RitaClientSettings(old_settings_client) => {
+                let new_settings = get_rita_client();
+                saving_to_disk_frequency = Duration::from_secs(new_settings.save_interval);
+
+                if old_settings_client != new_settings {
+                    if let Err(e) = new_settings.write(&file_path) {
+                        warn!("writing updated config failed {:?}", e);
+                    }
+                    info!("Config/settings is being saved to");
```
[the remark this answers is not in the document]
kennethAltheaSystems: Where did you want me to log this to?

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        let router_storage_small;
+        //settings
+        match old_settings.clone() {
+            SettingsOnDisk::RitaClientSettings(old_settings_client) => {
+                let new_settings = get_rita_client();
+                saving_to_disk_frequency = Duration::from_secs(new_settings.save_interval);
+
+                if old_settings_client != new_settings {
```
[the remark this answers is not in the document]
kennethAltheaSystems: Where is balance located? I couldn't find it.

rita_common/src/rita_loop/write_to_disk.rs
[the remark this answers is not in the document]
jkilpatr: just put it in the info log statement

settings/src/client.rs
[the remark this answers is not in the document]
jkilpatr: this variable needs a doc comment saying as much.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -34,8 +37,10 @@ use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use std::time::Instant;
 
-/// How often we save the nodes debt data, currently 30 minutes
-const SAVE_FREQUENCY: Duration = Duration::from_secs(1800);
+/// How often we save the nodes debt data, currently 48 minutes
+const SAVE_FREQUENCY_EXIT: Duration = Duration::from_secs(172800);
+/// How often we save the nodes debt data, currently 48 minutes
```
jkilpatr, line 42: comment time is incorrect, should be 5 minutes.

jkilpatr (changes requested):

rita_client/src/heartbeat/mod.rs
```diff
@@ -344,7 +344,7 @@ fn send_udp_heartbeat_packet(
     let message = HeartbeatMessage {
         id: our_id,
         organizer_address: settings::get_rita_client().operator.operator_address,
-        balance: payment.balance.clone(),
+        balance: payment.balance.clone().unwrap_or_else(|| 0u64.into()),
```
jkilpatr, line 347: we should modify heartbeat message to send an option.

rita_common/src/blockchain_oracle/mod.rs
```diff
@@ -195,7 +195,15 @@ async fn update_blockchain_info(our_address: Address, web3: Web3, full_node: Str
         join4(balance, nonce, net_version, gas_price).await;
     let mut settings = settings::get_rita_common();
     match balance {
-        Ok(balance) => update_balance(&full_node, &mut settings.payment.balance, balance),
+        Ok(balance) => update_balance(
+            &full_node,
+            &mut settings
```
jkilpatr, line 200: this won't work the way you expect it to. modify update balance to return a value and set settings.payment.balance in this scope.

The issue is your passing a mutable reference, but if its None you'll just pass a local var, it will be modifed and the result will not propogate into the settings struct.

rita_common/src/blockchain_oracle/mod.rs
```diff
@@ -326,7 +334,7 @@ pub fn low_balance() -> bool {
     let balance = payment_settings.balance.clone();
     let balance_warning_level = payment_settings.balance_warning_level;
 
-    balance < balance_warning_level
+    balance.unwrap_or_else(|| 0u64.into()) < balance_warning_level
```
jkilpatr, line 337: if the balnce is none this function should retun false and not do the comparison. Becuase we don't know if we have a low balance.

rita_common/src/dashboard/own_info.rs
```diff
@@ -41,7 +41,7 @@ pub async fn get_own_info(_req: HttpRequest) -> HttpResponse {
 
     let reply = OwnInfo {
         address: eth_address,
-        balance,
+        balance: balance.unwrap_or_else(|| 0u64.into()),
```
jkilpatr, line 44: own info should be modified to have an option for the balance as well.

rita_common/src/dashboard/wallet.rs
```diff
@@ -40,7 +40,7 @@ fn withdraw_handler(address: Address, amount: Option<Uint256>) -> HttpResponse {
     let mut amount = if let Some(amount) = amount {
         amount
     } else {
-        balance.clone()
+        balance.clone().unwrap_or_else(|| 0u64.into())
```
jkilpatr, line 43: if balance is none return an error that we're not yet ready to withdraw. (that should probably be above line 40 becuase we can't send any amount if we don't yet know our balance.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -297,19 +316,29 @@ impl Default for DebtKeeper {
         match file {
             Ok(mut file) => {
                 let mut contents = String::new();
+                let file_copy = file.try_clone().unwrap();
+                let deserialized_binary = deserialize_from_binary(common, file_copy);
                 match file.read_to_string(&mut contents) {
```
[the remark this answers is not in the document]
jkilpatr: bumping this comment.

rita_common/src/rita_loop/fast_loop.rs
```diff
@@ -111,9 +110,9 @@ pub fn start_rita_fast_loop() {
                     }
 
                     // Update debts
-                    if let Err(e) = send_debt_update() {
-                        warn!("Debt keeper update failed! {:?}", e);
-                    }
+                    // if let Err(e) = send_debt_update() {
```
jkilpatr, line 113: this should be uncommented.

rita_common/src/payment_controller/mod.rs
```diff
@@ -153,7 +153,7 @@ async fn make_payment(mut pmt: PaymentTx) -> Result<(), PaymentControllerError>
         "current balance: {:?}, payment of {:?}, from address {} to address {} with nonce {}",
         balance, pmt.amount, our_address, pmt.to.eth_address, nonce
     );
-    if balance < pmt.amount {
+    if balance.clone().unwrap_or_else(|| 0u64.into()) < pmt.amount {
```
jkilpatr, line 156: handle the balcne is none case above this, just return the same payment controller error with a different log statement. What you have here is correct in that it will do the right thing but handling the bootstrapping case explicitly is more clear and lets us log that the node is stuck in the bootstrapping case which is important info if a bug happens there.

rita_common/src/rita_loop/write_to_disk.rs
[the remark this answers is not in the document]
jkilpatr: Good comments 👍🏻

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,151 @@
...
+                info!("Config/settings is being saved to");
+            });
+        }
+
+        //usage tracker is invoked via trafficwatch/watch() and the block runner
+        let babel_port = settings::get_rita_common().network.babel_port;
+        trace!("Common tick!");
+
+        let res = tm_get_neighbors();
+        trace!("Currently open tunnels: {:?}", res);
```
jkilpatr, line 103: this can all be removed you're not running usage tracker here? You should really only have to call it's save function here you can leave the setup work and normal calling in the fast loop.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,151 @@
...
+            // Check on payments, only really needs to be run this quickly
+            // on large nodes where very high variation in throughput can result
+            // in blowing through the entire grace in less than a minute
+            validate().await;
+        });
+    });
+}
+/// If the router storage space is 16MB/ it's small,
+/// we want to prevent high amounts of writes to the
+/// disk as to not run out of space.
```
jkilpatr, line 136: this isn't about running out of space, it's about running out of write endurance and having the drive fail.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -116,8 +117,47 @@ pub struct UsageTracker {
...
+    if a {
+        b || c
+    } else {
+        b && c
+    }
+}
+
 impl UsageTracker {
     fn save(&mut self) -> Result<(), RitaCommonError> {
+        let minimum_number_transactions = 75;
```
jkilpatr, line 147: this save function should be called in the save_to_disk_loop

settings/src/payment.rs
```diff
@@ -107,8 +107,8 @@ pub struct PaymentSettings {
     pub eth_private_key: Option<PrivateKey>,
     // Our own eth Address, derived from the private key on startup and not stored
     pub eth_address: Option<Address>,
-    #[serde(default)]
-    pub balance: Uint256,
+    #[serde(skip_serializing)]
+    pub balance: Option<Uint256>,
```
jkilpatr, line 111: Add the balance as an option here and remove it from the config https://github.com/althea-net/althea_rs/blob/master/rita_common/src/blockchain_oracle/mod.rs#L76

the skip serializing is the right flag and will do what we want. So good call there, but if we don't need to deserialize it ever we should put it somewhere else.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -361,10 +414,15 @@ impl DebtKeeper {
     }
 
     fn save(&mut self) -> Result<(), IOError> {
+        let mut file_path: String = settings::get_rita_common().payment.debts_file;
         // convert to the serializeable format and dump to the disk
-        let serialized = serde_json::to_string(&debt_data_to_ser(self.debt_data.clone()))?;
-        let mut file = File::create(settings::get_rita_common().payment.debts_file)?;
-        file.write_all(serialized.as_bytes())
+        if file_path.ends_with("json") {
+            file_path.drain(0..file_path.len() - 4);
```
jkilpatr, line 420: that's an interesting way to do this. 👍🏻  You should put the new file path string into the settings.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,151 @@
...
+
+                old_settings = SettingsOnDisk::RitaExitSettingsStruct(new_settings);
+            }
+        }
+
+        // debt keeper
+        if !router_storage_small {
+            let runner = AsyncSystem::new();
+            runner.block_on(async move {
+                if let Err(e) = send_debt_update() {
```
jkilpatr, line 91: call the save function *only* there's no need to send a debt update, while that triggers saving right now if enough time has elapsed the point of this pr is to move that timer and saving handling into this thread and not call it from there.

rita_common/src/blockchain_oracle/mod.rs
```diff
@@ -162,6 +164,10 @@ pub fn get_oracle_net_version() -> u64 {
     ORACLE.read().unwrap().net_version
 }
 
+pub fn get_oracle_balance() -> Option<Uint256> {
```
jkilpatr, line 167: +1 like the getter / setter here

rita_common/src/dashboard/wallet.rs
```diff
@@ -54,8 +58,8 @@ fn withdraw_handler(address: Address, amount: Option<Uint256>) -> HttpResponse {
         };
 
     let tx_cost = gas_price * tx_gas;
-    if amount.clone() + tx_cost.clone() >= balance {
-        amount = balance - tx_cost;
+    if amount.clone() + tx_cost.clone() >= balance.clone().unwrap_or_else(|| 0u64.into()) {
+        amount = balance.unwrap_or_else(|| 0u64.into()) - tx_cost;
```
jkilpatr, line 62: this should handle the none case explicitly so it can send a proper 'not ready' error message rather than trying to create a transaction with a value of zero, which can actually execute and cost gas money.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -34,13 +37,15 @@ use std::sync::{Arc, RwLock};
...
+/// How often we save the nodes debt data, currently 48 minutes
+const SAVE_FREQUENCY_EXIT: Duration = Duration::from_secs(172800);
+/// How often we save the nodes debt data, currently 5 minutes
+const SAVE_FREQUENCY_ROUTER: Duration = Duration::from_secs(300);
 
 lazy_static! {
     /// A locked global ref containing the state for this module. Note that the default implementation
     /// loads saved data from teh disk if it exists.
-    static ref DEBT_DATA: Arc<RwLock<DebtKeeper>> = Arc::new(RwLock::new(DebtKeeper::default()));
+    pub static ref DEBT_DATA: Arc<RwLock<DebtKeeper>> = Arc::new(RwLock::new(DebtKeeper::default()));
```
jkilpatr, line 48: this shouldn't have to be made public

rita_common/src/debt_keeper/mod.rs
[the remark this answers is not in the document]
jkilpatr: bumping again.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,102 @@
...
+
+                old_settings = SettingsOnDisk::RitaExitSettingsStruct(new_settings);
+            }
+        }
+
+        // debt keeper
+        if !router_storage_small {
+            let mut dk = DEBT_DATA.write().unwrap();
+            trace!("sending debt keeper update");
+            dk.save_if_needed();
```
jkilpatr, line 74: to avoid making DEBT_DATA public just have save_if_needed as a independent function rather than a member function of the struct and it can grab the DEBT_DATA.write() there.

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        //usage tracker is invoked via trafficwatch/watch() and the block runner
+        match USAGE_TRACKER.write().unwrap().save() {
+            Ok(_val) => log::info!("Saved usage tracker successfully"),
```
jkilpatr, line 79: same consideration here, we shouldn't need to make USAGE_TRACKER writable from anywhere. I like to keep these global locks private to encourage the use of getters/setters which make deadlocks harder (prevents deadlocks between two modules and ensures that deadlocks only happen within one module)

jkilpatr (changes requested):
Good updates here. I'm very happy with how the balance storage changes where handled. One minor comment in wallet.rs to return a proper error message, otherwise the frontend won't be able to display it properly. 

the save_to_disk_loop function is now appropriately constrained. I'd like to avoid making the lazy static variables 'pub' if possible so you can change up the save functions a bit to prevent that.

The debt keeper load logic still needs attention/cleanup I've commented on that a few times, both load types should be tried and then matched on at the same time, prevents nesting. 

The debt keeper constants about how often to save should be in the write_to_disk_.rs file I think and passed as args to the saving function that way all the factors for saving that are simple constants and not computed values are visible there. 

I'm h

rita_common/src/dashboard/wallet.rs
```diff
@@ -54,8 +58,11 @@ fn withdraw_handler(address: Address, amount: Option<Uint256>) -> HttpResponse {
         };
 
     let tx_cost = gas_price * tx_gas;
-    if amount.clone() + tx_cost.clone() >= balance {
-        amount = balance - tx_cost;
+    if amount.clone() + tx_cost.clone() >= balance.clone().unwrap_or_else(|| 0u64.into()) {
```
jkilpatr, line 61: this if statement is a little overcomplicated, your unwrapping and then matching on the balance, the other way around would be cleaner.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -617,6 +680,13 @@ impl DebtKeeper {
     }
 }
 
+/// Saves the debt keeper to disk
+pub fn save_debt_to_disk() {
```
jkilpatr, line 684: This is correct in that it avoids DEBT_DATA being public. 

I would prefer if the 'if needed' part of save if needed was decided up in write_to_disk.rs not only for this but for all the other save functions.

rita_common/src/usage_tracker/mod.rs
```diff
@@ -116,8 +117,55 @@ pub struct UsageTracker {
     payments: VecDeque<PaymentHour>,
 }
 
+fn check_usage_hour(input: &mut VecDeque<UsageHour>, last_save_hour: u64) -> bool {
```
jkilpatr, line 120: this function is good, but needs a doc comment describing exactly what it does.

rita_common/src/usage_tracker/mod.rs
jkilpatr: This is called in save_to_disk_loop now 👍🏻 

make the number of transactions and bytes that need to be downloaded/uploaded into arguments, this way we can set different thresholds up in save_to_disk_loop

rita_common/src/rita_loop/write_to_disk.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        if start.elapsed() < saving_to_disk_frequency {
+            thread::sleep(saving_to_disk_frequency - start.elapsed());
+        }
+
+        //settings
+        match old_settings.clone() {
+            SettingsOnDisk::RitaClientSettings(old_settings_client) => {
+                let new_settings = get_rita_client();
+                save_client_settings(old_settings_client, file_path.clone());
```
jkilpatr, line 59: in the save_client_settings and save_exit_settings function an equality check is done to make sure that the settings have changed. Move that up to this level.

jkilpatr (changes requested):
This is looking good, no more pub lazy statics.

One final set of changes to make, which is to move control over all saving parameters up into write_to_disk_loop the idea is that the save functions should just save and any arguments/discretion about how they should operate should be centralized in write_to_disk.rs where it can be easily observed and tweaked.

rita_common/src/debt_keeper/mod.rs
kennethAltheaSystems: The issue here is that the save_if_needed for usage tracker is called in send_usage_update. Should I attempt to modify it there as well in order to make this more modularized?

jkilpatr: do not call save_if_needed in the send _usage_update function, that's one of the goals of this change is to only call the saving functions from wtire_to_disk.rs

rita_common/src/usage_tracker/mod.rs
```diff
@@ -427,8 +455,6 @@ pub fn save_usage_on_shutdown() {
     let history = &mut USAGE_TRACKER.write().unwrap();
     history.last_save_hour = current_hour;
     trim_payments(MAX_TX_ENTRIES, &mut history.payments);
-    let res = history.save();
-    info!("Shutdown: saving usage data: {:?}", res);
```
jkilpatr, line 431: shutdown save for usage tracker and debt keeper should be kept. Routers in the field rarely get explicit restart instructions. I guess this is is the one exception to the push for saving all in one place, we can check if the router has small storage and disable it then.

jkilpatr (approved):
This is good to go just two last changes

1) rename the title from WIP as this is now done
2) for the debt keeper/usage tracker save on shutdown routines, I did ask you to remove all saves outside of write_to_disk loop so good call catching them, but we do actually need those, disable them only on a small storage router.

PR #632, opened 2022-03-22
Changed alias for exits to resolve serde issue

kennethAltheaSystems:
Apparently, serde doesn't override a value and will instead
create duplicate key values when you alias something. In order,
to fix this the alias was changed to something that doesn't exi-
st.


settings/src/client.rs
```diff
@@ -106,7 +106,7 @@ fn default_balance_notification() -> bool {
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
 pub struct ExitClientSettings {
     /// This stores a mapping between an identifier (any string) to exits
-    #[serde(alias = "new_exits", default)]
+    #[serde(alias = "new_exits_temp", default)]
```
jkilpatr, line 109: this won't work, it needs to be just a "rename" with "new_exits". That will 'rename' the field when it loads it to just exits but keep it as new_exits in the config.

What you're doing here is interesting though, if we named our field new_exits and aliased to new exits tmp it might work. But I dont' think that's the way to go.

jkilpatr (changes requested):
Love the commit message 👍🏻 

we need a slightly different serde tag though, left a comment.

jkilpatr (approved):

PR #635, opened 2022-03-29
Bundle OpenSSL by default

jkilpatr:
Bundle openssl by default

Recently due to cross compile issues in our package repo we moved to
bundling openssl for our openwrt release builds. As it turns out this
doesn't really increase binary size as much as we thought. In fact, the
images end up slightly smaller.

This may be because what's added is easily compressed by squashfs or
because we're negating some advantages from the recent removal or legacy
futures and it's dependency tree. But for the time being we're moving to
bundled openssl in order to unify our cross compile config.

Strip flag needs to be in top level workspace

This binary strip flag was not effective since profile settings must be
performed at the top level of the workspace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,7 +47,7 @@ dependencies = [
  "pin-project-lite",
  "smallvec",
  "tokio",
- "tokio-util 0.7.0",
+ "tokio-util 0.7.1",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


kennethAltheaSystems (approved):
Looks good

ch-iara (approved):
👍

PR #638, opened 2022-04-07
Remove selectedId field from config

ptulugu:
Now with the new exit subnetting feature, having persistant-keepalives from every exit
caused an issue where the router would renegotiate a tunnel without that exit being selected.
This was not an issue previously where all exits shared an ip
Removing this ensures keepalives to be sent only from the client side to the exit they are connected to

### DIFF
--- a/rita_client/src/exit_manager/exit_switcher.rs
+++ b/rita_client/src/exit_manager/exit_switcher.rs
@@ -11,14 +11,18 @@
 //! 4.) Switch only if another exit has been considered better than our current exit for an extended period of time.
 //!
 //! See doc comment for 'set_best_exit' for a more detailed description of workflow
-use crate::exit_manager::{reset_exit_blacklist, EXIT_MANAGER};
+use crate::exit_manager::{
+    get_current_selected_exit, get_current_selected_exit_degradation,
+    get_current_selected_exit_metric, get_current_selected_exit_tracking, reset_exit_blacklist,
+    set_current_selected_exit, EXIT_MANAGER,
+};
 use crate::rita_loop::CLIENT_LOOP_TIMEOUT;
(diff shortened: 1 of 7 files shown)


rita_client/src/exit_manager/exit_switcher.rs
```diff
@@ -301,50 +292,50 @@ fn set_exit_state(
...
                 // We can use values in the vector.
-                rita_client_exit_ser_ref
-                    .selected_exit
-                    .selected_id_degradation = average_metric.checked_sub(
-                    rita_client_exit_ser_ref
-                        .selected_exit
-                        .selected_id_metric
-                        .expect("No selected Ip metric where there should be one"),
-                );
+                set_current_selected_exit(SelectedExit {
```
jkilpatr, line 314: this looks ripe for a utility function

rita_client/src/exit_manager/mod.rs
```diff
@@ -71,10 +71,46 @@ pub struct ExitBlacklist {
 /// An actor which pays the exit
 #[derive(Default)]
 pub struct ExitManager {
-    // used to determine if we've changed exits
-    pub last_exit: Option<ExitServer>,
     pub nat_setup: bool,
```
jkilpatr, line 74: it would be great to get some doc comments on the fields here.

jkilpatr (changes requested):
This looks great! Just a few requests for extra comments/docs 

You should expand on the reasoning for why there's no need to write this data to the disk in the commit message. We think that the selected exit will either be correct out the gate (clear metric advantage) or change often at which point having a saved one will be no advantage and cause excessive disk wear and tear as we write the changed config data.

jkilpatr (approved):

PR #640, opened 2022-04-18
Fix bug Exit manager bug with beta19rc5

ptulugu:
After removing selected Exit and mesh ip from the config and storing necessary data in a lazy static, only this ip
was being called when requesting exit information, which led to all exits getting only test exit information

### DIFF
--- a/rita_client/src/dashboard/exits.rs
+++ b/rita_client/src/dashboard/exits.rs
@@ -1,6 +1,6 @@
 //! The Exit info endpoint gathers infromation about exit status and presents it to the dashbaord.
 
-use crate::exit_manager::{exit_setup_request, get_current_selected_exit};
+use crate::exit_manager::{exit_setup_request, get_selected_exit};
 use crate::RitaClientError;
 use actix_web_async::http::StatusCode;
 use actix_web_async::{web::Json, web::Path, HttpRequest, HttpResponse};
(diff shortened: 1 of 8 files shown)


ptulugu:
Since we remove mesh_ip and use subnet.ip() instead, this relies on the fact that the subnets in the config are in the form of
<mesh_ip>/<prefix>, so for example: "fd00"::1337/120" instead of "fd00::1300/120"

rita_client/src/dashboard/exits.rs
```diff
@@ -65,20 +65,22 @@ pub fn dashboard_get_exit_info() -> Result<Vec<ExitInfo>, RitaClientError> {
 
                     for exit in exit_client.exits.clone().into_iter() {
                         let selected = is_selected(&exit.1, current_exit);
-                        let have_route = if let Some(selected_ip) = get_current_selected_exit() {
-                            do_we_have_route(&selected_ip, &route_table_sample)?
+                        let current_selected_ip = get_current_selected_exit();
+                        let route_ip = if let Some(c_ip) = current_selected_ip {
+                            if exit.1.subnet.contains(current_selected_ip.unwrap()) {
```
jkilpatr, line 70: can we add some sort of utility function for getting a selected exit it seems we're using it pretty often.

jkilpatr (changes requested):
Three major things

1) this logic seems repeated enough to put into a function, especially so that future users can just call the function and not wonder exactly how they should implement this
2) that function should have comments explaining what's going on and why
3) ideally we'll have a pretty comprehensive unit test for that function. One that loads a few fake exits and tries to at least get the right IP. We have tests in bable montior with simulated routing tables and stuff but I don't think we need to go that deep.

rita_client/src/dashboard/neighbors.rs
```diff
@@ -106,42 +110,42 @@ fn generate_neighbors_list(
...
-                        neigh_route.metric,
-                        identity.mesh_ip.to_string(),
-                        *identity,
-                        neigh.speed_limit,
-                    ));
-                    continue;
-                }
-                // we check that this is safe above
-                let exit_route = maybe_exit_route.unwrap();
+        let exit_ip = current_exit.map(|e| get_default_exit_ip(e.clone()));
```
jkilpatr, line 113: why map versus just passing the current exit to get default exit ip?

rita_client/src/exit_manager/mod.rs
```diff
@@ -562,6 +555,26 @@ fn correct_default_route(input: Option<DefaultRoute>) -> bool {
     }
 }
 
+/// This function takes an exit server and chooses the exit ip to return
+/// When the exit server provided is the one we are currently connected to, there is no guarantee
+/// that the exit ip in use is the one we are connected to, so we return get_current_selected_exit()
+/// When the exit server is not the one we are connected to, we return subnet.ip(), which is the default
+/// ip address of that exit. This is used in cases where we need to get general information about an exit for
```
jkilpatr, line 562: I think this needs to be split into two functions, one to get the current exit ip and one to get the best ip for a given exit passed to the function.

as it is this function behaves differently depending on the situation and may cause strange behavior if for example we don't have an exit selected.

jkilpatr (changes requested):
This is good, you're right that unit tests probably aren't going to help in this case, but I do think you shoud split this function into two, one that takes no args for getting the current exit ip and one that takes an exit and returns an ip regardless of the current exit.

rita_client/src/dashboard/neighbors.rs
ptulugu: Clippy, if if exitServer is none, return None

jkilpatr (approved):

PR #642, opened 2022-04-26
Halt payments when not updated

kennethAltheaSystems:
The code simply adds to the blockchain oracle struct a instant and
a function that returns true if the blockchain oracle is outdated.
The function will be modified to prevent further potential payment
issues in the future.

### DIFF
--- a/rita_common/src/debt_keeper/mod.rs
+++ b/rita_common/src/debt_keeper/mod.rs
@@ -10,6 +10,7 @@
 //! of the excess complexity you see, managing an incoming payments pool versus a incoming debts pool
 use crate::blockchain_oracle::get_oracle_close_thresh;
 use crate::blockchain_oracle::get_oracle_pay_thresh;
+use crate::blockchain_oracle::potential_payment_issues_detected;
 use crate::payment_controller::queue_payment;
 use crate::payment_validator::PAYMENT_SEND_TIMEOUT;
 use crate::simulated_txfee_manager::add_tx_to_total;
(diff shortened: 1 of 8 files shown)


rita_common/src/debt_keeper/mod.rs
```diff
@@ -239,6 +240,13 @@ pub fn send_debt_update() -> Result<(), RitaCommonError> {
     // (mainly on exits) we batch tunnel change operations before sending them over
     let mut debts_message = Vec::new();
 
+    if potential_payment_issues_detected() {
```
jkilpatr, line 243: right thing to do, but this isn't the right place, this should be in make payment, we want to maintain the right balances in memory for when we need them (eg payments recover)

rita_common/src/tunnel_manager/shaping.rs
```diff
@@ -40,6 +41,11 @@ pub enum ShapingAdjustAction {
 }
 
 pub fn handle_shaping() {
+    if potential_payment_issues_detected() {
```
jkilpatr, line 44: this is the shaper, so it's not actually enforcing on the node, that's in mod.rs I think? (of tunnel manager) this is just adjusting the latency shaper up and down to keep latency smooth.

rita_exit/src/network_endpoints/mod.rs
```diff
@@ -242,6 +243,11 @@ pub async fn get_client_debt(client: Json<Identity>) -> HttpResponse {
     let neg_one = Int256::from(neg_one);
     let zero: Int256 = 0u8.into();
 
+    if potential_payment_issues_detected() {
+        warn!("Potential payment issue detected");
+        return HttpResponse::Ok().json(zero);
```
jkilpatr, line 248: 👍🏻

jkilpatr (changes requested):
Clean architecture here 👍🏻 

I've left some feedback on where the checks in debt keeper and tunnel manager are, since those aren't quite the right places.

jkilpatr (approved):
tests need to pass then this can be merged

jkilpatr (approved):

PR #648, opened 2022-05-12
Removing installation details on ops checkin

ch-iara:
Installation details are stored in ops tools and are
unneeded on the router side. Checking in to ops will
now set installation details to None if successful


jkilpatr (approved):
Looks good, should only remove data once it has successfully reached ops tools.

PR #649, opened 2022-05-12
Fix Ctrl C handler panicing for rita exit

ptulugu:
Fix build_exit.sh


jkilpatr (changes requested):
These are all good changes but need to be structured as probably 3 different commits

1) fixing the dockerfile and build-exit.sh, modifying the docker file to build for debian
2) fixing the is_storage_small behavior
3) removing the todo!() that causes crashes on the exit

Each is an independent change with an independent reasoning/commit message

jkilpatr (approved):

PR #651, opened 2022-05-16
Pranay exits migration to ipv6

ptulugu:


exit_db/migrations/2022-05-16-200950_ipv6_migration/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE clients
+    ADD COLUMN internet_ipv6 varchar(132) NOT NULL DEFAULT ''
+;
+CREATE TABLE assigned_ips
+(
+    subnet varchar(132) CONSTRAINT secondkey PRIMARY KEY,
+    available_subnets varchar(512) NOT NULL,
+    iterative_index bigint DEFAULT 0 NOT NULL
```
jkilpatr, line 8: I'm probably just not up to speed on postgres but how can we have only one interative index to keep track of assigned ip's across many subnets.

Similar what is in the subnet field if we may have many different available subnets?

rita_exit/src/database/database_tools.rs
```diff
@@ -636,6 +636,50 @@ fn generate_index_from_subnet(exit_sub: IpNetwork, sub: IpNetwork) -> Result<u64
...
+    // 2.) for each ip, select ipv6 field, if empty set an ipv6, else continue
+    let filtered_list = clients.select(mesh_ip);
+    let ip_list = filtered_list.load::<String>(&conn)?;
+
+    for ip in ip_list {
+        let filtered_list = clients.select(internet_ipv6).filter(mesh_ip.eq(&ip));
+        let mut sub = filtered_list.load::<String>(&conn)?;
+
+        if sub.len() > 1 {
+            error!("More than one ipv6 for a given client");
```
jkilpatr, line 666: isnt' this an expected case when a client migrates? But I suppose we won't hit that on an initialization.

jkilpatr (commented):
I've got a few questions about the db structure here but this looks goo to me overall

rita_exit/src/database/database_tools.rs
ptulugu: My understanding was that each client would have only one subnet assigned for each exit cluster, but a different subnet for different exit clusters?

exit_db/migrations/2022-05-16-200950_ipv6_migration/up.sql
jkilpatr: 👍🏻  ok so this has multiple rows that's what I was missing. Thanks.

rita_exit/src/database/database_tools.rs
jkilpatr: If two exits within the same cluster have different ipv6 subnets a clietn may need to be assigned one of both to roam between them.

ptulugu:
Tested Initialization, adding new client, deleting client, and recliaming subnet for each of the cases
1.) Both instances dont have ipv6 subnet
2.) One instance had ipv6 subnet
3.) Both had ipv6 subnet, but the same subnet
4.) Both had ipv6, but different subnets

All cases behaved as expected and the logs seem panic free
To minimize risk on us west prod, we can do (1) and just migrate the database and run new rita exit, but not add a 'subnet' field
in the config. It should do no ipv6 related stuff then

rita_exit/src/database/database_tools.rs
```diff
@@ -285,52 +285,91 @@ pub fn delete_client(client: ExitClient, connection: &PgConnection) -> Result<()
...
+        }
+    }
+
+    delete(statement).execute(connection)?;
+    Ok(())
+}
+
+/// Given a vector of client subnet and exit subnets, reclaim all client subnets into the given exit subnets
+/// This relies on the fact that there are no overlapping subnets
+fn reclaim_all_ip_subnets(
```
jkilpatr, line 306: this could use some examples

rita_exit/src/database/database_tools.rs
```diff
@@ -636,6 +686,129 @@ fn generate_index_from_subnet(exit_sub: IpNetwork, sub: IpNetwork) -> Result<u64
         .expect("Unalbe to parse biguint into u64"))
 }
 
+/// This function run on startup initializes databases and other missing fields from the previous database schema
+/// for ipv6 support. Existing clients in the previous schema will not have an ipv6 addr assigned, so every client is
+/// given one on startup
```
jkilpatr, line 691: 👍🏻

jkilpatr (approved):

PR #652, opened 2022-05-17
Modification to neighbor status to perform better name lookup

kennethAltheaSystems:
Frontend previously relied on code to search for neighbor via
id. However, if we move this logic and modify the backend here,
we can better search over all networks for all possible names
and prevent the default name of wireguard key of hitting more
often since it's less useful.

### DIFF
--- a/rita_common/src/tunnel_manager/neighbor_status.rs
+++ b/rita_common/src/tunnel_manager/neighbor_status.rs
@@ -49,6 +49,7 @@ pub fn update_neighbor_status() {
                 id: *id,
                 shaper_speed: lowest_shaper_speed,
                 enforced,
+                neighbor_name: "".to_string(),
             },
         );
     }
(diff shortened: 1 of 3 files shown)


rita_common/src/tunnel_manager/neighbor_status.rs
```diff
@@ -49,6 +49,7 @@ pub fn update_neighbor_status() {
                 id: *id,
                 shaper_speed: lowest_shaper_speed,
                 enforced,
+                neighbor_name: id.wg_public_key.to_string(),
```
jkilpatr, line 52: Considering how this value is filled here this seems to be a placeholder for the ops side. I don't see a reason to add the field here rather than in operator tools then.

jkilpatr (changes requested):
I haven't looked at the operator tools side of this yet, but I don't think this patch will end up being needed as we can avoid adding cruft here by changing what we return to the frontend over in ops tools to include a similar field.

jkilpatr (changes requested):
Thanks for explaining the reasoning to me. 

Make neighbor status an empty string by default and also change the way it's sent to operator tools to be a Vec instead of an Option<Vec>

PR #654, opened 2022-05-24
Fixed clippy lint: holding rwlock across async/await calls

ptulugu:


rita_common/src/tunnel_manager/contact_peers.rs
```diff
@@ -94,10 +90,15 @@ pub async fn tm_neighbor_inquiry(
...
             }
         };
-        contact_neighbor(peer, our_port, udp_socket, peer.contact_socket).await
+
+        contact_neighbor(peer, our_port, udp_socket, peer.contact_socket)
     }
 }
 
+fn set_contacting_neighbors(set: bool) {
+    PEER_LISTENER.write().unwrap().contacting_neighbors = set;
```
jkilpatr, line 99: this needs a doc comment.

rita_common/src/peer_listener/mod.rs
```diff
@@ -117,8 +119,11 @@ pub fn peerlistener_tick() {
 
     let (a, b) = receive_im_here(&mut writer.interfaces);
     {
-        (*writer).peers = a;
-        (*writer).interface_map = b;
+        // Reset hashmaps only when not contacting peers
+        if !writer.contacting_neighbors {
```
jkilpatr, line 123: I'm not sure this will work how you expect, what if for some reason we're stuck contacting neighbors constantly? Lets try this out on an actual router first.

jkilpatr (changes requested):
This looks good overall, but the peer listener situation is complicated. I'm not convinced right now that it will work in all cases.

It seems right now that sending neighbor inquries could block processing ImHere's and cause them all to be thrown away. Saving them up for later processing rather than tossing them would be better, but we'd need to be sure that in that case the update hashmaps gets run before the next round of neighbor inquires start. Otherwise we may never process ImHere's

rita_common/src/peer_listener/mod.rs
ptulugu: I changed it to blocking the entire peer listener tick. This emulates what we were doing before when we held the lock out the for loop that calls contact_neighbors(), The only time we are stuck contacting neighbors is on a gateway when we run to_socket_addrs(), but as long as it times out, we run the tick

jkilpatr (approved):

PR #655, opened 2022-05-25
Pranay Adding magic number

ptulugu:


ptulugu:
tested it on the test exit, seems to work fine

PR #657, opened 2022-06-02
Client subnet size is now a config field

ptulugu:


jkilpatr (changes requested):
The commit message here needs to be improved. The reason the exit client needs to send a /60 is because client devices accepting an ip assignment via SLACC expect to receive a /64 which means that the router of course must have multiple /64 to assign.

What does confuse me is that there's only 16 /64 in a /60 so that's not many devices for the router to have before it runs out of subnet room. We may have to further decrease this value to say a 56 which would be 256 devices

ptulugu:
Im planning to make this a config configurable field

PR #658, opened 2022-06-06
Set the ipv6 address in router /etc/config/network

ptulugu:
When set under 'lan' in the network config, SLAAC take a subnet and assigns
a /64 address to hosts that connect. Rita exit sends a /60, we assign the router this
address in this commit and slaac does ip assignment

### DIFF
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -1,5 +1,5 @@
 use super::KernelInterface;
-use crate::KernelInterfaceError as Error;
+use crate::{open_tunnel::to_wg_local, KernelInterfaceError as Error};
 use althea_types::WgKey;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
@@ -31,7 +31,11 @@ pub struct ClientExitTunnelConfig {
 }
 
 impl dyn KernelInterface {
-    pub fn set_client_exit_tunnel_config(&self, args: ClientExitTunnelConfig) -> Result<(), Error> {
+    pub fn set_client_exit_tunnel_config(
+        &self,
+        args: ClientExitTunnelConfig,
+        local_mesh: Option<IpAddr>,
+    ) -> Result<(), Error> {
         self.run_command(
             "wg",
             &[
@@ -124,6 +128,26 @@ impl dyn KernelInterface {
                         "wg_exit",
                     ],
                 )?;
+            }
+        }
+
+        // If wg_exit does not have a link local addr, set one up
+        if self.get_link_local_device_ip("wg_exit").is_err() {
+            if let Some(mesh) = local_mesh {
+                if let Err(e) = self.run_command(
+                    "ip",
+                    &[
+                        "address",
+                        "add",
+                        &format!("{}/64", to_wg_local(&mesh)),
+                        "dev",
+                        "wg_exit",
+                    ],
+                ) {
+                    error!("IPV6 ERROR: Unable to set link local for wg_exit: {:?}", e);
+                }
+            } else {
+                error!("IPV6 ERRROR: No mesh ip, unable to set link local for wg_exit");
             }
         }
 
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -200,6 +200,34 @@ impl dyn KernelInterface {
         Ok(())
     }
 
+    pub fn set_ipv6_route_to_tunnel(&self) -> Result<(), Error> {
+
+        // Remove current default route
+        if let Err(e) = self.run_command("ip", &["-6", "route", "del", "default"]) {
+            warn!("Failed to delete default ip6 route {:?}", e);
+        }
+        // Set new default route
+        let output = self.run_command(
+            "ip",
+            &[
+                "-6",
+                "route",
+                "add",
+                "default",
+                "dev",
+                "wg_exit",
+            ],
+        )?;
+        if !output.stderr.is_empty() {
+            error!("IPV6 ERROR: Unable to set ip -6 default route");
+            return Err(Error::RuntimeError(format!(
+                "received error setting ip -6 route: {}",
+                String::from_utf8(output.stderr)?
+            )));
+        }
+        Ok(())
+    }
+
     /// Adds nat rules for lan client, these act within the structure
     /// of the openwrt rules which themselves create a few requirements
     /// (such as saying that zone-lan-forward shoul jump to the accept table)
--- a/althea_kernel_interface/src/exit_client_tunnel.rs
+++ b/althea_kernel_interface/src/exit_client_tunnel.rs
@@ -201,23 +201,13 @@ impl dyn KernelInterface {
     }
 
     pub fn set_ipv6_route_to_tunnel(&self) -> Result<(), Error> {
-
         // Remove current default route
         if let Err(e) = self.run_command("ip", &["-6", "route", "del", "default"]) {
             warn!("Failed to delete default ip6 route {:?}", e);
         }
         // Set new default route
-        let output = self.run_command(
-            "ip",
-            &[
-                "-6",
-                "route",
-                "add",
-                "default",
-                "dev",
-                "wg_exit",
-            ],
-        )?;
+        let output =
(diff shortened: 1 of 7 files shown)


scripts/openwrt_build_ipq40xx.sh
```diff
@@ -5,6 +5,6 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # Parse command line arguments
 source $DIR/build_common.sh
 
-cargo install cross
+# cargo install cross
```
jkilpatr, line 8: this needs to be uncommented

rita_client/src/exit_manager/mod.rs
```diff
@@ -777,4 +779,20 @@ pub async fn exit_manager_tick() {
             }
         }
     }
+
+    trace!("Setting up ipv6 for slaac");
+    let rita_settings = settings::get_rita_client();
```
jkilpatr, line 784: this needs a block comment explaining what it's doing.

integration-tests/integration-test-script/utils.py
```diff
@@ -361,10 +361,10 @@ def fuzzy_traffic_match(numA, numB):
     # signs must not match
     if numA > 0 and numB > 0 or numA < 0 and numB < 0:
         return False
-    if numA > 0:
+    if numA >= 0:
```
jkilpatr, line 364: rebase to pull in this change from the already merged commit

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -232,3 +289,56 @@ impl dyn KernelInterface {
         Ok(())
     }
 }
+
+#[test]
+fn test_parsing_babel_route() {
```
jkilpatr, line 294: we already parse babel routes in babel monitor, why are we duplicating here?

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -176,6 +195,44 @@ impl dyn KernelInterface {
...
+        let stdout = String::from_utf8(output.stdout)?;
+        if !stdout.contains(&ip_string) {
+            error!("IPV6 ERROR: Unable to setup ipv6 default route, no babel route found to exit");
+            return Err(Error::RuntimeError(
+                "No babel route found to exit".to_string(),
+            ));
+        }
+        let split: Vec<&str> = stdout.split(&ip_string).collect();
+        let split: Vec<&str> = split[1].split(' ').collect();
+        let gw = split[2];
```
jkilpatr, line 212: As a method of getting the gateway I'm not quite sure this works, or if it does it's only by accident. There's no reason the exit ip and exit fe80 address on wg_exit have to be the same or similar at all. We should have the exit either tell us this address or have a more reliable way of getting it (not via babel, since there is no babel routing within the wg_exit tunnel)

jkilpatr: the function get_reply_ip in linklocal tools is probably a more solid way of doing this.

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -127,6 +127,25 @@ impl dyn KernelInterface {
             }
         }
 
+        // If wg_exit does not have a link local addr, set one up
+        if self.get_link_device_ip_v6("wg_exit").is_err() {
+            //copy another wg interface and use the same linklocal addr
+            match self.get_link_device_ip_v6("wg0") {
```
jkilpatr, line 133: Look at how we do this in open_tunnel.rs in the open_interface function, you can re-use that method of generating an fe80 address here rather than depending on wg0 existing to copy it.

althea_kernel_interface/src/ip_addr.rs
```diff
@@ -44,6 +45,38 @@ impl dyn KernelInterface {
         }
     }
 
+    /// After receiving an ipv6 addr from the exit, this function adds that ip
+    /// to /etc/network/addr. SLAAC takes this ip and assigns a /64 to hosts that connect
+    /// to the router
+    pub fn setup_ipv6_slaac(&self, router_ipv6_str: IpNetwork) {
```
jkilpatr, line 51: this is very clean, does it disrupt the system when it's run?

althea_kernel_interface/src/link_local_tools.rs
```diff
@@ -75,6 +76,23 @@ impl dyn KernelInterface {
         }
     }
 
+    pub fn get_link_device_ip_v6(&self, dev: &str) -> Result<IpNetwork, KernelInterfaceError> {
```
jkilpatr, line 79: I believe this is a duplicate of get_link_local_device_ip at the top of this file but with an IpNetwork response. Maybe modify that function to return an IpNetwork?

jkilpatr (changes requested):
This is a great start. I've left a few suggestions on de-duplication and better ways to get the ip info you're looking for.

integration-tests/integration-test-script/utils.py
ptulugu: fixed

rita_client/src/exit_manager/mod.rs
ptulugu: fixed

althea_kernel_interface/src/ip_addr.rs
ptulugu: When running /etc/init.d/network reload it seemed like the router would hang for around half a second but would then resolve

althea_kernel_interface/src/exit_client_tunnel.rs
ptulugu: Fixed

althea_kernel_interface/src/link_local_tools.rs
ptulugu: fixed

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -127,6 +131,26 @@ impl dyn KernelInterface {
...
+                        "add",
+                        &format!("{}/64", to_wg_local(&mesh)),
+                        "dev",
+                        "wg_exit",
+                    ],
+                ) {
+                    error!("IPV6 ERROR: Unable to set link local for wg_exit: {:?}", e);
+                }
+            } else {
+                error!("IPV^ ERRROR: No mesh ip, unable to set link local for wg_exit");
```
jkilpatr, line 150: IPV6 intended here

althea_kernel_interface/src/exit_client_tunnel.rs
```diff
@@ -176,6 +200,49 @@ impl dyn KernelInterface {
         Ok(())
     }
 
+    pub fn set_ipv6_route_to_tunnel(&self, exit_ip: IpAddr) -> Result<(), Error> {
+        // For now gateway is set to fe80:: + exit ip, and this is ensured to be the case on the exit also
+        if let IpAddr::V6(ip_addr) = exit_ip {
+            let seg = ip_addr.segments();
+            let gw = IpAddr::V6(Ipv6Addr::new(
+                0xfe80, seg[1], seg[2], seg[3], seg[4], seg[5], seg[6], seg[7],
```
jkilpatr, line 208: this should use the function https://github.com/althea-net/althea_rs/blob/master/althea_kernel_interface/src/open_tunnel.rs#L6

jkilpatr (approved):

integration-tests/integration-test-script/world.py
```diff
@@ -448,7 +448,8 @@ def test_debts_values(self, test_traffic_pairs, time, speed, debts, all_routes,
                         node.id, intended_debts[node][owed], owed.id))
                     continue
                 if not fuzzy_match(debts[node.id][owed.id], intended_debts[node][owed]):
-                    print("{} has a predicted debt of {} for {} but actual debt is {} {:.2%} accurate".format(
+                    if debts[node.id][owed.id] != 0:
+                        print("{} has a predicted debt of {} for {} but actual debt is {} {:.2%} accurate".format(
```
jkilpatr, line 452: if the value is zero we should log in a different way. We don't want to just silence this message

jkilpatr (changes requested):
This looks ready to go. I've left one comment about making sure the test code still prints results properly.

PR #660, opened 2022-06-08
Adds multi port toggling

ch-iara:
Interfaces endpoint now accepts an array so that a user
may change and send multiple port reassignments at once
from the dashboard

goes with https://github.com/althea-net/althea-dash/pull/122

### DIFF
--- a/rita_client/src/dashboard/interfaces.rs
+++ b/rita_client/src/dashboard/interfaces.rs
@@ -11,8 +11,8 @@ use crate::RitaClientError;
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct InterfaceToSet {
-    pub interface: String,
-    pub mode: InterfaceMode,
+    pub interface: Vec<String>,
+    pub mode: Vec<InterfaceMode>,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Copy)]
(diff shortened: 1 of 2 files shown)


rita_client/src/dashboard/interfaces.rs
```diff
@@ -11,8 +11,8 @@ use crate::RitaClientError;
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct InterfaceToSet {
-    pub interface: String,
```
jkilpatr, line 14: Minor nit of no real value. struct name should probably also change to InterfacesToSet

rita_client/src/dashboard/interfaces.rs
```diff
@@ -156,204 +156,238 @@ pub fn ethernet2mode(ifname: &str, setting_name: &str) -> Result<InterfaceMode,
     })
 }
 
-fn set_interface_mode(iface_name: &str, mode: InterfaceMode) -> Result<(), RitaClientError> {
+fn set_interface_mode(
```
jkilpatr, line 159: If it's practical it would be better to have a new function multiset_interfaces() or somthing like that that contains the new validation logic running against the arrays of ports to toggle and then keep set_interface_mode and set_ethernet_mode as functions that operate on a single interface at a time.

This makes the multiple port toggling easier to read and reason about in my opinion.

jkilpatr (changes requested):
Really impressive!

My one serious piece of feedback is to try and factor out the loops from the interface mode setting functions. As it stands the complexity of those functions has been increased. They where already dense, the better solution is to add new code which checks over the vector of interfaces and then keep the existing toggle code on singular interfaces, so that it will be no more complex than when we started.

jkilpatr:
oh yeah one other note, update this doc as well https://github.com/althea-net/althea_rs/blob/master/docs/api/router-dashboard.md

rita_client/src/dashboard/interfaces.rs
ch-iara: fixed

jkilpatr (approved):

jkilpatr:
This is ready, waiting on additional work on the dash side.

jkilpatr (approved):

PR #661, opened 2022-06-09
Add exit time endpoint and update client router local time

thosmos:
Adds the /time endpoint to the exit
Adds a check during the client's exit_manager tick to get it and, if its value is more than an acceptable difference from the local time, then update the local time


thosmos:
This needs a bit more work to check if there's an existing tunnel.  As currently implemented, it will get the exit time on every tick.  The call currently happens in the exit_manager tick handler, but it's unclear to me how to get the status of existing tunnels there.

rita_exit/src/database/mod.rs
```diff
@@ -92,15 +93,6 @@ pub fn get_exit_info() -> ExitDetails {
...
 
-// lossy conversion, but it won't matter until 2.9 * 10^8 millenia from now
-pub fn secs_since_unix_epoch() -> i64 {
-    let start = SystemTime::now();
-    let since_the_epoch = start
-        .duration_since(UNIX_EPOCH)
-        .expect("Time went backwards");
-    since_the_epoch.as_secs() as i64
-}
-
```
thosmos, line 103: moved to rita_common

althea_types/src/interop.rs
```diff
@@ -736,3 +736,9 @@ pub struct HeartbeatMessage {
     /// The router version stored in semver format as found in the Cargo.toml
     pub version: String,
 }
+
+/// An exit's unix time stamp that can be queried by a downstream router
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ExitTimeStamp {
+    pub secs_unix_epoch: i64,
```
jkilpatr, line 743: Why not use a SystemTime Struct?  https://doc.rust-lang.org/std/time/struct.SystemTime.html

It's mostly just a u64 with associated functions.

rita_client/src/exit_manager/mod.rs
```diff
@@ -590,11 +591,62 @@ fn initialize_selected_exit_list(exit: String, server: ExitServer) {
     });
 }
 
+/// Retrieve a unix timestamp from the exit's mesh IPv6
```
jkilpatr, line 594: this comment should explain that this is used to sync the clients clock since routers have no internal clock.

rita_exit/src/database/mod.rs
jkilpatr: this function should probably be replaced with usage of SystemTime now since it's more standard. Didn't know about it when I made this func.

rita_client/src/exit_manager/mod.rs
```diff
@@ -590,11 +591,62 @@ fn initialize_selected_exit_list(exit: String, server: ExitServer) {
...
+            error!("Failed to get exit time stamp {:?}", e);
+            return None;
+        }
+    };
+
+    Some(exit_time.secs_unix_epoch)
+}
+
+/// Set the router's time using "date -s \"@12345678\""
+pub async fn set_local_time(unix_time: i64) {
```
jkilpatr, line 626: this should be in a time file in althea kernel interface

rita_client/src/exit_manager/mod.rs
```diff
@@ -590,11 +591,62 @@ fn initialize_selected_exit_list(exit: String, server: ExitServer) {
...
+    let t = format!("\"@{}\"", unix_time);
+    match KI.run_command("date", &["-s", t.as_str()]) {
+        Ok(_) => {}
+        Err(e) => error!("Failed to set local time {:?}", e),
+    }
+}
+
+/// Get time from the exit; if it's more than 60 secs different than local time, then set the local time to it  
+pub async fn maybe_set_local_to_exit_time(exit: ExitServer) {
```
jkilpatr, line 635: we should log when we set the time and maybe have the difference value be a constant? That way it can be more easily adjusted.

jkilpatr (changes requested):
This looks good to me overall. My biggest recommendation is to use SystemTime since it's more robust and has a whole host of nice accessory functions.

I agree it would be nice to stop making these requests once the router is successfully online, to do that we can look at the handshake time on the wg_exit tunnel here

https://github.com/althea-net/althea_rs/blob/master/althea_kernel_interface/src/setup_wg_if.rs#L101

If the handshake time isn't within the last 10 minutes, we should probably be running this time sync function. We may need to better handle the case where the wg tunnel does not exist yet, don't really know how well this function handles that.

rita_client/src/exit_manager/mod.rs
thosmos: According to the docs on SystemTime, it looks like your function uses the standard way to get unix time from it, so I think it's a convenient utility fn.  I actually added a second fn for converting from a SystemTime to secs using the same approach.

rita_client/src/exit_manager/mod.rs
thosmos: made a constant for it at the top of the file.  is there a better place?  Will add an info log statement as well.

althea_types/src/interop.rs
thosmos: will do

rita_exit/src/database/mod.rs
thosmos: According to the docs on SystemTime, it looks like your function uses the standard way to get unix time from it, so I think it's a convenient utility fn. I actually added a second fn for converting from a SystemTime to secs using the same approach.

rita_client/src/exit_manager/mod.rs
thosmos: done

rita_client/src/exit_manager/mod.rs
```diff
@@ -590,11 +598,92 @@ fn initialize_selected_exit_list(exit: String, server: ExitServer) {
...
 }
 
+/// Retrieve a unix timestamp from the exit's mesh IPv6
+pub async fn get_exit_time(exit: ExitServer) -> Option<SystemTime> {
+    info!("Getting the exit time");
+    let exit_ip = exit.subnet.ip();
+    let exit_port = exit.registration_port;
+    let url = format!("http://[{}]:{}/time", exit_ip, exit_port);
+
+    let client = awc::Client::default();
```
jkilpatr, line 608: default timeout here is 5 seconds, which I think is reasonable.

jkilpatr (approved):
This looks good, merge it once your done testing and I'll kick off the build.

jkilpatr:
this has been merged

PR #662, opened 2022-06-13
Exit side ipv6

ptulugu:


althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -89,6 +103,45 @@ impl dyn KernelInterface {
...
+            &[
+                "-A", "FORWARD", "-i", "wg_exit", "-o", "eth0", "-j", "ACCEPT",
+            ],
+        );
+        let res2 = self.run_command(
+            "ip6tables",
+            &[
+                "-A", "FORWARD", "-i", "eth0", "-o", "wg_exit", "-j", "ACCEPT",
+            ],
+        );
```
ptulugu, line 136: Will this always be the eth0 interface?

althea_kernel_interface/Cargo.toml
```diff
@@ -14,6 +14,7 @@ log = "0.4"
 serde_derive = "1.0"
 serde = "1.0"
 althea_types = { path = "../althea_types" }
+ipnetwork = "0.18"
```
jkilpatr, line 17: this is the latest version of ipnetwork, please update it across all the Cargo.toml in the repo (nearly all of our crates depend on it)

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -33,12 +34,19 @@ impl dyn KernelInterface {
         let mut client_pubkeys = HashSet::new();
 
         for c in clients.iter() {
+            let i_ipv6 = &c.internet_ipv6_list;
+            let mut allowed_ips = c.internal_ip.to_string().to_owned();
```
jkilpatr, line 38: a short comment about how this includes the internal ipv4 and then builds the ipv6 list would be good. I really like the implementation here, it's clean.

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -89,6 +102,64 @@ impl dyn KernelInterface {
             ],
         )?;
 
+        // Set up link local mesh ip in wg_exit as fe80 + rest of mesh ip of exit
+        if let IpAddr::V6(ip6) = exit_mesh {
```
jkilpatr, line 106: the first, second, and third segments must be zero, please use this functoin here rather than re-implementing it

https://github.com/althea-net/althea_rs/blob/master/althea_kernel_interface/src/open_tunnel.rs#L6

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -89,6 +102,64 @@ impl dyn KernelInterface {
...
+                ],
+            )?;
+        }
+
+        // Add iptable routes between wg_exit and eth0
+        let output = self
+            .run_command(
+                "ip6tables",
+                &[
+                    "-C", "FORWARD", "-i", "wg_exit", "-o", "eth0", "-j", "ACCEPT",
```
jkilpatr, line 128: this should use the external_nic parameter in rita_common network settings. See the ipv4 nat rules below

althea_kernel_interface/Cargo.toml
ptulugu: It already seems to be 0.18 everywhere?

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: fixed

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: fixed

scripts/openwrt_upload.sh
```diff
@@ -6,6 +6,6 @@ export ROUTER_IP=192.168.10.1
 bash scripts/openwrt_build_$TARGET.sh --features rita_bin/development
 set +e
 ssh root@$ROUTER_IP killall -9 rita
-set -e
+set -e
```
jkilpatr, line 9: extra change, remove the trailing space and amend to drop this file from the commit.

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -89,6 +129,88 @@ impl dyn KernelInterface {
...
+                ],
+            );
+            if res.is_err() {
+                error!("IPV6 ERROR: uanble to set ip6table rules");
+            }
+        }
+        let output = self
+            .run_command(
+                "ip6tables",
+                &[
```
jkilpatr, line 183: this looks like it's a re-implemetnation of some of the logic provided for iptables in althea-kernel-interface. Why not duplicate that function s add_ip6tables_rule and avoid these big if-else blocks

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -76,8 +90,34 @@ impl dyn KernelInterface {
...
+        }
+
+        let ipv6_list: Vec<&str> = client_ipv6_list.split(',').collect();
+
+        for ip in ipv6_list {
+            // Verfiy its a valid subnet
+            if let Ok(ip_net) = ip.parse::<IpNetwork>() {
+                let _res = self.run_command(
+                    "ip",
+                    &["-6", "route", "add", &ip_net.to_string(), "dev", "wg_exit"],
```
jkilpatr, line 107: the verification might be better done at the level above this function where we can attribute the error to a given client. Also we should for sure print an error if we encounter this situation.

Right now if a client had a subnet error in the db they would just mysteriously not work. Printing an error here would help, but just say 'hey this subnet is broken'. But if we print an error in the user setup function then we'll be able to say what router with what subnet is having the issue.

jkilpatr (changes requested):
This looks great! I've left some very minor final nits.

scripts/openwrt_upload.sh
ptulugu: fixed

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: fixed

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: fixed

PR #665, opened 2022-06-24
Routine to log babel/ truncate logs

ptulugu:


althea_kernel_interface/src/babel.rs
```diff
@@ -0,0 +1,7 @@
+use crate::KernelInterface;
+
+impl dyn KernelInterface {
+    pub fn setup_babel(&self) {
```
jkilpatr, line 4: please call this restart babel, more obvious, since there's a lot of babel setup rita actually does later we don't want to cause confusion.

rita_bin/src/client.rs
```diff
@@ -60,6 +60,9 @@ fn main() {
     })
     .expect("Error setting Ctrl-C handler");
 
+    //Restart babeld
+    KI.setup_babel();
```
jkilpatr, line 64: this needs a doc comment

/// Because Rita clears and sets up new Wireguard Tunnels on every restart Babel, which was attached and listening to /// the old tunnels is now in an incorrect state. We must either restart babel or empty it's interfaces list so that the newly /// created wireguard tunnels can be re-added by this instance of Rita. Due to errors in babel (see git history there) 
/// restarting is the way to go as removing dead interfaces often does not work

jkilpatr (approved):

althea_kernel_interface/src/babel.rs
ptulugu: fixed

rita_bin/src/client.rs
ptulugu: fixed

jkilpatr (approved):

PR #666, opened 2022-06-26
Fixes for usage tracker

jkilpatr:
This patch includes a large number of fixes for mistakes in the usage
tracker module. A series of implementation errors and oversights have
built up in this module during it's recent changes.

Large Errors:

1) Premature removal of json conversion code

Due to an oversight when this file was modified to reduce the frequency
of saving to the disk json conversion was removed as part of the beta
17->18 transition. A clear comment had been placed at the top of the
function when this transition was put in place outlining that it was not
to be removed until beta 20, after the 18->19 transition had occured.

This was not noticed by the author or reviewer. Resutling in most
production users usage tracker history being discarded during the 18->19 transition.

2) convert_legacy_usage_tracker test did not actually test conversion

Because this test used an empty dummy struct if the json file was not
actually present or failed to deserialize it would succeed anyways as it
was comparing to the default value subsituted in this case. This test
would locate any panics or control flow issues but was incomplete from
the time it was written.

3) Accidental editing of the usage tracker struct

The function check_usage_hour should never have taken a mutable
reference. The intent of the function is clearly read only, to check how
much data has been used. The algorithm used to do this involves
modifying the struct becuase it was passed mutably it was dropping
actual user data and resulting in the user losing some of their latest
usage data every time it was run.

This usage of a mutable reference probably came from a compiler
suggestion and was overlooked by the author and reviewer

Changes:

This patch corrects these oversights and introduces a new test suite for
usage tracker which generates enormous test structs containing up to
several hundred megabytes of fake usage data. This should allow for more
comprehensive testing in the future and prevent data-corruption bugs
from slipping through so easily.

Changes unrelated to these errors include setting the trigger level for
regular tx saving to only 5 transactions for routers with large fash
rather than 75.

Transaction history saving on shutdown has been simplified, resulting in
a save on shutdown if the write conditions have been met. Rather than
creating an odd exception where routers with small storage would not
save on graceful restart despite being prepared to save a few minutes
later under normal circumstances.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2707,6 +2707,7 @@ dependencies = [
  "clarity",
  "compressed_log",
  "docopt",
+ "env_logger",
  "flate2",
  "futures 0.3.21",
  "hex-literal",
--- a/rita_common/Cargo.toml
+++ b/rita_common/Cargo.toml
@@ -36,11 +36,14 @@ awc = {version = "3.0.0-beta.21", default-features = false, features=["openssl",
 actix-service = "2.0.2"
 web30 = "0.19"
 althea_types = { path = "../althea_types" }
-
 [dependencies.regex]
 version = "1.4"
 default-features = false
 features = ["std"]
+
+[dev-dependencies]
+env_logger = "0.9"
+
 [features]
 # disables cors for dash debugging
 dash_debug = []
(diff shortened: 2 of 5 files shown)


ch-iara (approved):
The new tests with a functional dummy usage tracker are excellent. Might be worth adding that the convert_legacy_usage_tracker test can also be removed in beta 21 migration along with the rest of the legacy conversion but all looks good

PR #670, opened 2022-07-11
Experiments with TC

ptulugu:


ptulugu:
I added an unenforce flag in the config to help with easily change between enforced/unenforced during testing, but leaving it as null should be the same as it not being there. Setting it to true deletes all flows for all clients

jkilpatr:
does not seem to improve anything, good work on making the re-structure feature complete but if there are any capacity restrictions they seem to be elsewhere.

PR #675, opened 2022-07-21
Remove the Config crate from rita

ptulugu:
It looks like all this crate is used for is to load a toml file


ptulugu:
Once merged I can pull this into rita internal and remove it there too

jkilpatr (approved):

ptulugu:
Tested on router, seem to work fine

PR #676, opened 2022-07-26
Unique Exit Identities

ptulugu:
This is the first step to give exits instances unique identities. This allows for older clients to connect
to exits normally with a single ip, while newer clients can connect to a newer interfaces which allows for exit switching

### DIFF
--- a/rita_client/src/exit_manager/exit_switcher.rs
+++ b/rita_client/src/exit_manager/exit_switcher.rs
@@ -17,8 +17,8 @@ use crate::exit_manager::{
 };
 use crate::rita_loop::CLIENT_LOOP_TIMEOUT;
 use crate::RitaClientError;
+use althea_types::ExitList;
 use babel_monitor::{open_babel_stream, parse_routes, Route};
-use ipnetwork::IpNetwork;
 use rita_common::FAST_LOOP_SPEED;
 use settings::client::ExitSwitchingCode;
 use settings::client::SelectedExit;
@@ -193,8 +193,8 @@ impl From<ExitMetrics>
 /// Look at the enum 'ExitSwitchingCode' to see all state and function 'update_metric_value' to see when these are triggered.
 pub fn set_best_exit(
     exit_name: String,
-    exits: IpNetwork,
     routes: Vec<Route>,
+    exit_list: &ExitList,
 ) -> Result<IpAddr, RitaClientError> {
     if routes.is_empty() {
         return Err(RitaClientError::MiscStringError(
@@ -216,11 +216,11 @@ pub fn set_best_exit(
     // Parse all babel routes and find useful metrics
     let exit_metrics = get_exit_metrics(
         routes,
-        exits,
         current_exit_ip,
         tracking_exit,
         current_exit_ip,
         current_adjusted_metric,
+        exit_list,
         exit_map,
     );
 
@@ -423,11 +423,11 @@ fn set_exit_state(
 /// 7.) Metric of the best exit during this tick
 fn get_exit_metrics(
     routes: Vec<Route>,
-    exits: IpNetwork,
     current_exit_ip: Option<IpAddr>,
     tracking_exit: Option<IpAddr>,
     initial_best_exit: Option<IpAddr>,
     initial_best_metric: u16,
+    exit_list: &ExitList,
     exit_map: &mut HashMap<IpAddr, ExitTracker>,
 ) -> ExitMetrics {
     let mut best_exit = None;
@@ -454,7 +454,7 @@ fn get_exit_metrics(
         // All babel routes are advertised as /128, so we check if each 'single' ip is part of exit subnet
         let ip = route.prefix.ip();
 
-        if check_ip_in_subnet(ip, exits) && !blacklisted.contains(&ip) {
+        if check_if_ip_in_exit_list(ip, exit_list) && !blacklisted.contains(&ip) {
             // Not all exits in subnet are blacklisted, so set bool
             all_exits_blacklisted = false;
 
@@ -557,16 +557,6 @@ fn set_last_added_to_zero(exit_map: &mut HashMap<IpAddr, ExitTracker>) {
     for (_, v) in exit_map.iter_mut() {
         v.last_added_metric = 0;
     }
-}
-
-/// Helper function that checks if the provided ip address is within the subnet specified by the provided IpNetwork
-fn check_ip_in_subnet(ip: IpAddr, exits: IpNetwork) -> bool {
-    //check that both are either V4 or V6
-    if ip.is_ipv4() != exits.is_ipv4() {
-        return false;
-    }
-
-    exits.contains(ip)
 }
 
 /// Updates the current metric value during this tick in the global metric tracker. There can be several situations we can be in since there are
@@ -750,8 +740,22 @@ pub fn get_babel_routes(babel_port: u16) -> Result<Vec<Route>, RitaClientError> 
     Ok(routes)
 }
 
+fn check_if_ip_in_exit_list(ip: IpAddr, exit_list: &ExitList) -> bool {
+    for id in exit_list.exit_list.clone() {
+        if ip == id.mesh_ip {
+            return true;
+        }
+    }
+
+    false
+}
+
 #[cfg(test)]
 mod tests {
+
+    use althea_types::{FromStr, Identity, WgKey};
+    use clarity::Address;
+    use ipnetwork::IpNetwork;
 
     use super::*;
     use crate::exit_manager::{reset_blacklist_warnings, ExitBlacklist, MAX_BLACKLIST_STRIKES};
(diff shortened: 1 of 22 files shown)


ptulugu:
Still need to add support for multiple eth keys

rita_client/src/exit_manager/exit_switcher.rs
```diff
@@ -454,7 +454,7 @@ fn get_exit_metrics(
         // All babel routes are advertised as /128, so we check if each 'single' ip is part of exit subnet
         let ip = route.prefix.ip();
 
-        if check_ip_in_subnet(ip, exits) && !blacklisted.contains(&ip) {
+        if check_if_ip_in_exit_list(ip, exit_list) && !blacklisted.contains(&ip) {
```
jkilpatr, line 457: do we still need the concept of blacklisting? I remember we used it mostly for rouge exits in the ip range, but it may also have use for just exits in the cluster that are not acting properly, which we do want this code to be robust against.

jkilpatr: Also can't we avoid iterating over routes and instead just iterate over the list of exits and lookup their router objects perhaps via a hashmap? This would be far more efficient, not that it probably matters in this case bu tit's good to optimize where possible.

rita_client/src/exit_manager/mod.rs
```diff
@@ -81,6 +81,21 @@ pub struct ExitManager {
...
     /// information on what these mean
     pub selected_exit_list: HashMap<String, SelectedExit>,
+    /// Every tick we query an exit endpoint to get a list of exits in that cluster. We use this list for exit switching
+    pub exit_list: ExitList,
+}
+
+/// This functions sets the exit list ONLY IF the list arguments provived is not empty. This is need for the following edge case:
+/// When an exit goes down, the endpoint wont repsond, so we have no exits to switch to. By setting only when we have a length > 1
+/// we assure that we switch when an exit goes down
+pub fn set_exit_list(list: ExitList) {
```
jkilpatr, line 91: very good comment 👍🏻

rita_client/src/exit_manager/time_sync.rs
```diff
@@ -13,7 +13,7 @@ const MAX_EXIT_TUNNEL_HANDSHAKE: Duration = Duration::from_secs(60 * 3);
 /// Retrieve a unix timestamp from the exit's mesh IPv6
 pub async fn get_exit_time(exit: ExitServer) -> Option<SystemTime> {
     info!("Getting the exit time");
-    let exit_ip = exit.subnet.ip();
+    let exit_ip = exit.root_ip;
```
jkilpatr, line 16: shouldn't this pick the selected ip?

rita_exit/src/database/mod.rs
```diff
@@ -402,6 +437,33 @@ pub fn setup_clients(
...
+        (None, Some(_)) => Ok("wg_exit".into()),
+        (Some(new), Some(old)) => {
+            if new < old {
+                Ok("wg_exit_new".into())
+            } else {
+                Ok("wg_exit".into())
+            }
+        }
+        _ => {
+            error!("WG EXIT SETUP: Wg Exit not setup correctly. Client {}, does not have handshake with any wg exit interface", c.wg_pubkey);
```
jkilpatr, line 459: this could be true for any client that is down but registered. Is this trying to solve issues about where to set routes?

rita_exit/src/database/mod.rs
```diff
@@ -381,8 +380,35 @@ pub fn setup_clients(
...
+    let wg_exit_clients: HashMap<WgKey, SystemTime> = KI
+        .get_last_handshake_time("wg_exit")
+        .expect("There should be a wg_exit interface")
+        .into_iter()
+        .collect();
+
+    for c in clients_list.iter() {
+        let interface =
+            get_client_interface(c, new_wg_exit_clients.clone(), wg_exit_clients.clone())?;
+        KI.setup_client_routes(c.internet_ipv6.clone(), c.mesh_ip.clone(), &interface)
```
jkilpatr, line 409: This is going to have to be re-thought for a number of issues

1) we only go and run this code when we know there's a change to enforcement status or registration list (or maybe I'm not quite right and this is above that level double check me)
2) lots of clients remain disconnected and don't have an active handshake on either tunnel. Now this isn't the end of the world since we don't need a route for a client that's offline. But we'll need to be able to detect them coming online and attach them to the right interface. Including situations where clients upgrade and switch interfaces while the exit is online. Maybe instead write a function that's more imdepotent and just checks that every route goes to the most recent handshaked interface.

rita_exit/src/traffic_watcher/mod.rs
```diff
@@ -244,6 +281,35 @@ pub fn watch(
         }
     }
 
+    for (wg_key, bytes) in new_counters.clone() {
+        let state = (
```
jkilpatr, line 285: there's no need to duplicate all this logic, just add a little routine to merge new counters with counters summing the amounts and then you can avoid duplicating these blocks.

settings/src/client.rs
```diff
@@ -33,8 +32,8 @@ pub fn default_config_path() -> String {
 /// ip's within the provided subnet.
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
 pub struct ExitServer {
-    /// Subnet of exits in cluster
-    pub subnet: IpNetwork,
+    /// Ip of exit we first connect to when connected to this cluster
+    pub root_ip: IpAddr,
```
jkilpatr, line 36: So this is the current shared ip across all exits, the multihomed one? Do we need to depend on that or can we just grab an ip from teh list at random to start with?

jkilpatr (changes requested):
This is looking good logic wise.

It looks like the integration tests need some attention and updates here. 

You've done a good job with the encrypted responses 👍🏻 

I've left some comments around the route logic about detecting which interface a peer is on. That will need to be improved.

settings/src/client.rs
ptulugu: This is the first ip in the cluster that the clients knows about. In order to get a list of ips, the clients needs to know an endpoint address to HTTP to, so it uses this address first. 
This field used to be 'subnet' which was of the form <root_ip>/<sub_len>, but now we just use root_ip

rita_client/src/exit_manager/time_sync.rs
ptulugu: fixed

rita_exit/src/database/mod.rs
ptulugu: This is the case where neither wg_exit nor wg_exit_new have a handshake, we just continue in this case

rita_exit/src/database/mod.rs
ptulugu: I moved this upward where it is called every tick. In the case when neither interface has a handshake, i simply skip setting up routes. I set them up when i finally see a handshake

rita_exit/src/traffic_watcher/mod.rs
ptulugu: fixed

rita_client/src/exit_manager/exit_switcher.rs
ptulugu: fixed

rita_exit/src/database/mod.rs
```diff
@@ -402,6 +451,51 @@ pub fn setup_clients(
...
+fn delete_generic_ipv4_route() {
+    // Delete generic ipv4 routes (Remove hardcoding)
+    KI.run_command("ip", &["route", "del", "172.168.0.0/16", "dev", "wg_exit"])
+        .expect("Fix command");
+    KI.run_command(
+        "ip",
+        &["route", "del", "172.168.0.0/16", "dev", "wg_exit_new"],
+    )
+    .expect("Fix command");
+}
```
ptulugu, line 463: I couldnt find where these routes were being set in the first place. If you point me to those, I can remove this code

ptulugu:
Need to test on test cluster
1.) General Exit switching
2.) ipv6 and v4 routes correctly setup
3.) Traffic flows thorugh wg_exit on old routers

rita_client/src/exit_manager/mod.rs
```diff
@@ -170,6 +187,18 @@ fn linux_setup_exit_tunnel(
...
+/// From a exit list, find the corresponding wg public key of our current connected exit to
+/// set up wg tunnels
+fn get_exit_pubkey(ip: IpAddr, exit_list: &ExitList) -> WgKey {
+    for id in exit_list.exit_list.clone() {
+        if id.mesh_ip == ip {
+            return id.wg_public_key;
+        }
+    }
+
+    panic!("Unable to find a valid wg key for current exit, please check that all exit Identities are properly setup on the exit");
```
jkilpatr, line 199: we should consider adding some sort of integrity check for the ExitList struct so that we can error out right away when we get an invalid exit list or at least print a warning.

jkilpatr (commented):
First off for that clippy error you should just be able to rebase off of master.

👍🏻  this looks good to me, not quite ready for approval yet but it's only missing testing in my opinion.

I'd also like to get your throughts on setting up the routes to one of the two exit tunnels (wg_exit for now) by default? that way for at least one case we won't have to wait to add routes. It should improve the experience of users using one of the two tunnels without damaging the experience for the other side at all (switching takes the same amount of time)

rita_client/src/exit_manager/mod.rs
ptulugu: I added a check, should I remove the panic?

ptulugu:
Requires: 
1.) Testing
2.) Edit dash and manually exit addition docs since ExitServer struct changes

ptulugu:
To setup rita exit config, 
1.) Add vector of exits in cluster as below under exit_network
Screenshot from 2022-09-29 09-55-42

2.) Add wg_exit_v2 listening port under exit_netowrk
Screenshot from 2022-09-29 09-55-56

jkilpatr (approved):

PR #679, opened 2022-08-12
Changed serialization for OperatorUpdateMessage

kennethAltheaSystems:
Changed serialization for OperatorUpdateMessage to serialize as a
string to prevent the bincode serialization error. Also, wrote a unit
test to validate that there was no issue with serde.

### DIFF
--- a/althea_types/src/interop.rs
+++ b/althea_types/src/interop.rs
@@ -6,6 +6,8 @@ use babel_monitor::Route as RouteLegacy;
 use clarity::Address;
 use ipnetwork::IpNetwork;
 use num256::Uint256;
+use serde::de::Error;
+use serde::{Deserialize, Deserializer, Serializer};
 use std::collections::hash_map::DefaultHasher;
 use std::fmt;
 use std::fmt::Display;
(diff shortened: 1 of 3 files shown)


althea_types/src/interop.rs
```diff
@@ -516,10 +518,42 @@ pub struct OperatorUpdateMessage {
...
+        serialize_with = "data_serialize",
+        deserialize_with = "data_deserialize"
+    )]
     pub contact_info: Option<ContactType>,
 }
 
+/// Serializes slice of data as "UNFORMATTED DATA" format required
+/// by Ethereum JSONRPC API.
+///
+/// See more https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding
```
jkilpatr, line 532: these comments need to be updated to say that we're just serializing as a string becuase of bincode and ANY values

althea_types/src/interop.rs
```diff
@@ -744,3 +778,40 @@ pub struct HeartbeatMessage {
...
+                max_speed: 0,
+                min_speed: 0,
+            },
+            contact_info: Some(ContactType::Email {
+                email: Address::new("something", "1.1.1.1").unwrap(),
+                sequence_number: Some(0),
+            }),
+        };
+        let data = bincode::serialize(&entry).unwrap();
+        let _try_bincode: Result<OperatorUpdateMessage, Box<ErrorKind>> =
```
jkilpatr, line 814: you're not unwrapping this.

jkilpatr (changes requested):

althea_types/src/interop.rs
```diff
@@ -744,3 +773,50 @@ pub struct HeartbeatMessage {
...
+    pub fn data_serialize<S>(value: &Option<ContactType>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let string_value = serde_json::to_string(&value).unwrap_or_default();
+        serializer.serialize_str(&string_value)
+    }
+
+    /// Deserializes a string as a ContactType
+    pub fn data_deserialize<'de, D>(deserializer: D) -> Result<Option<ContactType>, D::Error>
```
jkilpatr, line 787: you should be able to call the other functions? why duplicate? You just need a 'use' statement to import them at the top of the mod block.

jkilpatr (changes requested):
One comment otherwise good to go

jkilpatr (approved):

PR #685, opened 2022-09-02
add and remove authorized keys

jr-dexter:
Extends operator Actions
 - moves access to Dropbear key file to rita_common
 - updates dropbear authorized keys used for access
 - retains Network Operators keys previous added
 -  Network Operator or management keys can be cycled out


rita_client/src/operator_update/mod.rs
```diff
@@ -335,6 +347,36 @@ fn perform_operator_update(
...
+
+    let auth_keys_to_add = auth_keys_to_add;
+    //let auth_keys_file = auth_keys_file.to_string();
+    let key_file = File::open(auth_keys_file.clone()).expect("Unable to read file");
+    let buf_reader = BufReader::new(key_file);
+
+    for key in buf_reader.lines() {
+        existing_keys.push(key.unwrap());
+    }
+    for key in auth_keys_to_add.lines() {
```
jkilpatr, line 363: here we can just remove the .lines() and iterate over individual key strings

rita_client/src/operator_update/mod.rs
```diff
@@ -498,4 +540,17 @@ mod tests {
             panic!("Not a json map!");
         }
     }
+    fn test_update_auth_keys() {
```
jkilpatr, line 543: this does not have a #[test] flag? I'm not sure how it was being run

jkilpatr (changes requested):
Ok this is a decent start.

I've left a few comments as far as structure and I think we def want to have a few more options like

The process of adding ops managed keys should remove existing ops managed keys (to facilitate key rotation without deleting say keys a home user has added or some network operator)

There should also be a keys_to_delete option where we check for specific keys to remove, we'll use this to remove our old keys as they will be re-inserted in the new managed by ops format.

I know these are pretty big logic changes, so they will each need a test case.

Still needs some clippy fixes, be sure to run rustup update to update rust and then you can have clippy auto fix by running cargo clippy --fix --all

althea_types/src/interop.rs
```diff
@@ -455,6 +455,11 @@ pub enum OperatorAction {
     /// Sets the min gas value to the provided value, primarily intended for use on xDai where
     /// the validators set a minimum gas price as a group without warning
     SetMinGas { new_min_gas: Uint256 },
+    /// Modifies the authorized keys used for access to routers
```
jkilpatr, line 458: this should note that keys in the add list are added to the authorized keys if they are not found there already and that the drop list keys are removed if found.

rita_client/src/operator_update/mod.rs
```diff
@@ -335,6 +350,60 @@ fn perform_operator_update(
...
+    let key_file = File::open(keys_file);
+    match key_file {
+        Ok(key_file_open) => {
+            let buf_reader = BufReader::new(key_file_open);
+
+            for key in buf_reader.lines() {
+                print!("leaving existing key");
+                existing_keys.push(key.unwrap());
+            }
+            if !add_list.is_empty() {
```
jkilpatr, line 376: this if statement isn't needed the for loop will simply do nothing if the list is empty.

rita_client/src/operator_update/mod.rs
```diff
@@ -335,6 +350,60 @@ fn perform_operator_update(
...
+                for key in add_list {
+                    if !existing_keys.contains(&key.to_string()) {
+                        keys_to_add.push(key);
+                    }
+                }
+            }
+            if !drop_list.is_empty() {
+                for (pos, key) in drop_list.iter().enumerate() {
+                    if existing_keys.contains(&key.to_string()) {
```
jkilpatr, line 385: I'm not sure contains will work the way you want it to, lets say we try to add key 123abc in the file it will be 123abc // managed by OpsTools 

which won't  match 'contains' (becuase contains on a vec requires equality whereas contains on a string is a substring match). It may help to iterate over and do contains on each string within the vec using a utility function. These files are small so I wouldn't worry about iterating over them several times during this process.

jkilpatr (changes requested):
I've left a few comments here

jkilpatr (approved):

PR #686, opened 2022-09-02
Restructure Peer listener by cloning a UDP socket to send a hello in …

ptulugu:
…tunnel manager

This change removes the need for calling the peer listener lock in Tunnel Manager. If this
was where a panic/deadlock took place, this change fixes it

### DIFF
--- a/rita_common/src/peer_listener/mod.rs
+++ b/rita_common/src/peer_listener/mod.rs
@@ -26,7 +26,6 @@ lazy_static! {
 
 #[derive(Debug)]
 pub struct PeerListener {
-    pub contacting_neighbors: bool,
     pub interfaces: HashMap<String, ListenInterface>,
     pub peers: HashMap<IpAddr, Peer>,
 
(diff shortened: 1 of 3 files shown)


ptulugu:
Now access to peer listener lock is request only in two location, in peerlistener_tick() and in unlisten_interface()

rita_common/src/peer_listener/mod.rs
```diff
@@ -83,14 +82,57 @@ impl Default for PeerListener {
...
         Ok(PeerListener {
-            contacting_neighbors: false,
             interfaces: HashMap::new(),
             peers: HashMap::new(),
             interface_map: HashMap::new(),
         })
     }
 }
 
+impl Clone for PeerListener {
```
jkilpatr, line 92: Do we have to clone all of peer listener or could we get away with grabbing a copy of just the interfaces for sending new hellos?

rita_common/src/peer_listener/mod.rs
```diff
@@ -103,36 +145,38 @@ fn listen_to_available_ifaces(peer_listener: &mut PeerListener) {
...
     }
 }
 
 /// Ticks the peer listener module sending ImHere messages and receiving Hello messages from all
 /// peers over UDP
-pub fn peerlistener_tick() {
+pub fn peerlistener_tick() -> PeerListener {
     trace!("Starting PeerListener tick!");
 
     let mut writer = PEER_LISTENER.write().unwrap();
```
jkilpatr, line 161: we're holding this lock while performing operations, if we end up blocking trying to recieve an I'm here it will be problematic, is there any way we can eliminate that? Maybe by getting a copy of the data we need and then integrating the response with getter/setter functions

jkilpatr (commented):
This is definitely a good simplification of the design, but I think we can take it a little further to avoid holding locks while doing any sort of work. That would make a deadlock logically impossible in this code.

rita_common/src/peer_listener/mod.rs
ptulugu: updated with getter/setters

rita_common/src/peer_listener/mod.rs
```diff
@@ -103,36 +145,46 @@ fn listen_to_available_ifaces(peer_listener: &mut PeerListener) {
...
-        let (a, b) = receive_im_here(&mut writer.interfaces);
-        {
-            // Reset hashmaps only when not contacting peers
-            (*writer).peers = a;
-            (*writer).interface_map = b;
-        }
-        receive_hello(&mut writer);
-
-        listen_to_available_ifaces(&mut writer);
+    let mut pl = get_pl_copy();
```
ptulugu, line 171: Since we are cloning here, we also clone a UdpSocket, whcih performs reads in recieve_im and receive_hello. But im assuming this is fine in this case since the original UdpSocket never does reading/writing and we only do so with cloned versions?

rita_common/src/peer_listener/mod.rs
```diff
@@ -103,36 +145,46 @@ fn listen_to_available_ifaces(peer_listener: &mut PeerListener) {
...
+    let (a, b) = receive_im_here(&mut pl.interfaces);
+    {
+        pl.peers = a;
+        pl.interface_map = b;
     }
+    receive_hello(&mut pl);
+    listen_to_available_ifaces(&mut pl);
+
+    set_pl(pl.clone());
+    pl
```
ptulugu, line 182: Im assuming this is fine or should I return a cloned udpsocket and save the original?

rita_common/src/peer_listener/mod.rs
jkilpatr: this should be fine as they store the same data behind the scenes.

rita_common/src/peer_listener/mod.rs
jkilpatr: 👍🏻

jkilpatr (approved):

PR #688, opened 2022-09-08
Added Conntrack info to hardware details

ptulugu:


althea_kernel_interface/src/hardware_info.rs
```diff
@@ -345,6 +349,70 @@ fn get_wifi_devices() -> Vec<WifiDevice> {
     ret
 }
 
+fn get_conntrack_info() -> Option<ConntrackInfo> {
+    let max_path = "/proc/sys/net/netfilter/nf_conntrack_max";
+    let table_count_path = "/proc/sys/net/netfilter/nf_conntrack_count";
+    let curr_conns_path = "/proc/net/nf_conntrack";
```
ptulugu, line 355: Let me know if my understanding is correct. I writing this based on this article
https://morganwu277.github.io/2018/05/26/Solve-production-issue-of-nf-conntrack-table-full-dropping-packet/

nf_conntrack_max is total number of connections possible
nf_conntrack_count is number of table entries in a present moment, seems like OS changes this continuously
nf_conntrack shows how many connections are present at a given time in the kernel

jkilpatr: this looks correct, they should be constants though.

jkilpatr (changes requested):
Looks good just needs 

1) var names as constants
2) a unit test that parses conntrack on the current machine

althea_kernel_interface/src/hardware_info.rs
```diff
@@ -345,6 +349,70 @@ fn get_wifi_devices() -> Vec<WifiDevice> {
     ret
 }
 
+fn get_conntrack_info() -> Option<ConntrackInfo> {
+    const MAX_PATH: &str = "/proc/sys/net/netfilter/nf_conntrack_max";
+    const TABLE_COUNT_PATH: &str = "/proc/sys/net/netfilter/nf_conntrack_count";
```
jkilpatr, line 354: can you put a doc comment here about what each value represents?

althea_kernel_interface/src/hardware_info.rs
```diff
@@ -345,6 +349,70 @@ fn get_wifi_devices() -> Vec<WifiDevice> {
...
+    const CURR_CONNS_PATH: &str = "/proc/net/nf_conntrack";
+
+    let curr_conn = Command::new("wc")
```
jkilpatr, line 357: instead of using command line commands can we read the file and count the number of lines that way? I actually believe rust has a function for this when given a file handle but maybe I'm wrong.

I'd like to avoid depending on programs like wc being installed. Some like ip are very hard to eliminate but I don't want to depend on things that we don't have to.

jkilpatr (changes requested):
This looks good, two minor nits

althea_kernel_interface/src/hardware_info.rs
ptulugu: Fixed

althea_kernel_interface/src/hardware_info.rs
ptulugu: Fixed

jkilpatr (approved):

PR #693, opened 2022-10-04
Added code that checks that tunnels present in rita are being monitor…

ptulugu:
…ed by babel

Sometimes due to babel restarting under rita, babel loses all interfaces it monitors leading to communication issues. This commits checks every loop for interfaces that are not being monitored and re-monitors them with babel


ptulugu:
Tested by 'service babeld restart', interfaces are added back and I believe rita works normally

jkilpatr (approved):
This looks generally good to me, maybe put it in the slow loop rather than the fast loop (we don't need to run this that often and make it clear in the logs that a rescue operation is happening, this should not happen during normal operations

PR #696, opened 2022-10-13
Added enforcement support for multiple wg interfaces and ipv6

ptulugu:
With two wg interfaces now, qdiscs, classes and filters are duplicated for existing ipv4 tc configuartions Along with this, a new ipv6 filter is being added on both interfaces for all client classes based on their ipv6 subnet allowing for enforcement on ipv6 data also.

### DIFF
--- a/althea_kernel_interface/src/exit_server_tunnel.rs
+++ b/althea_kernel_interface/src/exit_server_tunnel.rs
@@ -1,6 +1,5 @@
-use crate::open_tunnel::to_wg_local;
-
 use super::{KernelInterface, KernelInterfaceError};
+use crate::open_tunnel::to_wg_local;
 use althea_types::WgKey;
 use ipnetwork::IpNetwork;
 use std::collections::HashSet;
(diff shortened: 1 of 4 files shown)


ptulugu:
IPv6 filter command used is
tc filter add dev wg_exit_v2 parent 1: protocol ipv6 u32 match ip6 dst 2600:3c01:e002:f100::/60 flowid 1:8198

based on this forum
https://serverfault.com/questions/459518/linux-tc-traffic-shaping-for-ipv6

ptulugu:
Tested enforcement, seems to work fine

Screenshot from 2022-10-13 11-23-55

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -75,14 +74,30 @@ impl dyn KernelInterface {
...
                 _ => panic!("Could not derive ipv4 addr for client! Corrupt DB!"),
             }
+
+            // Add ipv6 flows
+            let ipv6_list = c.internet_ipv6_list.clone();
+            let ipv6_list: Vec<&str> = ipv6_list.split(',').collect();
+
+            for ip in ipv6_list {
+                // Verfiy its a valid subnet
+                if let Ok(ip_net) = ip.parse::<IpNetwork>() {
```
jkilpatr, line 95: can we do the validation further up? maybe with a utility function to convert the response from the database and log errors? In this case if there are any problems with the data in the db we just silently ignore it. Ideally we can do verification and then biz logic, not mix them.

althea_kernel_interface/src/traffic_control.rs
```diff
@@ -4,9 +4,39 @@
...
 use crate::KernelInterface;
 use crate::KernelInterfaceError as Error;
+use std::collections::HashSet;
 use std::net::Ipv4Addr;
+use std::sync::Arc;
+use std::sync::RwLock;
+
+lazy_static! {
+    pub static ref TC_DATASTORE: Arc<RwLock<TcDatastore>> =
+        Arc::new(RwLock::new(TcDatastore::default()));
```
jkilpatr, line 18: I don't like this being in the kernel interface, could we have this stored at the rita-exit level and then passed into the kernel interface functions? Adding state to the kernel interface that's in memory internal in addition to system state seems to really increase complexity to me.

jkilpatr (changes requested):
Great to see this, obviously we need all of this to be working properly for a full ipv6 rollout.

I've left a few change requests as far as ways to simplify the design.

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: Fixed

althea_kernel_interface/src/traffic_control.rs
ptulugu: Removed all lazy static call from althea kernel interface

jkilpatr (approved):

PR #702, opened 2022-11-08
Remove uci for setting slaac config

ptulugu:
Prebiously uci was used to setup an ipv6 subnet in slaac, which required a /etc/init.d/network reload. Doing this reload would cause other addresses added to br-lan to get cleared which was problem in lte routers which required a subnet on br-lan. This commit changes this by manually setting up an address on br-lan


althea_kernel_interface/src/interface_tools.rs
```diff
@@ -101,6 +101,41 @@ impl dyn KernelInterface {
...
+            }
+        }
+
+        Ok(ret)
+    }
+
+    /// Gets all the IPv6 addresses from an interface and returns the address and it's netmask
+    /// as a tuple.
+    pub fn get_ipv6_from_iface(&self, name: &str) -> Result<Vec<(IpAddr, u8)>, Error> {
+        let output = self.run_command("ip", &["address", "show", "dev", name])?;
```
jkilpatr, line 132: you should use the ipv6 and ipv4 types instead of ipaddr that way there can't possibly be any confusion about what type of ip we're getting back.

jkilpatr (changes requested):
Just need to change the return types on those functions and get the tests looking good.

althea_kernel_interface/src/interface_tools.rs
ptulugu: fixed

PR #706, opened 2022-11-21
WIP use nftables for NAT forwarding

thosmos:
This is basically an extremely dirty hack at this point.

 [KI.add_lan_forward](https://github.com/althea-net/althea_rs/compare/thosmos/openwrt22?expand=1#diff-1dbb6e7e8e2e0a75f20b488637f10fdfae32ba1b348dbb9b7aa47c9e36297642R29)  uses the nftables command to set up the basic `lan->wg_exit` forward.  That plus some one-time nftables rules in `/etc/rc.local` enables IPv4 to work on reboot:
```
nft add table ip nat
nft add chain nat postrouting '{ type nat hook postrouting priority 100 ; }'
nft add rule nat postrouting oifname wg_exit masquerade
```
I attempted to code some `KI.run_command` calls to run those above rule commands, but they are not working as expected, nor is the check to see if the lan forward already exists.  All of that is in a new `KernelInterface/nftables.rs` file.

This has proven that running some new CLI commands can get things "working", but a big unresolved question is whether the bandwidth counters are getting added and read correctly.  Also, this would all be better implemented as netlink, but that is likely more more work than just modifying the existing CLI commands.


jkilpatr:
This is really great! We'll have to come by when we do the openwrt update to get this all working. Sadly we'll need to work out backwards compatibility with either iptables or nft in order to maintain compatibility with older router versions.

thosmos:
This is now working using the nftables-json opkg and solely with rita's nft commands (nothing in /etc/rc.local).  This still does not handle bandwidth counters which may or may not be working fine.  Now that this is a bit stable, I'll test out a router as my home router and see if I get enforced.

jkilpatr:
A working version of this has been merged into master. Thanks for starting off this work @thosmos  I'm excited to see it go live in the next firmware version.

PR #707, opened 2022-11-21
fix key file permissions

jr-dexter:


rita_client/src/operator_update/mod.rs
```diff
@@ -447,6 +453,18 @@ fn update_authorized_keys(
                 remove_file(&temp_key_file)?
             }
         };
+        match File::open(keys_file) {
+            Ok(file) => {
+                let metadata = file.metadata().unwrap();
```
jkilpatr, line 458: here you're editing the perms of the file after creating and moving it. You could up on line 427 create the file with the file create handle and then set it's permissions there, removing this second operation where you open and then have error handling for opening a file you just created. 

This also has the advantage of being able to return if we have any metadata errors rather than just unwrapping them. You've done the right thing unwrapping them here because there's no way to recover, the program has failed if for some reason we can't edit file perms, but if we do the perms setting up when we create our replacement file we can safely print an error and return having not done any damage to the existing authorized keys.

jkilpatr (changes requested):
A slight change in code structure here can increase reliability.

jkilpatr (changes requested):
This needs to be rebased, also on line 440 we need to return if we can't set permissions.

PR #709, opened 2022-11-22
Delete enforcement class for clients when not enforced

ptulugu:


jkilpatr:
I've merged this with pr #708 to test thanks!

PR #715, opened 2022-12-06
Exits advertises two mesh ip addrs

ptulugu:
In continuation of setting up exit roaming and switching, in order to add backward compatibility between b19 and b20 routers, two wg exit interfaces were added. This commit continues that by getting the exit to advertise two ip addrs, one for b19 routers to connect to as normal, and one for b20 routers to allow for exit roaming B20 routers retrieve a list of IP addrs from an exit and uses these ips to migrate between exits. Since all b19 routers are connected on a seperate ip this should not affect them.

### DIFF
--- a/althea_kernel_interface/src/open_tunnel.rs
+++ b/althea_kernel_interface/src/open_tunnel.rs
@@ -63,6 +63,9 @@ pub struct TunnelOpenArgs<'a> {
     pub private_key_path: &'a Path,
     /// our mesh ipv6 address
     pub own_ip: IpAddr,
+    /// Exit second mesh ip used for exit roaming
+    /// This can be removed after all client migrate to Beta20 and have exit swithcing enabled
+    pub own_ip_v2: Option<IpAddr>,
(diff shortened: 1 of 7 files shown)


althea_kernel_interface/src/open_tunnel.rs
```diff
@@ -63,6 +63,8 @@ pub struct TunnelOpenArgs<'a> {
     pub private_key_path: &'a Path,
     /// our mesh ipv6 address
     pub own_ip: IpAddr,
+    /// our second mesh ip used for exit roaming
```
jkilpatr, line 66: this should not that it's only used on the exits as a transition method and needs to be removed after all routers are updated to beta 20. We'll need a vector on the root ip parameter in the client exit config so that we can expand maintain reliability if one exit is down or unreachable.

jkilpatr (approved):

ptulugu:
Need to test ops side changes for displaying exit info

jkilpatr (approved):

PR #717, opened 2022-12-08
Add a new container for integration tests

ch-iara:


integration_tests_v2/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ENV SPEEDTEST_DURATION=$SPEEDTEST_DURATION
+ENV VERBOSE=$VERBOSE
+ENV NODES=$NODES
+# we pull in the git tar instead of the local folder becuase the raw code is much much smaller
+# note that changes have to be checked in to be pulled in and tested! we pull this in near
+# the bottom to maximize caching of earlier containers
+ADD rita.tar.gz /
+# CMD PATH=$PATH:$HOME/.cargo/bin SPEEDTEST_THROUGHPUT="200" SPEEDTEST_DURATION="15" INITIAL_POLL_INTERVAL=5 BACKOFF_FACTOR="1.5" VERBOSE=1 /althea_rs/integration-tests/rita.sh
+# cargo run --bin tester <- from test_runner folder
+RUN PATH=$PATH:$HOME/.cargo/bin cargo run --manifest-path /althea_rs/test_runner/Cargo.toml --bin tester
```
jkilpatr, line 20: huh manifest path is more elegant than changing directory here. I like it! I think you can eliminate the --bin argument since the crate only produces one bin but that's a very minor nit.

scripts/integration-test-v2.sh
```diff
@@ -0,0 +1,16 @@
...
+set +e
+docker rm -f integration-test
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DOCKERFOLDER=$DIR/../integration_tests_v2/container/
+REPOFOLDER=$DIR/..
+git archive --format=tar.gz -o $DOCKERFOLDER/rita.tar.gz --prefix=althea_rs/ HEAD
+pushd $DOCKERFOLDER
+time docker build -t integration-test --build-arg NODES=$NODES .
```
jkilpatr, line 11: no need for the build arg

integration_tests_v2/container/Dockerfile
jkilpatr: oh this needs to be CMD rather than RUN so that it executes on docker run right now it executes during the container build.

jkilpatr (approved):
I've left a few nits but otherwise this is ready to go in. Please add it to the github CI and we can expand it from here!

PR #718, opened 2022-12-09
Display Client public ipv6 addr on Ops

ptulugu:


althea_types/src/interop.rs
```diff
@@ -595,6 +595,14 @@ pub struct CurExitInfo {
     pub instance_ip: Option<String>,
 }
 
+/// This struct is sent up to op to display info related to a current exit connection
+/// This includes exit deatails, client ipv6 address, etc
+#[derive(Default, Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+pub struct ExitConnection {
```
jkilpatr, line 601: shouldn't the client pub ipv6 be an extension of current exit info? This will be much easier to handle ops side as far as serialization changes (adding a field with an option is free, changing structs is not and will require some transition system). Also it seems that the pub ipv6 is directly related to the cluster and instance

rita_client/src/exit_manager/mod.rs
```diff
@@ -991,3 +992,19 @@ pub async fn exit_manager_tick() {
         }
     }
 }
+
+pub fn get_client_pub_ipv6() -> Option<IpNetwork> {
```
jkilpatr, line 996: 👍🏻

jkilpatr (changes requested):
Looks good, I have some minor struct feedback.

althea_types/src/interop.rs
ptulugu: I added this struct to be more general for sending things in the future, but Exit info and IpAddr are being sent together. The previous struct change for sending exitInfo hasnt been merged and run on ops yet, which is why I changed the format of the struct. Im planning on merging these two commits together

althea_types/src/interop.rs
```diff
@@ -595,6 +595,14 @@ pub struct CurExitInfo {
     pub instance_ip: Option<String>,
```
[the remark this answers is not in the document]
ptulugu: Im in the works of adding more specific types, but in theory this should be impossible right? RitaSettings stores this as an IpAddr, and before we send to Ops, we do .to_string()

jkilpatr (approved):

PR #719, opened 2022-12-16
Babel failure resiliance and faster exit setup

jkilpatr:
Please see individual commits and commit messages.

The general theme of this pr is robustness and speed fixes across the board including some major finds. As it turns out we would panic immediately if babel was not online to accept our connection requests, this could include situations where babel was just slow starting up.

This is a smoking gun as far as recent issues with babel go, it explains a lot of the instability we have seen and a simple couple of line fix would address it. This pr goes much further than that, doing a general refactor of babel monitor into discrete files and handling errors that where previously fatal.

It also optimizes the rita client and common fast loops, reducing the number of duplicate requests and parallelizing via async join in the case of the rita client fast loop this is a dramatic (6x) performance improvement. 

Finally some code is added to restart babel in case it has been killed or manually stopped. This is actually pretty redundant because the openwrt service will restart babel on it's own if it crashes, so it really addresses babel being in strange states like not responding to commands or producing parsable output.


ptulugu (approved):

ch-iara (approved):
max amount of time this goes for is 10 minutes without babel before restarting? looks good

jkilpatr:
> max amount of time this goes for is 10 minutes without babel before restarting? looks good

in my testing experience it was somewhat variable, in theory it should be exactly 11 minutes, ten loops then the next loop it fixes stuff. But these run faster/slower for whatever reason (expected variability)

PR #720, opened 2022-12-19
Add namespaces to new integration test container

ch-iara:


test_runner/src/main.rs
```diff
@@ -1,3 +1,155 @@
+use core::time;
+use std::thread;
+
+use althea_kernel_interface::{KernelInterfaceError, KI};
+
 fn main() {
-    println!("Hello, world!");
+    /*
+    These are currently connected as such:
+    A---------B
```
jkilpatr, line 9: go ahead and move this namespace setup into a function call something arbitrary like "five_node_config". Over time we'll add different configs for different tests and eventually a function to generate random topologies

test_runner/src/main.rs
```diff
@@ -1,3 +1,155 @@
...
+            (testb.clone(), testc.clone()),
+            (testa, testc.clone()),
+            (testc, testd.clone()),
+            (testb, testd),
+        ],
+    };
+    let res = setup_ns(namespaces);
+    println!("Namespaces setup: {:?}", res);
+    // this sleep is for debugging so that the container can be accessed to poke around in
+    thread::sleep(ten_mins);
```
jkilpatr, line 47: we probably want to remove this for CI, otherwise we're just burning CI minutes.

test_runner/src/main.rs
```diff
@@ -1,3 +1,155 @@
...
+        );
+        println!("{:?}", res);
+    }
+    for link in spaces.linked {
+        let veth_ab = format!("veth-{}-{}", link.0 .0, link.1 .0);
+        let veth_ba = format!("veth-{}-{}", link.1 .0, link.0 .0);
+        let ip_ab = format!("192.168.{}.{}/24", link.0 .1, counter);
+        let ip_ba = format!("192.168.{}.{}/24", link.1 .1, counter);
+        let subnet_a = format!("192.168.{}.0/24", link.0 .1);
+        let subnet_b = format!("192.168.{}.0/24", link.1 .1);
```
jkilpatr, line 74: embedding what ip's to assign into the original struct is a potential future improvement, for now this is good.

test_runner/src/main.rs
```diff
@@ -1,3 +1,155 @@
...
+
+/// This struct holds the setup instructions for namespaces
+//#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct NamespaceInfo {
+    /// the names of the namespaces
+    pub names: Vec<String>,
+    /// linked nodes written as tuple pairs
+    /// The string is for the namespace name(NOTE: names must be <=4 characters as interfaces cannot be more than 15 char, and we input as veth-{}-{})
+    /// and the u32 is for the subnet on the 3rd octet
+    pub linked: Vec<((String, u32), (String, u32))>,
```
jkilpatr, line 154: The only thing I can think of to add here right now is a function to validate the topology. For example by checking that each link has the appropriate two ends and no names are too long.

jkilpatr (approved):
I've left several nits but no concrete changes. Great to see this ready!

jkilpatr (approved):
This is awesome! Left a few comments as to cleanups but otherwise ready to go

scripts/settings.toml
```diff
@@ -0,0 +1,34 @@
+[payment]
+pay_threshold = "10000000000000"
+close_threshold = "-10000000000000"
+close_fraction = "100"
+buffer_period = 3
+eth_address = "0x0101010101010101010101010101010101010101"
```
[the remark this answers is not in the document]
ch-iara: Whoops, yes this is a duplicate of settings/test.toml... We can't use the rita_test.toml from the old test because some of the fields in that one (namely the wg priv key) prevent this version of spinning up rita from running properly

jkilpatr (approved):

jkilpatr (approved):

PR #722, opened 2022-12-20
Pay thresh is a const

ptulugu:


rita_common/src/blockchain_oracle/mod.rs
```diff
@@ -93,65 +96,33 @@ pub struct BlockchainOracle {
...
-
 // Set Xdai default
 fn default_net_version() -> u64 {
     100u64
 }
 
+pub fn calculate_close_thresh() -> Int256 {
+    let pay_thresh = get_pay_thresh();
+
+    let neg_one = -1i32;
```
jkilpatr, line 107: this needs a doc comment with as much context as possible.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -337,10 +336,15 @@ fn deserialize_from_json(file_path: String) -> Option<DebtDataSer> {
     }
 }
 
+pub fn get_pay_thresh() -> Int256 {
+    let payment = settings::get_rita_common().payment;
```
jkilpatr, line 340: this function and the function for calculating the close threshold should probably be in the same file, also comment needed here.

rita_common/src/debt_keeper/mod.rs
```diff
@@ -555,6 +559,15 @@ impl DebtKeeper {
     }
 
     /// This updates a neighbor's debt and outputs a DebtAction if one is necessary.
+    /// /// 2.) payment_threshold : This is the ammount at which a router will make a payment. Below this value, the router will not may a payment since
+    ///     a large portion of the payment will be eaten in fees which is not desirable. This is caluculated by
+    ///     gas_price * gas_required * dynamic_fee_multiplier, where dynamic fee multiplier determines how much larger the payment is
+    ///     compared to the fees (fees are essentially gas_price * gas_required). When this is larger, the router pays less often and
+    ///     vice versa
```
jkilpatr, line 566: comment data is good, but should maybe be down on the variables when we declare them? not as a function comment since the function doesn't take these as arguments, also the numbering is odd.

settings/src/payment.rs
```diff
@@ -75,6 +69,16 @@ fn default_min_gas() -> Uint256 {
     2_000_000_000u128.into()
 }
 
+pub fn default_payment_threshold() -> Int256 {
+    // This value is set to 1 eth constant (1e^18) * 0.3
+    // 1 eth constant is 1 dollar, so this is 30 cents
+    300_000_000_000_000_000i64.into()
```
jkilpatr, line 75: 👍🏻

jkilpatr (changes requested):
This looks ok. I've left some requests for doc comments and also I don't see where enforcement enabled is handled in rita_exit?

Also for the enforcement toggle, that should be it's own commit and it needs to actively undo all enforcement currently happening. Which will require changes over in the enforcement code itself rather than debt keeper (which if you make the changes there there's no reason to check the value in debt keeper like you are now)

rita_common/src/debt_keeper/mod.rs
```diff
@@ -602,12 +616,17 @@ impl DebtKeeper {
...
+                    info!(
+                        "debt {} is below close threshold {} for {}. suspending forwarding",
+                        debt_data.debt, close_threshold, ident.wg_public_key
+                    );
+                    debt_data.action = DebtAction::SuspendTunnel;
+                    Ok(DebtAction::SuspendTunnel)
+                } else {
+                    debt_data.action = DebtAction::OpenTunnel;
+                    Ok(DebtAction::OpenTunnel)
+                }
```
ptulugu, line 629: This is where we use enable enforcement. On subsequent ticks, it should return OpenTunnel instead of SuuspendTunnel which would undo any active enforcement?

rita_common/src/debt_keeper/mod.rs
ptulugu: done

rita_common/src/blockchain_oracle/mod.rs
ptulugu: done

rita_common/src/debt_keeper/mod.rs
ptulugu: done

rita_common/src/blockchain_oracle/mod.rs
```diff
@@ -93,39 +95,32 @@ pub struct BlockchainOracle {
...
+// Set Xdai default
+fn default_net_version() -> u64 {
+    100u64
 }
 
-fn default_close_threshold() -> Int256 {
-    Int256::from(-1) * default_pay_threshold() * CLOSE_THRESH_MULT.into()
+/// payment_threshold : This is the ammount at which a router will make a payment. Below this value, the router will not may a payment since
+/// a large portion of the payment will be eaten in fees which is not desirable. This is caluculated by
+/// gas_price * gas_required * dynamic_fee_multiplier, where dynamic fee multiplier determines how much larger the payment is
```
jkilpatr, line 105: this comment needs updating as well it discussing the gas price and such.

rita_common/src/debt_keeper/mod.rs
jkilpatr: yes I believe that will do it.

jkilpatr (changes requested):
One nit and one question left.

Where is the unenforcement for exit clients if enforcement is disabled? The diff in wg_exit doesn't reference the new config value at all.

jkilpatr (approved):

rita_common/src/blockchain_oracle/mod.rs
ptulugu: done

PR #724, opened 2022-12-28
Use exit manager's selected ip to query debts

ptulugu:
This matters when we have a router that have two exits in a cluster in manual peers, we need to ensure we are querying and paying the correct exit in the cluster


ptulugu:
1.) payment controller now stores all payments sent and received since start
2.) When making payment, get all payments made, send list over. validate new payment
3.) recipient compares this to their own list. Send difference to payment validator
4.) Special case for exits, we send this list to all exits in cluster, correct exit will account our payment

PR #725, opened 2023-01-04
Make payment V2

ptulugu:


rita_client/src/traffic_watcher/mod.rs
```diff
@@ -96,14 +96,13 @@ pub async fn query_exit_debts(msg: QueryExitDebts) {
...
     let start = Instant::now();
-    let exit_addr = msg.exit_internal_addr;
     let exit_id = msg.exit_id;
     let exit_port = msg.exit_port;
     // actix client behaves badly if you build a request the default way but don't give it
     // a domain name, so in order to do peer to peer requests we use with_connection and our own
     // socket specification
     let our_id = settings::get_rita_client().get_identity();
-    let request = format!("http://{}:{}/client_debt", exit_addr, exit_port);
+    let request = format!("http://[{}]:{}/client_debt", exit_id.mesh_ip, exit_port);
```
jkilpatr, line 105: I don't think this ends up doing anything. We're paying the exit internal ip (so the one inside wg_exit) so there's no chance of this being the wrong ip. You could make the argument that using the mesh_ip lets us pay before wg_exit is online, but we need internet access to pay, so it doesn't allow us to make payments when we otherwise couldn't and also isn't any better about not paying the wrong exit.

rita_common/src/network_endpoints/mod.rs
```diff
@@ -45,6 +27,26 @@ pub async fn make_payments(item: Json<PaymentTx>) -> HttpResponse {
...
 }
 
+/// The recieve side of the make payments v2 call. This processes a list of payments instead of a single payment
+pub async fn make_payments_v2(item: Json<HashSet<PaymentTx>>) -> HttpResponse {
+    let pmt_list = item.into_inner();
+
+    for pmt in pmt_list {
+        let local_payment_list = get_payment_txids(pmt.from, false);
+
+        if !local_payment_list.contains(&pmt) {
```
jkilpatr, line 37: I don't see a reason to compare to the local payments list, we already check that we don't double validate a payment in the payment validator, so this is a duplicate.

rita_common/src/payment_controller/mod.rs
```diff
@@ -252,32 +271,49 @@ async fn make_payment(mut pmt: PaymentTx) -> Result<(), PaymentControllerError>
...
+                tx_id,
+                val.status(),
+                val.body().await,
+                neighbor_url_v2,
+                full_node,
+                pmt.amount
+            );
+        } else {
+            error!(
+        "We published txid: {:#066x} via url {} but our neighbor responded with status {:?} and body {:?}",
```
jkilpatr, line 290: would you do a match on the tuple with both results and print only one message even if both succeed? We don't want to print extra log messages if possible becuase we parse these later on in graylog

rita_common/src/payment_validator/mod.rs
```diff
@@ -92,6 +97,50 @@ impl Default for PaymentValidator {
...
+        e.insert(set);
+    } else {
+        let set = data
+            .get_mut(&neighbor)
+            .expect("This key should have an initialized set");
+        set.insert(pmt);
+    }
+}
+
+pub fn get_payment_txids(id: Identity, sent_from_us: bool) -> HashSet<PaymentTx> {
```
jkilpatr, line 123: I'd separate this into two function calls, it's no additonal lines actually and it's easier to see what's going on when a specific function is called.

rita_common/src/payment_validator/mod.rs
```diff
@@ -92,6 +97,50 @@ impl Default for PaymentValidator {
...
+    let neighbor;
+    let data = if sent_from_us {
+        neighbor = pmt.to;
+        &mut lock.successful_transactions_sent
+    } else {
+        neighbor = pmt.from;
+        &mut lock.successful_transactions_rec
+    };
+
+    if let std::collections::hash_map::Entry::Vacant(e) = data.entry(neighbor) {
```
jkilpatr, line 111: once again these can be distinct functions, it's net less lines and less chance of human error.

rita_common/src/payment_validator/mod.rs
```diff
@@ -142,14 +182,14 @@ struct Remove {
...
+fn remove(msg: Remove, history: &mut PaymentValidator, sent_from_us: Option<bool>) {
     let was_present = history.unvalidated_transactions.remove(&msg.tx);
     // store successful transactions so that they can't be played back to us, at least
     // during this session
     if msg.success && was_present {
-        history
-            .successful_transactions
-            .insert(msg.tx.payment.clone().txid.unwrap());
+        let sent_from_us =
+            sent_from_us.expect("For a valid transaction, this needs to be some value!");
```
jkilpatr, line 191: I don't understand why this is needed? Why does it matter if we sent it or not?

rita_common/src/payment_validator/mod.rs
```diff
@@ -115,24 +164,15 @@ pub fn calculate_unverified_payments(router: Identity) -> Uint256 {
...
-            // insert is safe to run multiple times just so long as we check successful tx's for duplicates
-            history.unvalidated_transactions.insert(ts);
-        }
-    } else {
-        error!(
-            "Someone tried to insert an unpublished transaction to validate!? {:?}",
-            ts
-        );
-        // in a development env we want to draw attention to this case
-        #[cfg(feature = "development")]
```
jkilpatr, line 130: are you sure this is checked later? I'm not sure.

rita_common/src/payment_validator/mod.rs
```diff
@@ -74,14 +75,18 @@ impl fmt::Display for ToValidate {
 
 pub struct PaymentValidator {
     unvalidated_transactions: HashSet<ToValidate>,
-    successful_transactions: HashSet<Uint256>,
+    /// All successful transactions sent FROM this router, mapped To Address-> list of PaymentTx
+    successful_transactions_sent: HashMap<Identity, HashSet<PaymentTx>>,
+    /// All successful transactions sent TO this router, mapped From Address-> list of PaymentTx
+    successful_transactions_rec: HashMap<Identity, HashSet<PaymentTx>>,
```
jkilpatr, line 81: I think all we need to store here is txid's, what about the rest of the tx do we use?

jkilpatr (changes requested):
This iteration is better and has reduced duplication. I've left several comments.

I'm not a fan of the 'to us' boolean pattern it seems to add complexity to functions that could just be broken up into two functions.

I've also commented on what I think is an error in the validation done in to_validate.

rita_client/src/traffic_watcher/mod.rs
ptulugu: From what I understand all exits in the cluster have the same internal ip 172.168.0.254? So in that case if traffic is routed through a specific 2602 addr, how do query_exit_debts know which exit 172.168.0.254 refers to?

rita_common/src/network_endpoints/mod.rs
ptulugu: If your talking about the chekc in validate_later that used to see 
If pmt in successful.transactions {
dont add to unvalidated.trasactions
}
I removed that check and moved it here since it was a little more compliacted to write that with new new hashmap<id -> hashset> format of datastore

If youre talking about payment validator not adding it to successfull transactions since we use a hashmap at the end, why would we want to send payment validator a bunch of tx that we know we already accounted for?

rita_common/src/payment_validator/mod.rs
ptulugu: I moved the datastore from Hashset<all tx> to HashMap<id -> Hashset<tx>> (one for receiving and sent) since it was much easier to query for a particular client

For example, if a client sends me a list of txs that it made, if i stored all success transactions, I would have to iterate through all txs i have stored and get the ones that are 1.) addrssed to me 2.) sent from client. 

WIth a large exit, i didnt think it would make sense to do that, instead of just using separate hashmaps, and making small changes to how payment validator stored data

rita_common/src/payment_validator/mod.rs
ptulugu: Are you saying txid is none?, Im checking that here in the same function
Or are you saying printing this log in development? If so ill add that

ptulugu: Are you saying txid is none? Looks likes im currently just printing a log, Ill add a fix here where we dont add the tx to the unvalidated set

rita_common/src/payment_validator/mod.rs
ptulugu: If the client sends a list of old pmts over to exit, and the exit sees an old pmt that needs to be accounted for, doesnt the exit need the whole pmt struct to pass it to validate_later?

rita_client/src/traffic_watcher/mod.rs
jkilpatr: It will always be the exit they are currently connected to because that's what wg_exit is pointed at.

rita_common/src/network_endpoints/mod.rs
jkilpatr: Sending payment validator a bunch of tx we know are accounted for is fine so we can de duplicate the process of checking

Payment validator can validate millions of tx. We don't really have to worry about putting more load on it. We do have to worry about code or logic errors and the more places we do different things and the more spread out the validation logic is the more likely it is that we will make a mistake, which can have expensive consequences.

rita_common/src/payment_validator/mod.rs
jkilpatr: If we just query the hashset for the txid it's O(1). So I think we can use the Hashset and have an easier time processing the data.

rita_common/src/payment_controller/mod.rs
ptulugu: Fixed

rita_common/src/payment_validator/mod.rs
ptulugu: Fixed

rita_common/src/payment_validator/mod.rs
ptulugu: fixed

rita_common/src/payment_validator/mod.rs
ptulugu: Fixed

rita_common/src/payment_validator/mod.rs
ptulugu: Fixed

rita_common/src/network_endpoints/mod.rs
```diff
@@ -10,38 +10,45 @@ use actix_web_async::web::Json;
...
+    for pmt in pmt_list {
+        let ts = ToValidate {
+            payment: pmt.clone(),
+            received: Instant::now(),
+            checked: false,
+        };
+
+        // Duplicates will be removed here
+        if let Err(e) = validate_later(ts) {
+            build_err.push_str(&format!("{}\n", e));
```
jkilpatr, line 45: this is nice 👍🏻

rita_common/src/payment_controller/mod.rs
```diff
@@ -252,8 +271,51 @@ async fn make_payment(mut pmt: PaymentTx) -> Result<(), PaymentControllerError>
...
+                }
+                (false, true) => {
+                    error!(
+                        "Make_payment_v2 with txid {:#066x} failed with status {:?} and body {:?}",
+                        tx_id,
+                        val2.status(),
+                        val2.body().await
+                    );
+                    info!(
+                        "Payment pmt with txid: {:#066x} is sent to our neighbor with status {:?} and body {:?} via url {}, using full node {} and amount {:?}",
```
jkilpatr, line 300: can you use regular print not debug print for amount, it will make the numbers unusable to log parsers

jkilpatr (approved):
This looks good, approving with one very minor request (regular print not debug print for amounts in payments)

rita_common/src/payment_controller/mod.rs
ptulugu: Fixed

PR #729, opened 2023-01-11
Rita Extender

ptulugu:


rita_bin/src/extender.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#![warn(clippy::all)]
+#![allow(clippy::pedantic)]
+#![forbid(unsafe_code)]
+
+#[cfg(feature = "jemalloc")]
+use jemallocator::Jemalloc;
+#[cfg(feature = "jemalloc")]
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
```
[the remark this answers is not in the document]
jkilpatr: we really don't there was a time there when jemalloc was often faster than the system allocator and I thought it might matter, but it can be removed from all crates.

rita_bin/src/extender.rs
```diff
@@ -0,0 +1,75 @@
...
+use rita_client::get_device_mac;
+use rita_common::logging::enable_remote_logging;
+use rita_common::utils::env_vars_contains;
+use rita_extender::extender_checkin;
+use rita_extender::get_checkin_message;
+use rita_extender::start_rita_extender_loop;
+use rita_extender::DEFAULT_UPSTREAM_ENDPOINT;
+
+#[actix_rt::main]
+async fn main() {
```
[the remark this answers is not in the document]
jkilpatr: the actix_rt main is good, but you can't use the wg key of the router since the extender won't have any wireguard keys (or anything to do with wireguard actually).

Why do we need to hardcode? we should get the mac address from the system on the fly.

rita_bin/src/extender.rs
```diff
@@ -0,0 +1,75 @@
...
+    // On Linux static builds we need to probe ssl certs path to be able to
+    // do TLS stuff.
+    openssl_probe::init_ssl_cert_env_vars();
```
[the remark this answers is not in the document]
jkilpatr: yes, you will need this to make requests to https sites on the internet

rita_client/src/dashboard/extender_checkin.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use rita_common::dashboard::wifi::{
+    get_device_mac, get_wifi_config_internal, ExtenderAdditionalSettings, ExtenderCheckin,
+    ExtenderUpdate,
+};
+
+lazy_static! {
+    /// This is a mapping of Mac Addr of extender -> Its checkin struct
+    /// Do we need to store all this information here? What this rita client need to keep track of?
+    static ref EXTENDER_LIST: Arc<RwLock<HashMap<String, ExtenderCheckin>>> = Arc::new(RwLock::new(HashMap::new()));
```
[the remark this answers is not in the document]
jkilpatr: we need to have an endpoint that displays recently connected extenders for the main router dashboard.

rita_client/src/dashboard/extender_checkin.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+pub async fn extender_checkin_handler(extender: Json<ExtenderCheckin>) -> HttpResponse {
+    let extender = extender.into_inner();
+    // Store this in local database, what info do we need to store here
+    add_wifi_extender(extender);
```
jkilpatr, line 21: we need to keep track of what extenders we have connected and also their wifi info so we can upload it. (maybe other hardware info too? we mostly care about wifi info though)

rita_common/Cargo.toml
```diff
@@ -36,6 +36,7 @@ awc = {version = "3.0.0-beta.21", default-features = false, features=["openssl",
 actix-service = "2.0.2"
 web30 = "0.22"
 althea_types = { path = "../althea_types" }
+mac_address = "1.1.4"
```
jkilpatr, line 39: no need for an additional dependency https://serverfault.com/questions/357811/find-mac-addresses-in-proc-or-somewhere-despite-bonding-device

We can grab the mac address of the first interface by parsing proc. Easy to implement.

rita_common/src/dashboard/wifi.rs
```diff
@@ -1,18 +1,18 @@
 //! These endpoints are used to modify mundane wireless settings
 
+use crate::dashboard::nickname::maybe_set_nickname;
+use crate::{RitaCommonError, KI};
```
jkilpatr, line 4: why make all these changes? why not import from rita client?

rita_extender/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::sync::RwLock;
+use std::thread;
+use std::time::Duration;
+use std::time::Instant;
+
+use rita_common::dashboard::wifi::ExtenderUpdate;
+
+const EXTENDER_CHECKIN_TIMEOUT: Duration = Duration::from_secs(5);
+const EXTENDER_LOOP_SPEED: Duration = Duration::from_secs(30);
+pub const DEFAULT_UPSTREAM_ENDPOINT: &str = "http://[{}]:{}/extender_checkin";
```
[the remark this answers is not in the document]
jkilpatr: we need to use the routers ip on the lan which would be 192.168.10.1 by default for ipv6 and is harder to predict for ipv6 since it's based on what we're assigned via dhcpv6

rita_common/src/dashboard/wifi.rs
```diff
@@ -722,6 +743,52 @@ fn strip_string_quotes(mut name: String) -> String {
...
+
+/// This is the list of settings rita extender stores internally
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExtenderCheckin {
+    pub device_mac: String,
+    pub wifi_info: Vec<WifiInterface>,
+}
+
+impl Default for ExtenderCheckin {
+    fn default() -> ExtenderCheckin {
```
jkilpatr, line 768: this should be a function with a result, not default. Since default is a popular trait it will get called often even by other traits which expect it not to fail. In this case we need to be able to keep trying if we experience an error.

jkilpatr (changes requested):
This looks like a good start.

I've responded to your comments and left my own. The biggest question I have is why the wifi file was moved to common. I don't really see the need for that and all the associated changes.

rita_bin/src/extender.rs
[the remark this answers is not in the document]
ptulugu: I meant using the wg key of the main router
extender queries endpoint -> main router return its wgkey -> we use it to log so that we know which router an extender belongs to in greylog

rita_common/Cargo.toml
ptulugu: Couldnt find proc/net/bonding, but found in /sys/class/net

rita_common/src/dashboard/wifi.rs
ptulugu: fixed

rita_common/src/dashboard/wifi.rs
ptulugu: fixed

rita_extender/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+fn settings_changed(
+    _our_settings: Vec<WifiInterface>,
+    _router_settings: Vec<WifiInterface>,
+) -> bool {
+    todo!()
+}
+
+fn apply_wifi_settings() {
+    todo!()
+}
```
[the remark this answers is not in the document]
jkilpatr: How this is handled is somewhat arbitrary, we should pick the same ssid for all the 5ghz networks and 2.4ghz networks in my opinion and when different interfaces disagree just take whichever has the lowest index.

As an example if a router has 3 interfaces, two of which are 5ghz we take the ssid of the first 5ghz radio and put it on all the other devices.

rita_extender/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+use rita_client::extender::ExtenderUpdate;
+
+const EXTENDER_CHECKIN_TIMEOUT: Duration = Duration::from_secs(5);
+const EXTENDER_LOOP_SPEED: Duration = Duration::from_secs(30);
+pub const DEFAULT_UPSTREAM_ENDPOINT: &str = "http://192.168.10.1:4877/extender_checkin";
+
+lazy_static! {
+    /// A list of settings used to interally set /etc/config/wireless, and to sent up to main router.
+    /// Stored here in memory to check if settings changed before manually running the commands to modify them
+    pub static ref EXTENDER_SETTINGS: Arc<RwLock<ExtenderCheckin>> = Arc::new(RwLock::new(ExtenderCheckin::default()));
```
ptulugu, line 33: Currently this storage doesnt do anything since I apply the settings I get from the main router every tick, without checking if the settings changed. Wanted to get your opinion if this was ok before I removed this

althea_types/src/wifi_info.rs
```diff
@@ -47,6 +47,12 @@ pub struct WifiStationData {
     pub current_time_ms: Option<u64>,
 }
 
+/// Format client uses to send to optools and frontend
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct ClientExtender {
+    pub mac_addr: String,
```
jkilpatr, line 53: don't we already have a mac address type somewhere? If not a number less than 2^48 would be a more restricted representation than a string, which could be literally anything.

rita_extender/src/lib.rs
jkilpatr: if that's the case go ahead and remove these.

rita_client/src/dashboard/extender_checkin.rs
jkilpatr: doc comments would be good here as well.

jkilpatr (approved):
This looks good, go ahead and clean up the lazy static if we can avoid it and add some doc comments to your new functions like extend hardware info.

PR #734, opened 2023-01-31
Chiara/integration tests

ch-iara:
adding in tests now

### DIFF
--- a/althea_kernel_interface/src/interface_tools.rs
+++ b/althea_kernel_interface/src/interface_tools.rs
@@ -231,6 +231,19 @@ impl dyn KernelInterface {
             Ok(())
         }
     }
+
+    /// Gets the network namespace name that holds the thread this function was called from
+    pub fn get_namespace(&self) -> Result<String, Error> {
+        let output = self.run_command("ip", &["netns", "identify"]).unwrap();
+        match String::from_utf8(output.stdout) {
+            Ok(mut s) => {
+                //this outputs with an extra newline \n on the end which was messing up the next command
+                s.truncate(s.len() - 1);
+                Ok(s)
+            }
+            Err(_) => Err(Error::RuntimeError("Could not get netns name!".to_string())),
+        }
+    }
 }
 
 #[test]
--- a/althea_kernel_interface/src/interface_tools.rs
+++ b/althea_kernel_interface/src/interface_tools.rs
@@ -232,16 +232,29 @@ impl dyn KernelInterface {
         }
     }
 
-    /// Gets the network namespace name that holds the thread this function was called from
-    pub fn get_namespace(&self) -> Result<String, Error> {
-        let output = self.run_command("ip", &["netns", "identify"]).unwrap();
+    /// Gets the network namespace name that holds the thread this function was called from.
+    /// If the calling thread was not inside a network namespace/in the default namespace, this
+    /// function returns a None
+    pub fn get_namespace(&self) -> Option<String> {
+        let output = match self.run_command("ip", &["netns", "identify"]) {
+            Ok(output) => output,
+            Err(_) => {
+                warn!("Could not run ip netns- is ip netns installed?");
+                return None;
+            }
+        };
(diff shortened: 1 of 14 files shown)


althea_kernel_interface/src/interface_tools.rs
```diff
@@ -231,6 +231,19 @@ impl dyn KernelInterface {
             Ok(())
         }
     }
+
+    /// Gets the network namespace name that holds the thread this function was called from
+    pub fn get_namespace(&self) -> Result<String, Error> {
+        let output = self.run_command("ip", &["netns", "identify"]).unwrap();
+        match String::from_utf8(output.stdout) {
```
jkilpatr, line 238: what happens if we run this in the default namespace? Does it return an empty string? 

Also what happens if we run it and ip netns is not installed? We should special case that and return the default namespace.  Right now it will panic on line 238 it doesn't make a huge amount of sense to handle the panic for interpreting the result as utf8 but not handling the error for if the command fails.

rita_common/src/peer_listener/mod.rs
```diff
@@ -161,21 +164,34 @@ fn listen_to_available_ifaces(peer_listener: &mut PeerListener) {
...
 /// peers over UDP
 pub fn peerlistener_tick() -> PeerListener {
     trace!("Starting PeerListener tick!");
 
-    let mut pl = get_pl_copy();
+    let key = if cfg!(feature = "integration_test_v2") {
+        KI.get_namespace().unwrap()
+    } else {
+        "default".to_string()
+    };
```
jkilpatr, line 189: the namespace handling should be abstracted into the get_pl_copy function. Also consider changing the pl functions from a total getter and setter to more operation specific stuff. Take a look at what's been done in the payment tracker module. For example a remove interface function would grab a write lock, remove the interface, then return. This sort of function structure of atomic changes avoids the risk of having a getter/setter pattern where another thread may make and save changes between when this thread gets the data from the lock and sets the version with it's own changes.

test_runner/src/tests.rs
```diff
@@ -0,0 +1,126 @@
...
+            //within routes there must be a route that matches the expected price between the dest (fd00::id) and the expected next hop (fe80::id)
+
+            let expected_data = expected.get(&ns1).unwrap().destination.get(ns2).unwrap();
+            let expected_cost = expected_data.clone().0;
+            let expected_hop = expected_data.clone().1;
+            let neigh_ip = IpAddr::V6(Ipv6Addr::new(
+                0xfe80,
+                0,
+                0,
+                0,
```
jkilpatr, line 93: I could swear we have a function somewhere for doing this sort of ip generation.

jkilpatr (changes requested):
This is absolutely awesome progress! 

I've left a few comments but I still need to grab this and run through it locally.

test_runner/src/tests.rs
ch-iara: We... do have a version of this in the existing test, but I figure until I'm needing to do this more often for multiple namespaces I can leave it like this (I did edit it to a one-liner which was annoying me as well because this is so long for no reason)

rita_common/src/peer_listener/mod.rs
```diff
@@ -156,52 +51,56 @@ fn listen_to_available_ifaces(peer_listener: &mut PeerListener) {
...
-
 /// Ticks the peer listener module sending ImHere messages and receiving Hello messages from all
 /// peers over UDP
 pub fn peerlistener_tick() -> PeerListener {
     trace!("Starting PeerListener tick!");
 
-    let mut pl = get_pl_copy();
+    let pl = match get_pl_copy() {
+        Some(pl) => pl,
+        // we should never get here in prod as the default is created in the initialization, but this is a failsafe either way...
```
jkilpatr, line 65: Why not move this check into get_pl_copy that way it's not in the main program logic.

jkilpatr (approved):
Just one last comment then we're good to go!

PR #735, opened 2023-01-31
Revamp the token bridge module

jkilpatr:
Revamp the token bridge module

This patch updates the token bridge module and changes how it works in
several ways

1) No more support for sending ETH directly, gas management caused the
   majority of the complexity in this module and also frustration for
   end users, who expect to get their total value into the router. In
   order to address this we will be sending ETH to routers to pay the
   fees. So no more ETH management is required.
2) Module refactor splitting what was previously one file into three to
   help with readability and make room for an upcoming Alhtea blockchain
   module

This will require an update to the dashboard as extensive changes to the
display states have also been made. But we're now envisoning a deposit
situation where the user does not really look at the router. So detailed
display isn't as high on our list of priorities.

This pr also contains an especially large clippy run so you may want to review each commit individually.

### DIFF
--- a/auto_bridge/src/lib.rs
+++ b/auto_bridge/src/lib.rs
@@ -227,8 +227,7 @@ pub async fn get_relay_message_hash(
         Ok(a) => a,
         Err(e) => {
             return Err(TokenBridgeError::Web3Error(Web3Error::BadInput(format!(
-                "Error: {}",
-                e
+                "Error: {e}"
             ))))
         }
     };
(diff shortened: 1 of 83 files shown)


auto_bridge/src/lib.rs
```diff
@@ -23,6 +23,10 @@ pub use error::TokenBridgeError;
 pub static UNISWAP_GAS_LIMIT: u128 = 150_000;
 pub static ERC20_GAS_LIMIT: u128 = 40_000;
 pub static XDAI_FUNDS_UNLOCK_GAS: u128 = 180_000;
```
ch-iara, line 25: Can these now be deleted? Looks like they are no longer in use

ch-iara (approved):

PR #736, opened 2023-02-03
fix wire_enabled bug

thosmos:
The value was lost in translation when converting from the Op Tools response to the local settings.  The merge_settings_safely fn was previously just dropping the new values.  It now actually merges them into settings which are then saved to memory later.


thosmos:
I added another commit that fixes the operator_fee display and behavior.  Happy to split into its own PR but they're both pretty small changes.

jkilpatr (approved):

PR #737, opened 2023-02-03
Add set wifi operator action

ptulugu:
This replaces the reset password operator action allows operators to set numerous wifi parameters remotely such as ssid, pass, etc

### DIFF
--- a/althea_kernel_interface/src/hardware_info.rs
+++ b/althea_kernel_interface/src/hardware_info.rs
@@ -1,5 +1,6 @@
 use crate::file_io::get_lines;
 use crate::KernelInterfaceError as Error;
+use crate::KI;
 use althea_types::extract_wifi_station_data;
 use althea_types::extract_wifi_survey_data;
 use althea_types::ConntrackInfo;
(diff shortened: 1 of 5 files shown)


althea_kernel_interface/src/hardware_info.rs
```diff
@@ -437,6 +441,57 @@ fn get_wifi_station_info(dev: &str) -> Vec<WifiStationData> {
...
+            None
+        }
+    }
+}
+
+pub fn get_radio_enabled(radio: &str) -> Option<bool> {
+    let radio = radio.replace("wlan", "radio");
+    let path = format!("wireless.{radio}.channel");
+    match KI.get_uci_var(&path) {
+        Ok(a) => Some(!a.contains('0')),
```
jkilpatr, line 460: what does this check do? comment it please.

althea_kernel_interface/src/hardware_info.rs
```diff
@@ -437,6 +441,57 @@ fn get_wifi_station_info(dev: &str) -> Vec<WifiStationData> {
...
+                "Unable to get radio channel for radio: {} with {:?}",
+                radio, e
+            );
+            None
+        }
+    }
+}
+
+pub fn get_radio_channel(radio: &str) -> Option<u16> {
+    let radio = radio.replace("wlan", "radio");
```
jkilpatr, line 472: this doesn't specify it's expected input or check that it's not getting a junk string input. At least comment what's the expected input format is since it seems to be doing a transformation on it.

jkilpatr (changes requested):
This looks good, two minor requests for improved comments

jkilpatr (approved):

PR #739, opened 2023-02-10
Exits dont panic when they see an ipv6 address is already taken on st…

ptulugu:
…artup

Instead of panicing, we simply skip adding an ipv6 addr here and add it at a different stage


jkilpatr (changes requested):
Please write a bit better of a commit message for this one, I find it a little difficult to understand the current commit title, it would be interpreted as exits *should* panic when they seen an ipv6 address already taken. It's also not exactly clear when 'later' is for adding the ip address.

jkilpatr (approved):

PR #741, opened 2023-03-01
Add display impl for OperatorAction

ch-iara:


ch-iara:
No longer needed; went with a different implementation to display to frontend ops

PR #751, opened 2023-04-03
WIP: Figure out exit bottleneck

ptulugu:


rita_exit/src/rita_loop/mod.rs
```diff
@@ -65,9 +65,19 @@ pub fn start_rita_exit_loop() {
...
                 // proper failover
                 let mut successful_setup: bool = false;
+                // cache of b19 routers we have successful rules and routes for
+                let mut wg_exit_clients: HashSet<WgKey> = HashSet::new();
+                // cache of b20 routers we have successful rules and routes for
+                let mut wg_exit_v2_clients: HashSet<WgKey> = HashSet::new();
 
                 loop {
-                    rita_exit_loop(&mut wg_clients, &mut debt_actions, &mut successful_setup)
+                    rita_exit_loop(
```
jkilpatr, line 74: in general you want to avoid mutable pointers are they are hard to reason about and debug. I'm sure you can refactor this such that it returns the data rather than taking a mutable reference.

rita_exit/src/database/mod.rs
```diff
@@ -370,20 +371,24 @@ pub fn cleanup_exit_clients(
 pub fn setup_clients(
     clients_list: &[exit_db::models::Client],
     old_clients: &HashSet<ExitClient>,
+    setup_wg_exit: &mut HashSet<WgKey>,
+    setup_wg_exit_v2: &mut HashSet<WgKey>,
```
jkilpatr, line 375: return pattern instead of mutable pointers here as well.

jkilpatr (changes requested):
This looks like it does what we need it to, what's the resulting exit loop time on the test machine? Also what is the startup time for first steup, lets look into shortening that next. 

"exit bottleneck" is vague. Please describe the issue in full in the commit message and I think the title could be a little more exact as far as describing the issue

rita_exit/src/database/mod.rs
ptulugu: done

rita_exit/src/rita_loop/mod.rs
ptulugu: done

rita_exit/src/database/mod.rs
```diff
@@ -464,46 +470,120 @@ pub fn setup_clients(
...
+        };
+    }
+
+    let new_v2 = &v2_clients - existing_v2;
+    let new_v1 = &v1_clients - existing_v1;
+
+    *existing_v2 = v2_clients;
+    *existing_v1 = v1_clients;
+
+    (new_v2, new_v1)
```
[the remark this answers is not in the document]
ptulugu: This is returning only clients that werent accounted for in previous tick, added comments

jkilpatr (approved):

jkilpatr (approved):

PR #752, opened 2023-04-05
Optimise startup setup time for rita exit

ptulugu:
Setup clients takes ~5 to complete on the first tick since it is settings up routes and rules for all clients in the database. This commit modifies so it sets ups routes and rules only for clients that have an active handshake value. There are over 1000 clients, but only around 200 actively handshaking, so this should reduce startup time dramtically

### DIFF
--- a/althea_kernel_interface/src/setup_wg_if.rs
+++ b/althea_kernel_interface/src/setup_wg_if.rs
@@ -124,6 +124,44 @@ impl dyn KernelInterface {
         }
         Ok(timestamps)
     }
+
+    /// Returns the last handshake time of every ACTIVE client on this tunnel.
+    /// An active handshake mean a wireguard tunnel that has a latest handshake value
+    /// When running wg show wg_exit latest-handshake, a entries with timestamp 0 are inactive
+    pub fn get_last_active_handshake_time(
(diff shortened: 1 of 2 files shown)


althea_kernel_interface/src/setup_wg_if.rs
```diff
@@ -124,6 +124,42 @@ impl dyn KernelInterface {
         }
         Ok(timestamps)
     }
+
+    /// Returns the last handshake time of every ACTIVE client on this tunnel.
```
jkilpatr, line 128: you should describe what active means here, essentially any device that has handshaked since the tunnel was created

jkilpatr (approved):

rita_exit/src/database/mod.rs
```diff
@@ -487,17 +487,36 @@ pub fn setup_clients(
     }
 
     // Setup ipv6 and v4 routes and rules for clients
+    // We optimise by setting up routes/rules only for those routers we see a wg tunnel with.
```
jkilpatr, line 490: this should be less ambiguous and specify that we look for handshake timesamps

jkilpatr (approved):

PR #764, opened 2023-04-28
Remove panics on ip addr commands

ptulugu:
As seen on large gateways, ip addr commands seem to fail when there are no file descriptors. This is a temporary fix and we will eventually move to parsing file systems or using netstat


jkilpatr (changes requested):
This is not the right way to approach solving this problem. By changing the way the functions return without changing how their calling functions expect them to behave we risk some really really weird and hard to debug behavior.

PR #776, opened 2023-05-11
Pranay/openwrt22

ptulugu:


althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -221,7 +168,6 @@ impl dyn KernelInterface {
         local_v4: Option<(IpAddr, u8)>,
         external_v6: Option<(IpAddr, u8)>,
         exit_mesh: IpAddr,
-        external_nic: String,
```
thosmos, line 224: why are you removing external_nic?  This is currently used to recognize a gateway router.  Is the expectation that this interface be moved to the peer_interfaces instead?  How does the gateway price get used instead of the relay price?

rita_client/src/exit_manager/exit_loop.rs
```diff
@@ -334,12 +334,13 @@ fn update_exit_server_to_althea_addresses() {
...
                             .expect("This shouldnt be failing"),
                     );
+                    info!("Updated us west alteha address to {:?}", e.althea_address)
                 }
             }
             _ => {}
         }
     }
-    // Save into settings
+    // save new rita client
```
thosmos, line 343: just to clarify this does not actually save the settings to disk but only updates the settings in the memory var

thosmos (commented):
It looks like the choice was made to parse and serialize CLI nftables commands instead of using either the JSON or the NetLink libraries.  I'm curious of the reason because there was the intention to move away from parsing CLI text.

I haven't actually tested this code yet.  I'll do that and follow up here.

althea_kernel_interface/src/exit_server_tunnel.rs
ptulugu: I moved the iptables commands from this function into the setup_nat function to keep things congruent since that function was already setting up v4 forward rules. the external_nic parameter moved with it.

Regarding using KI over libnftables json, the library wasnt able to parse some rules, for example
'nft insert rule inet fw4 input ip6 daddr . meta iifname != @rita_input add @rita_input { ip6 daddr . meta iifname counter }', if we were to use that crate we need to update the structs as well as do better error handling, I recall there being an unwrap that caused panics during testing.
In general I also think the Statement::Match syntax it uses makes it hard to read. Less external dependencies is a plus

rita_client/src/exit_manager/exit_loop.rs
ptulugu: updated wording here

jkilpatr:
@thosmos  we would really like to move away from cli commands where possible, the issue is that the json api for whatever reason didn't support some of the commands we needed for atomic counters (ipset replacement) which is key to billing.

jkilpatr (approved):

PR #779, opened 2023-05-16
Bugfix: save settings after setting up althea addresses on trasitioni…

ptulugu:
…ng routers


ptulugu (commented):
Are these hardcoded addresses correct? I do eth bytes -> ethermint private key -> address here

ptulugu (commented):
Are these hardcoded addresses correct?
I do eth bytes -> ethermint privatekey -> address here

PR #783, opened 2023-05-24
Validate payment with Usage History before adding pmt

ptulugu:
This commit fixes the following edge case. Router A is paying B. B reboots and loses all payment history in payment validator. Next time A sends all its txids over, B revalidates pmts that have already been accounted for  leading to double counting. This patch fixes that by looking through entire usage history for the payment txid before adding that payment to debt keeper and usage tracker

### DIFF
--- a/rita_common/src/payment_validator/mod.rs
+++ b/rita_common/src/payment_validator/mod.rs
@@ -794,7 +794,6 @@ fn handle_tx_messaging(transaction: TransactionDetails, ts: ToValidate, current_
                 pmt.amount,
                 denom.expect("How did this happen when we already verified existence"),
             );
-
             // update the usage tracker with the details of this payment
             update_payments(pmt);
         }
@@ -819,13 +818,13 @@ fn handle_tx_messaging(transaction: TransactionDetails, ts: ToValidate, current_
                 tx: ts,
                 success: true,
             });
+
             // update debt keeper with the details of this payment
             let _ = payment_succeeded(
                 pmt.to,
                 pmt.amount,
                 denom.expect("How did this happen when we already verified existence"),
             );
-
             // update the usage tracker with the details of this payment
             update_payments(pmt.clone());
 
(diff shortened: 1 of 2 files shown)


rita_common/src/usage_tracker/mod.rs
```diff
@@ -84,19 +84,27 @@ pub struct FormattedPaymentTx {
...
+    if let Some(txid) = input.txid {
+        FormattedPaymentTx {
             to: input.to,
             from: input.from,
             amount: input.amount,
             txid: format!("{txid:#066x}"),
-        },
-        None => FormattedPaymentTx {
+        }
+    } else if let Some(tx_hash) = input.tx_hash {
```
jkilpatr, line 94: why was this change needed?

rita_common/src/payment_validator/mod.rs
```diff
@@ -788,15 +789,17 @@ fn handle_tx_messaging(transaction: TransactionDetails, ts: ToValidate, current_
...
 
-            // update debt keeper with the details of this payment
-            let _ = payment_received(
-                pmt.from,
-                pmt.amount,
-                denom.expect("How did this happen when we already verified existence"),
-            );
+            if !payment_accounted_for(pmt.clone()) {
+                // update debt keeper with the details of this payment
+                let _ = payment_received(
```
jkilpatr, line 794: call payment accounted for inside update payments, but send it off to the payment validator, part of hte goal of re-sending payment data was to ensure routers didn't enforce on each other, right now the only bug we're trying to fix is duplicate data in the earnings display, which won't be affected by us applying a payment twice due to the way the internals of debt keeper work.

jkilpatr (changes requested):

rita_common/src/usage_tracker/mod.rs
ptulugu: In b21 pmt struct can either be txid from eth or tx_hash from althea, guess i missed this change during the adding althea wallets pr

jkilpatr (approved):

PR #787, opened 2023-06-05
Optimize exit enforcment flow creation

jkilpatr:
Enforcement on the exits for clients has two parts, one where we create a classifier in tc that is used to classify traffic for a specific ip and then another where we take the traffic class we have assigned a given ip to and throttle that traffic class.

When we upgraded to ipv6 exits we the code added traffic classes to every exit interface for every possible user and each user ipv6 ip. For reference if there are 1k active users across 6 different exits with distinct ipv6 ip's that means there where 7k different classes on the exit interface despite it not being possible to use 5/7 of those ips at all! Even worse only a few hundred of those users would be active, and an even smaller number enforced. But the exit kernel spends time attempting to sort every incoming ip into one of those classes.

This change makes it so that we only create the class when it is time to enforce on a specific user, decreasing the scope of traffic classification from nearly 10k ip's down to probably about 100 at most.

### DIFF
--- a/althea_kernel_interface/src/exit_server_tunnel.rs
+++ b/althea_kernel_interface/src/exit_server_tunnel.rs
@@ -23,8 +23,7 @@ impl dyn KernelInterface {
         listen_port: u16,
         private_key_path: &str,
         if_name: &str,
-        ipv6_filter_handles: HashSet<(String, u32)>,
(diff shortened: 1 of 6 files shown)


ch-iara (approved):
Looks much more cohesive with the new constants as well as the functional change for optimization

PR #789, opened 2023-06-07
Exits in testrunner

jkilpatr:
This is a big conglomeration of the code I've gotten done today, mostly around getting the test environment prepped for running exits and then reducing unused lazy statics in a quest to prepare them all for the test environment.

Big changes
* settings now works like peer listener used to work (checking namespaces and returning the correct data)
*  peer listener doesn't use lazy static at all anymore
*  exit running in five_node_testnet


ch-iara (approved):
Namespace agnostic constants with the check_integration_test_netns is really good

PR #790, opened 2023-06-11
 testrunner exit registration

jkilpatr:
This patch gets exit registration and setup working in the test
environment.

Sadly the exit clients still can't reach the outside world, since the
nftables exit forwarding setup does not seem to have been tested.


jkilpatr:
One fun thing I noticed while doing this is that all encrypted messages between the client and the exit take an exit client identity, which contains all the info needed for *registration* this results in strange codepaths where you need contact details set to request the list of active exits even though nothing on the other end touches this data.

This breaks my usual condition of 'never send data that's not used' and is also problematic as we go forward trying to re-model the exit organization to match the new paradigm.

jkilpatr:
I've also removed the separate althea_address field, it was simply more trouble than it was worth by a long shot handling the various possibilities is just not easy.

ch-iara (approved):

PR #793, opened 2023-06-21
WIP: Add althea payments test

ptulugu:


integration_tests/src/payments_althea.rs
```diff
@@ -1,2 +1,125 @@
...
+        .unwrap()
+        .into()
+}
+
+pub fn get_althea_evm_pub() -> AltheaAddress {
+    get_althea_evm_priv().to_address(ALTHEA_PREFIX).unwrap()
+}
+
+pub fn get_althea_grpc() -> String {
+    format!("http://{}:9091", NODE_IP)
```
jkilpatr, line 35: This should probably be used as the constant here as well https://github.com/althea-net/rita/blob/master/integration_tests/src/utils.rs#L212 just for consistency.

integration_tests/src/payments_althea.rs
```diff
@@ -1,2 +1,125 @@
...
+};
+use althea_types::ALTHEA_PREFIX;
+use deep_space::Address as AltheaAddress;
+use deep_space::{EthermintPrivateKey, PrivateKey};
+use log::info;
+
+const USDC_TO_WEI_DECIMAL: u64 = 1_000_000_000_000u64;
+
+const ALTHEA_EVM_PRIV_BYTES: &str =
+    "3b23c86080c9abc8870936b2eb17ecb808f5ad3b318018b3e23873013379e4d6";
```
jkilpatr, line 21: it looks like this is just a random testing key, it would be better to have a utility function that generates a random key rather than using a hardcoded one,. A random key is both a better test (may find issues that only some keys have) and prevents things like hardcoding addresses in the test that don't need to be constants.

integration_tests/src/payments_althea.rs
```diff
@@ -1,2 +1,125 @@
...
+
+    let mut from_node: Option<Namespace> = None;
+    let mut forward_node: Option<Namespace> = None;
+    let mut end_node: Option<Namespace> = None;
+    // top up routers
+    for ns in namespaces.names.clone() {
+        info!("Namespace : {:?}", ns);
+        if ns.id == 1 {
+            from_node = Some(ns.clone());
+        }
```
jkilpatr, line 88: I think there's a utility function for grabbing namespaces by id that could replace this.

integration_tests/src/payments_althea.rs
```diff
@@ -1,2 +1,125 @@
...
+    );
+    send_althea_tokens(to_topup.clone()).await;
+
+    let balances = print_althea_balances(to_topup.clone(), "uUSDC".to_string()).await;
+    info!("USDC Balances are {:?}", balances);
+    let balances = print_althea_balances(to_topup, "aalthea".to_string()).await;
+    info!("Althea Balances are {:?}", balances);
+
+    //info!("After registering routers, sleeping for 30 sec");
+    //thread::sleep(Duration::from_secs(3000));
```
jkilpatr, line 114: maybe remove this commented code if it's not used?

integration_tests/src/utils.rs
```diff
@@ -261,3 +324,370 @@ pub fn set_sigterm() {
...
+            &["netns", "exec", &from.get_name(), "iperf3", "-c", &ip],
+        )
+        .expect("Could not setup iperf client");
+    let output = from_utf8(&output.stdout).expect("could not get output for client setup!");
+    info!("Client out: {}", format!("{}", output));
+}
+
+pub async fn query_debts(node: Namespace, for_node: Namespace) -> GetDebtsResult {
+    let for_node_ip = match for_node.id {
+        4 => EXIT_ROOT_IP.to_string(),
```
jkilpatr, line 353: we should check if the namespace is an exit, rather than relying on hardcoding id's

one of the next tests we want to implement is a multi exit test to test exit roaming

integration_tests/src/utils.rs
```diff
@@ -261,3 +324,370 @@ pub fn set_sigterm() {
...
+            key.clone(),
+            Some(timeout),
+        )
+        .await;
+    while let Err(e) = res {
+        contact.wait_for_next_block(TOTAL_TIMEOUT).await.unwrap();
+        res = contact
+            .vote_on_gov_proposal(
+                proposal_id,
+                VoteOption::Yes,
```
jkilpatr, line 457: why are these failing? timeout?

integration_tests/src/utils.rs
```diff
@@ -261,3 +324,370 @@ pub fn set_sigterm() {
...
+        .into_inner()
+        .metadata
+        .expect("No metadata for erc20 coin");
+
+    let coin_params = RegisterCoinProposalParams {
+        coin_metadata: metadata.clone(),
+        proposal_desc: "Register Coin Proposal Description".to_string(),
+        proposal_title: "Register Coin Proposal Title".to_string(),
+    };
+    execute_register_coin_proposal(
```
jkilpatr, line 622: This looks like it's always running the coin proposal even if it's already been passed, that's going to really increase the amount of time it takes to run this function from a few seconds to two minutes or so for no particularly good reason.

test_runner/src/main.rs
```diff
@@ -32,7 +32,7 @@ async fn main() {
         } else if test_type == "PAYMENTS_ETH" {
             run_eth_payments_test_scenario()
         } else if test_type == "PAYMENTS_ALTHEA" {
-            run_althea_payments_test_scenario()
+            run_althea_payments_test_scenario().await;
```
jkilpatr, line 35: ok you also need to add this test case to .github/workflows/integration-test.yml so that it runs in CI

jkilpatr (changes requested):
This looks good I've left a few comments about cleanups and hopefully reducing the runtime by about 8 minutes by removing duplicate proposals.

integration_tests/src/payments_althea.rs
ptulugu: done

integration_tests/src/payments_althea.rs
ptulugu: done

integration_tests/src/utils.rs
ptulugu: Done

test_runner/src/main.rs
ptulugu: done

integration_tests/src/payments_althea.rs
ptulugu: done

integration_tests/src/payments_althea.rs
ptulugu: I grab this from setup-validators.sh

integration_tests/src/utils.rs
ptulugu: Not sure what you mean here, doesnt seem like its failing?

integration_tests/src/utils.rs
ptulugu: I separated this into its own function

integration_tests/src/utils.rs
jkilpatr: I mean why do you need the retry?

jkilpatr (approved):

PR #800, opened 2023-07-13
Debts test

ptulugu:
Debts generated by traffic between two nodes as well as between a node and the internet can be
tested by hardcoding an optimal path in code. Current accuracy threshold between calculated debt
and the actual debt is set to at least 85% accurate

### DIFF
--- a/integration_tests/src/debts.rs
+++ b/integration_tests/src/debts.rs
@@ -1,0 +1,360 @@
+use std::collections::HashMap;
+use std::thread;
+use std::time::Duration;
+
+use crate::five_nodes::five_node_config;
+use crate::setup_utils::database::start_postgres;
+use crate::setup_utils::namespaces::*;
+use crate::setup_utils::rita::thread_spawner;
+use crate::utils::{
+    generate_traffic, get_default_settings, get_ip_from_namespace, query_debts, register_to_exit,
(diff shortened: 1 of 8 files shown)


integration_tests/src/debts.rs
```diff
@@ -0,0 +1,353 @@
...
+    5(40)         6(20)
+    */
+
+    // Hardcoded debt pairs and optimal routes. 0 represents the internet
+    let pairs: Vec<(u16, u16)> = vec![(7, 0), (3, 0), (7, 2), (5, 1), (1, 4), (2, 5), (7, 5)];
+
+    // Hardcoded optimal paths for each node pair. Validation uses this to test debts for each node in path
+    let mut optimal_routes: HashMap<(u16, u16), Vec<u16>> = HashMap::new();
+    optimal_routes.insert((7, 0), vec![4]);
+    optimal_routes.insert((3, 0), vec![6, 4]);
```
jkilpatr, line 46: this could be pulled out into it's won function like the five node setup.

jkilpatr (approved):
Looks good, my only comment is that it would be nice if the main body of the test where broken into a few functions (only a few things look like they could reused, but it would be nice to have those easily available in their own tests) 

I haven't seen this run successfully yet so I still need to watch the test.

integration_tests/src/debts.rs
ptulugu: fixed this

PR #805, opened 2023-07-27
Add NO_IMAGE_BUILD flag to test runner tests

ptulugu:


jkilpatr:
closing this because we would need to increase default runner size to make docker layer caching work. I'm not sure why this works over on the gravity bridge repo but not here. In this case github loses since we'll end up spending more free CI minutes than we would otherwise with this optimization.

PR #806, opened 2023-07-27
Send usage data in 1-month portions

ch-iara:
This change is meant to speed up the checkin loop when routers have a lot of data to upload for usage history. We now only iterate through and send up to a month's worth of hours per checkin.


rita_client/src/operator_update/mod.rs
```diff
@@ -168,48 +169,69 @@ pub async fn operator_update(ops_last_seen_usage_hour: Option<u64>) -> u64 {
...
-                    break;
                 }
             }
-
-            // pop front, add to front of new vecdeque if exists.
-            while let Some(data) = usage_data_relay.pop_front() {
-                if data.index > last_seen_hour {
+        } else {
+            // binary search until we find the last seen hour or hour in the vecdeque with index before that (as in cases of gaps in data)
+            let pos = find_position_since_last_seen(&usage_data_client, last_seen_hour);
```
jkilpatr, line 210: are we sure the array is sorted? you should probably call sort before you call this search (most sort algos are pretty efficient if the array is already sorted) 

Even if we *think* it's sorted on most routers I'm sure we'll have at least one weird one.

Also need a unit test that tries feeding in unsorted data just to be sure here.

rita_client/src/operator_update/mod.rs
```diff
@@ -168,48 +169,69 @@ pub async fn operator_update(ops_last_seen_usage_hour: Option<u64>) -> u64 {
...
+    // We check that the difference is >1 because we leave a 1 hour buffer to prevent from sending over an incomplete current hour
+    let send_hours = current_hour - last_seen_hour > 1;
     let mut usage_tracker_data: Option<UsageTracker> = None;
-    // only deal with this if we actually do need to send some hours
     // if ops_last_seen_usage_hour is a None the thread has restarted and we are waiting for ops to tell us how much
-    // data we need to send, which will be populated with the next checkin cycle
-    if hours_to_send != 0 && ops_last_seen_usage_hour.is_some() {
+    // data we need to send, which will be populated with the next checkin cycle. 730 is the average numbers of hours
+    // in a month, and we only send 1 month at a time if ops is requesting the full usage history.
+    if send_hours && ops_last_seen_usage_hour.is_some() {
```
jkilpatr, line 178: this should all be broken out into a function for easier unit testing + organization this is a big if block.

jkilpatr (changes requested):
This looks like it works ok but I'm going to insist on some unit tests.

I've also recommended how you can avoid using a self-implemented binary search. You may want to add sorting the usage tracker list to the startup cleanup (where we also remove duplicate hours)

rita_client/src/operator_update/mod.rs
ch-iara: way compressed the block but this can be broken out if it's still too big a chunk

jkilpatr: you shouldn't be running any of the loops in here. That should be in it's own function to be tested.

rita_client/src/operator_update/tests.rs
```diff
@@ -0,0 +1,248 @@
...
+        // Options here to optimize are either a/write my own binary sort again which will compare for the existing structure
+        // where the saved vecdeque is highest index to lowest index, b/rework usage tracker so that we save data lowest index
+        // to highest index, or c/the current solution(inefficient, as we will be fully reversing the whole vecdeque of each
+        // client and relay every checkin): sort the entire list in reverse order to use with the builtin bin search from vecdeque
+
+        // this here sorts from lowest index to highest index, so we end with a vecdeque that we can use binary search on
+        usage_data_client
+            .make_contiguous()
+            .sort_by(|a, b| a.index.cmp(&b.index));
+        println!(
```
jkilpatr, line 169: you are testing what is currently the same code but might change.. break out whats in the update loop into a function so you can call that function here.

rita_client/src/operator_update/mod.rs
```diff
@@ -680,134 +649,65 @@ fn contains_forbidden_key(map: Map<String, Value>, forbidden_values: &[&str]) ->
...
+    // sort client and relay data in case they have come out of order somehow. This sorts by index increasing so newest data at back
+    usage_data_relay
+        .make_contiguous()
+        .sort_by(|a, b| a.index.cmp(&b.index));
+    usage_data_client
+        .make_contiguous()
+        .sort_by(|a, b| a.index.cmp(&b.index));
+
+    // so this spits out the index for where last seen is, or the index of the next highest hour(returned in an error).
+    // we take the result -1 just in case, limit 0, since it's possible we might get back an index out of bounds at the back.
```
jkilpatr, line 691: this means we may send a duplicate usage hour to ops right? Because we send result -1. Have you checked if the code there will handle that gracefully?

rita_client/src/operator_update/tests.rs
```diff
@@ -0,0 +1,260 @@
...
+        assert!(res_usage_relay.get(0).unwrap().index == last_seen_hour - 1);
+        assert!(res_usage_client.get(0).unwrap().index == last_seen_hour - 1);
+
+        // check that our iteration function does indeed stop at a month of data:
+        assert!(res_usage_client.len() <= 730);
+        assert!(res_usage_relay.len() <= 730);
+    }
+
+    // generates a usage tracker struct for testing without payments since these do not get sent up in ops updates.
+    // using this while I can't get the import working... as a note the original function needs to be updated to push to back
```
jkilpatr, line 231: this comment may need to be updated? It doesn't look like you ended up needing to duplicate the function.

rita_client/src/operator_update/tests.rs
```diff
@@ -0,0 +1,260 @@
...
+            exit_bandwidth: VecDeque::new(),
+            payments: VecDeque::new(),
+        }
+    }
+    #[cfg(test)]
+    // generates dummy usage hour data randomly
+    fn generate_bandwidth(starting_hour: u64) -> VecDeque<RCUsageHour> {
+        use rand::{thread_rng, Rng};
+        // 8760 is the max number of saved usage entries(1 year)
+        let num_to_generate: u16 = thread_rng().gen_range(50..8760);
```
jkilpatr, line 248: instead of using a hardcoded 8760 the constant MAX_USAGE_ENTRIES should be made public and used.

rita_client/src/operator_update/tests.rs
jkilpatr: also please generate with payments if you can, you never know when we might start having those here.

rita_client/src/operator_update/mod.rs
ch-iara: Ops does currently handle in case it has been sent duplicates, but regardless I will need to update ops because as is now I've switched the order in which the data comes in

rita_client/src/operator_update/tests.rs
ch-iara: revised the rita common usage tracker function to be public to operator update in rita client

jkilpatr (approved):

PR #814, opened 2023-08-04
Update exit checkin to send a list of registered wgkey

ptulugu:
The exit checkin now sends a list of WgKey it retrieves from the database during exit checkin. Ops can use this to display more accurately whether a router is registered or not.]

### DIFF
--- a/rita_exit/src/operator_update/update_loop.rs
+++ b/rita_exit/src/operator_update/update_loop.rs
@@ -13,6 +13,7 @@ pub fn start_operator_update_loop() {
     thread::spawn(move || {
         // this will always be an error, so it's really just a loop statement
         // with some fancy destructuring
+
         while let Err(e) = {
             thread::spawn(move || loop {
                 let start = Instant::now();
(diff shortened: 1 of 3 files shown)


rita_exit/src/operator_update/update_loop.rs
```diff
@@ -13,24 +27,35 @@ pub fn start_operator_update_loop() {
...
-                    operator_update(rita_started).await;
-                });
-
-                info!(
-                    "exit Operator Update loop completed in {}s {}ms",
-                    start.elapsed().as_secs(),
-                    start.elapsed().subsec_millis()
-                );
+            thread::spawn(move || {
+                // No need to query database every 60secs to get the registered list, keep a cache and
```
jkilpatr, line 33: why cache versus setting the checkin time to every 10 minutes? 

On the ops side people may be a little confused to not see the output of their register command for 10 minutes. 

Considering that the exits have hundreds of routers making db requests pretty often,  I think we shouldn't bother with this cache until we can confirm we really need it. Any caching or optimization really needs to happen on the router status checkin side.

jkilpatr (changes requested):

rita_exit/src/operator_update/update_loop.rs
ptulugu: Ok I removed the cache

jkilpatr (approved):

PR #817, opened 2023-08-10
Fix 'last exit' check in exit loop

ptulugu:
Previously, to reestablish exit wgtunnels on the client, the ExitServer struct would be compared to check for changes across ticks, which accounted for changes in registration details (ExitState struct). While introducing clustering, this was changed to check that instance ip changed which led to a bug where client who lose their reg details arent able to update the wg exit tunnels. Now we check both to verify that wg tunnels should be setup again


jkilpatr (approved):
Great test here!

PR #824, opened 2023-08-18
Exit db refactor

ptulugu:
The goal of this change is to make it possible for many thousands of exits
to easily run off a single database with no shared credentials and also to
make it possible for many different organizations to cooperate on this database.

In order to achieve this we will be replacing the postgresql database currently
used by the exits with a smart contract which will be used to register users

Client register flow has changed from client -> Exit -> Ops. Exit forwards a request to an external server with sms api key in order to
verify a client. That server will have write access to the smart contract to add the client, and then return their registration status to
exit, which then forwards that request back to the client

- Solidity contract currently doesnt have any access restrictions
- There is currently no cleanup for old clients


integration_tests/src/contract_test.rs
```diff
@@ -0,0 +1,200 @@
...
+    .unwrap();
+    assert_eq!(res, user_2);
+
+    let res = get_registered_client_using_meship(
+        user_3.mesh_ip,
+        ETH_MINER_KEY
+            .parse::<clarity::PrivateKey>()
+            .unwrap()
+            .to_address(),
+        get_altheadb_contract_addr(),
```
jkilpatr, line 194: 👍🏻  this is good as far as what's tested.

You could conserve a lot of lines by grabbing the ETH_MINER_KEY once at the top of the function and assigning it to a local var.

There's also a function in the debts tests in rita common that will generate random identities for you, might be useful to move that around so you can import it here and generate the identities, it's a little better at finding problems than hardcoded ID's and also is more compact.

integration_tests/src/utils.rs
```diff
@@ -131,6 +135,40 @@ pub fn get_eth_node() -> String {
...
+        .args([
+            "ts-node",
+            "/althea_rs/solidity/contract-deployer.ts",
+            &format!("--eth-privkey={}", MINER_PRIVATE_KEY),
+            &format!("--eth-node={}", ETH_NODE),
+        ])
+        .output()
+        .expect("Failed to deploy contracts!");
+
+    info!("Contract deploy returned {:?}", from_utf8(&res.stdout));
```
jkilpatr, line 169: it might be easiest to parse the output here and return the deployed contract address.

integration_tests/src/utils.rs
```diff
@@ -131,6 +135,40 @@ pub fn get_eth_node() -> String {
     format!("http://{}:8545", NODE_IP)
 }
 
+pub fn get_altheadb_contract_addr() -> Address {
+    "0xb9b674D720F96995ca033ec347df080d500c2230"
```
jkilpatr, line 139: this hardcoded address won't always be correct, in fact I'm not sure it would even work twice, (it might if all conditions on the testnet chain are exactly repeatable, but I wouldn't bet on it) 

you should parse the contract address from the deployment script. I'm sure there's an example in the gravity repo.

integration_tests/src/utils.rs
```diff
@@ -63,6 +64,9 @@ pub const STAKING_TOKEN: &str = "aalthea";
 pub const MIN_GLOBAL_FEE_AMOUNT: u128 = 10;
 pub const TOTAL_TIMEOUT: Duration = Duration::from_secs(300);
 pub const DEBT_ACCURACY_THRES: u8 = 15;
+pub const ETH_NODE: &str = "http://localhost:8545";
+pub const MINER_PRIVATE_KEY: &str =
```
jkilpatr, line 68: instead of parsing this every time everywhere just having a function that does it and returns the actual PrivateKey struct will probably save us a lot of duplicated code.

integration_tests/src/registration_server.rs
```diff
@@ -0,0 +1,83 @@
...
+    );
+
+    // Start endpoint listener
+    thread::spawn(move || {
+        let runner = System::new();
+        runner.block_on(async move {
+            // Exit stuff, huge threadpool to offset Pgsql blocking
+            let _res = HttpServer::new(|| {
+                App::new()
+                    .route("/register_router", web::post().to(register_router))
```
jkilpatr, line 37: 👍🏻

integration_tests/src/payments_althea.rs
```diff
@@ -65,11 +73,28 @@ pub async fn run_althea_payments_test_scenario() {
 
     test_routes(namespaces.clone(), expected_routes);
 
+    // Exits need to have funds to request a registered client list, which is needed for proper setup
+    info!("Topup exits with funds");
+    let web3 = Web3::new("http://localhost:8545", WEB3_TIMEOUT);
+    let mut to_top_up = Vec::new();
+    for c in rita_identities.client_identities.clone() {
```
jkilpatr, line 80: this should probably be a utility function since we'll want to do it at the start of every test.

rita_client_registration/src/client_db.rs
```diff
@@ -0,0 +1,338 @@
...
+                        "Cannot parse array offset: {:?} with {}",
+                        byte_chunks[index + 1],
+                        e
+                    );
+                    continue;
+                }
+            };
+
+        ret.push(
+            match parse_identity_abi_internal(byte_chunks[(index + arr_start_offset)..].to_vec()) {
```
jkilpatr, line 323: this module definitely needs some unit tests.

rita_client_registration/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+};
+
+#[macro_use]
+extern crate log;
+#[macro_use]
+extern crate lazy_static;
+
+pub mod client_db;
+
+lazy_static! {
```
jkilpatr, line 37: this lazy static looks like logic that should be internal to the server logic. Maybe try and break it out as 'client registration server' and 'client registration lib' where Rita client for example only imports the lib.

rita_exit/src/database/mod.rs
```diff
@@ -361,19 +305,46 @@ pub fn setup_clients(
...
+                }
+            }
+            Err(e) => {
+                error!(
+                    "Unable to convert client to ExitClient! {} with error {}",
+                    c.wg_public_key, e
+                );
+            }
+        }
+        //key_to_client_map.insert(c.wg_public_key, c.clone());
```
jkilpatr, line 321: was this needed?

rita_exit/src/database/mod.rs
```diff
@@ -87,150 +68,171 @@ pub async fn signup_client(
...
-    conn: &PgConnection,
+    our_address: Address,
+    contract_addr: Address,
+    contact: &Web3,
 ) -> Result<ExitState, Box<RitaExitError>> {
     trace!("Checking if record exists for {:?}", client.global.mesh_ip);
 
-    if let Some(their_record) = get_client(&client, conn)? {
-        trace!("record exists, updating");
+    match get_registered_client_using_wgkey(
```
jkilpatr, line 179: I need to review the logic here in more detail.

rita_exit/src/database/struct_tools.rs
```diff
@@ -1,89 +1,246 @@
...
+pub const CLIENT_WG_PORT: u16 = 59999;
+
+/// Max number of time we try to generate a valid ip addr before returning an eror
+pub const MAX_IP_RETRIES: u8 = 10;
+
+lazy_static! {
+    /// Keep track of ip addrs assigned to clients and ensure collisions dont happen. In worst case
+    /// the exit restarts and loses all this data in which case those client they had collision may get new
+    /// ip addrs and would need to setup wg exit tunnel again
+    static ref IP_ASSIGNMENT_MAP: Arc<RwLock<IpAssignmentMap>> = Arc::new(RwLock::new(IpAssignmentMap::default()));
```
jkilpatr, line 24: it's a bit of a nit but I'd call this file 'in memory database' because effectively we're building an in memory database based on what we get from the smart contract. Also we want there to be a single lazy static for all the exit setup state so that different lazy statics remain internally consistent. Also this needs to be indexed for the test env.

rita_exit/src/operator_update/mod.rs
```diff
@@ -51,86 +41,16 @@ pub async fn operator_update(rita_started: Instant) {
...
                 pass,
                 exit_uptime: rita_started.elapsed(),
-                registered_keys: get_registered_list(),
                 // Since this checkin works only from b20, we only need to look on wg_exit_v2
                 users_online: KI.get_wg_exit_clients_online(EXIT_INTERFACE).ok(),
             })
             .await;
-
-        let response = match response {
-            Ok(mut response) => {
```
jkilpatr, line 68: good cleanup here.

jkilpatr (changes requested):
Wow this is a big one! 

I've left some comments, mostly about duplicated code that should be de-duplicated and cleaned up into utility functions.

Looks like clippy still needs fixing along with the CI tests not completing

integration_tests/src/contract_test.rs
ptulugu: done

integration_tests/src/utils.rs
ptulugu: done

integration_tests/src/utils.rs
ptulugu: done

integration_tests/src/utils.rs
ptulugu: done

integration_tests/src/payments_althea.rs
ptulugu: done

rita_exit/src/database/struct_tools.rs
ptulugu: done

ptulugu: done

jkilpatr (approved):

PR #831, opened 2023-09-15
Refactor UsageTracker

jkilpatr:
This simple double ended iterator based design of UsageTracker has served us pretty well for a long time. We would push to the front a new usage hour and pop old ones off the back.

Sadly this format has proven to not be very robust to errors, if for any reason the usage or payments list was to become out of order invalid duplicate data would exist leading to confusing questions about which one to select.

We saw this most recently with payments where the make_payments_v2 change caused duplicate payments and incorrect income data to pollute the databases of many routers. That issue was patched with a duplicate check but in this new structure duplicates of payments in the payment list and duplicates of data usage in the data usage list are simply not possible.

By storing the usage data as a hashmap indexed by unix timestamp we will always retrieve a copy of the usage data for a given hour if we request it and only one version of any hour can exist.

Likewise for payments we've moved to a flat HashSet where the Hash value for the payment struct is the txid, this means that duplicate payments with the same txid can not co-exist in the hashset.

This new more robust format will be consumed by operator tools as well making it easier to manage usage data for many routers in the database there.

### DIFF
--- a/rita_client/src/operator_update/mod.rs
+++ b/rita_client/src/operator_update/mod.rs
@@ -20,8 +20,8 @@ use num256::Uint256;
 use rita_common::rita_loop::is_gateway;
 use rita_common::tunnel_manager::neighbor_status::get_neighbor_status;
 use rita_common::tunnel_manager::shaping::flag_reset_shaper;
-use rita_common::usage_tracker::structs::UsageType::{Client, Relay};
-use rita_common::usage_tracker::{get_current_hour, get_usage_data_map};
+use rita_common::usage_tracker::structs::UsageType::{self, Client, Relay};
+use rita_common::usage_tracker::{get_current_hour, get_current_throughput, get_usage_data_map};
 use rita_common::utils::option_convert;
 use rita_common::DROPBEAR_AUTHORIZED_KEYS;
 use rita_common::KI;
@@ -161,6 +161,8 @@ pub async fn operator_update(
             rita_uptime: RITA_UPTIME.elapsed(),
             user_bandwidth_usage: None,
             user_bandwidth_usage_v2: prepare_usage_data_for_upload(ops_last_seen_usage_hour)?,
+            client_mbps: get_current_throughput(UsageType::Client),
+            relay_mbps: get_current_throughput(UsageType::Relay),
         })
         .await;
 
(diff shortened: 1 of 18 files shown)


rita_client/src/operator_update/mod.rs
```diff
@@ -626,65 +596,39 @@ fn contains_forbidden_key(map: Map<String, Value>, forbidden_values: &[&str]) ->
...
+    ops_last_seen_hour: Option<u64>,
+) -> Result<Option<UsageTrackerTransfer>, RitaClientError> {
+    // if this is none we will not send usage data to ops
+    // as it indicates we don't yet know what data ops has
+    match ops_last_seen_hour {
+        Some(ops_last_seen_hour) => {
+            let current_hour = get_current_hour()?;
+            let usage_data_client = get_usage_data_map(Client);
+            let usage_data_relay = get_usage_data_map(Relay);
+            // We check that the difference is >1 because we leave a 1 hour buffer to prevent from sending over an incomplete current hour
```
ch-iara, line 610: Also I believe this comment is now stale as this is no longer checking for 1 hour behind (This may still be warranted- since part of the original problem was that routers were sending through incomplete hours, as the "current hour" is updated every couple seconds, and this can still happen if we send all the way up to the current hour. We're handling this on the ops side as well and only saving an hour behind the current to allow routers to complete their usage hours before we save the data but for the sake of not sending through hours that will not be saved may be worth changing)

ch-iara: Unless we still want this data as displayed the way we have minute-spectrum with the latency graph, but I think that will need a different  struct as you mentioned way back

rita_client/src/operator_update/mod.rs
```diff
@@ -626,65 +596,39 @@ fn contains_forbidden_key(map: Map<String, Value>, forbidden_values: &[&str]) ->
...
+            let min_send_hour = current_hour - ops_last_seen_hour;
+            let mut client_bandwidth = HashMap::new();
+            let mut relay_bandwidth = HashMap::new();
+            for (index, usage) in usage_data_client {
+                if index > min_send_hour {
```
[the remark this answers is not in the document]
jkilpatr: You're right this logic is incorrect. I intended min_send_hour to be the last seen hour, thanks for the flag.

As far as looping O(n) is the best possible in this case, because keys in a hashmap are unordered, we could get the keys, sort them, then request only keys greater than last seen hour. But that sorting operation would be O(n)log(n) so just iterating over everything once is actually slightly faster. At the scale of data we're working at an O(n) iteration over data in memory is pretty trivial as far as computation time goes especially considering it will be a few years worth of data at most, that's just a few thousand hours.

rita_client/src/operator_update/mod.rs
jkilpatr: I've modified the "prepare_usage_data_for_upload" function to avoid sending the current hour. Last 5 minutes data use is somthing we need to add to this though good catch

ch-iara (approved):

PR #835, opened 2023-09-21
Paring wifi devices from /etc/config/wireless instead of /proc/net/wi…

ptulugu:
…reless

### DIFF
--- a/althea_kernel_interface/src/hardware_info.rs
+++ b/althea_kernel_interface/src/hardware_info.rs
@@ -17,6 +17,7 @@ use std::io::BufRead;
 use std::io::BufReader;
 use std::process::Command;
 use std::process::Stdio;
+use std::str::from_utf8;
 use std::time::Duration;
 use std::u64;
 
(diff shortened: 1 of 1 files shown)


althea_kernel_interface/src/hardware_info.rs
```diff
@@ -401,19 +402,33 @@ fn get_conntrack_info() -> Option<ConntrackInfo> {
...
-        if line.contains(':') {
-            let name: Vec<&str> = line.split(':').collect();
-            let name = name[0];
-            let name = name.replace(' ', "");
-            ret.push(name.to_string());
+    // We parse /etc/config/wireless which is an openwrt config. We return an error if not openwrt
+    if KI.is_openwrt() {
+        let mut ret = Vec::new();
+
+        let lines = KI.run_command("uci", &["show", "wireless"])?;
```
jkilpatr, line 411: why not keep and fall back to the old proc logic if it's not an openwrt device?

jkilpatr (approved):

PR #836, opened 2023-09-25
This commits adds a migration binary to moves clients on a postgresql…

ptulugu:
… db to a smart contract db

### DIFF
--- a/rita_bin/src/database_migration.rs
+++ b/rita_bin/src/database_migration.rs
@@ -1,0 +1,64 @@
+//! This binary is run to have access to a postgresql db, takes its client info and writes it into
+//! smart contract. This is necessary setup to move existing registered clients from the previous
+//! sql db format to a smart contract
+
+#![warn(clippy::all)]
+#![allow(clippy::pedantic)]
+#![forbid(unsafe_code)]
+
+use docopt::Docopt;
+use log::{error, info};
+use rita_client_registration::register_client_batch_loop::register_client_batch_loop;
+use rita_db_migration::start_db_migration;
+use serde::Deserialize;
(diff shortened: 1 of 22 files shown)


rita_bin/src/database_migration.rs
```diff
@@ -0,0 +1,56 @@
...
+    let private_key = args
+        .private_key
+        .parse()
+        .expect("Please provide a valid eth private key with funds");
+
+    // Start registration loop
+    register_client_batch_loop(web3_url, contract_addr, private_key);
+
+    match start_db_migration(db_url) {
+        Ok(_) => println!("Successfully migrated all clients!"),
```
ptulugu, line 42: This will return once it adds all clients to a datastore, a separate thread will pull from that datastore to register clients.
Will this program terminate that thread before it able to register users or will it stay running as long as the child threads are running?

rita_bin/src/database_migration.rs
```diff
@@ -0,0 +1,64 @@
+//! This binary is run to have access to a postgresql db, takes its client info and writes it into
+//! smart contract. This is necessary setup to move existing registered clients from the previous
+//! sql db format to a smart contract
+
+#![warn(clippy::all)]
+#![allow(clippy::pedantic)]
+#![forbid(unsafe_code)]
+
+use docopt::Docopt;
```
jkilpatr, line 9: this is nice 👍🏻

jkilpatr (approved):

jkilpatr:
This is a good way to implement everything here. I appreciate it, now we just have to migrate a test exist instance and see if that breaks.

PR #837, opened 2023-09-28
Add authentication and exit registration to exit db contract

jkilpatr:
This patch overhauls the exit db contract in a number of ways to prepare it for a production deployment.

1) The identity struct now stores mesh ip's as a uint128 and wireguard
   keys as a uint256 these are both superior storage methods to strings
   becuase valid ipv6 addresses and wireguard addresses are exactly
   these sizes. It's not otherwise possible to perform sufficient
   vlaidation of incoming data to ensure that any given string is a
   valid wireguard key or ipv6 address

2) Functions for adding and removing admins and permissions for user
   management in general. This is a required feature for a production
   implementation of this contract. Currently the 'super admin' capable
   of adding and removing individual admins is intended to be a Genosis
   Safe but could in theory simply be a secure single address

3) The capability to remove users has been added

4) The exit registry has been added as a parallel but otherwise
   identical data store to the user data store.

Currently this contract does not have any valid unit tests (also a requirement for production) and needs to be hooked into the router software.

### DIFF
--- a/solidity/contracts/AltheaDB.sol
+++ b/solidity/contracts/AltheaDB.sol
@@ -8,6 +8,16 @@ struct Identity {
     address eth_addr;
 }
 
+/// Identity struct for an exit, including a list of allowed region codes
+/// This is to protect a user from connecting to an exit that does not allow
+/// their region despite providing the best connection metrics otherwise
+struct ExitIdentity {
+    uint128 mesh_ip;
+    uint256 wg_key;
+    address eth_addr;
+    uint256[] allowed_regions;
+}
+
 /// Thrown when the caller is not authorized to register users
 error UnathorizedCaller();
 error DuplicateUser();
@@ -28,12 +38,12 @@ contract AltheaDB {
     // Mappings to regsitered clients
     Identity[] public state_registeredUsers;
     // Mappings to regsitered exits
-    Identity[] public state_registeredExits;
+    ExitIdentity[] public state_registeredExits;
 
     event UserRegisteredEvent(Identity indexed _user);
     event UserRemovedEvent(Identity indexed _user);
-    event ExitRegisteredEvent(Identity indexed _user);
-    event ExitRemovedEvent(Identity indexed _user);
+    event ExitRegisteredEvent(ExitIdentity indexed _user);
+    event ExitRemovedEvent(ExitIdentity indexed _user);
     event UserAdminAddedEvent(address indexed _admin);
     event UserAdminRemovedEvent(address indexed _admin);
     event ExitAdminAddedEvent(address indexed _admin);
@@ -45,8 +55,21 @@ contract AltheaDB {
 
     // start utility function 
 
+    // Used to convert an exit identity struct to an identity struct essentially just dropping the
+    // region codes component. This is mostly used for comparison and duplicate checking as we want
+    // to ignore the region codes when adding and removing an exit to avoid having identical exits
+    // with different region codes.
+    function exit_id_to_id(ExitIdentity memory input) public pure returns (Identity memory) {
+        return Identity({mesh_ip: input.mesh_ip, wg_key: input.wg_key, eth_addr: input.eth_addr});
+    }
+
     function get_null_identity() public pure returns (Identity memory) {
         return Identity({mesh_ip: 0, wg_key: 0, eth_addr: address(0)});
+    }
+
+    function get_null_exit_identity() public pure returns (ExitIdentity memory) {
+        uint256[] memory empty_array;
+        return ExitIdentity({mesh_ip: 0, wg_key: 0, eth_addr: address(0), allowed_regions: empty_array});
     }
 
     function is_null_identity(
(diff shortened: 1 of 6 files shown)


solidity/contracts/AltheaDB.sol
```diff
@@ -1,54 +1,369 @@
...
+
+            state_registeredExits.push(entry);
+            emit ExitRegisteredEvent(entry);
+        } else {
+            revert UnathorizedCaller();
+        }
     }
 
-    // Get all registered users
-    function get_all_registered_users()
```
ptulugu, line 29: This looks like this was removed and I dont see a replacement. Core requirement of rita exit is to get all users every tick, how will this be done?

solidity/contracts/AltheaDB.sol
```diff
@@ -1,54 +1,369 @@
...
 
+    // start user query functions
+
+
     function get_registered_client_with_wg_key(
-        string memory wg_key
+        uint256 wg_key
     ) public view returns (Identity memory) {
-        return wg_key_to_reg_users_map[wg_key];
+        for (uint256 i = 0; i < state_registeredUsers.length; i++) {
```
ptulugu, line 255: Why was the mapping(uint256 => Identity) removed and doing iteration instead?

solidity/contracts/AltheaDB.sol
```diff
@@ -1,54 +1,369 @@
...
+    address public immutable state_admin;
+    /// A list of addresses allowed to add and remove users from the list of users
+    address[] public state_UserAdmins;
+    /// A list of addresses allowed to add and remove exits from the list of exits
+    address[] public state_ExitAdmins;
+
+    // Mappings to regsitered clients
+    Identity[] public state_registeredUsers;
+    // Mappings to regsitered exits
+    Identity[] public state_registeredExits;
```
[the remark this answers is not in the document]
jkilpatr: The purpose of this is to replace the cluster members list in the config of current exits. Right now exits need to have all cluster members configured manually. Since our goal is plug and go exits with almost no configuration making exit operators put in a list of cluster members goes against that goal.

solidity/contracts/AltheaDB.sol
jkilpatr: Note that I changed the users lists from private to public, public variables have automatic getters meaning they are simply accessible by calling the variable name as a function. Give it a try.

In this case 'public' means 'readable by other code' but does not allow anyone else to modify the data.

solidity/contracts/AltheaDB.sol
jkilpatr: EVM code optimization requires a bit of background.

1) storage is very expensive, in fact it's a lot more expensive than computation  in the EVM the cost of every operation is strictly defined for gas purposes 

https://ethereum.org/en/developers/docs/evm/opcodes/

As you can see here iteration, addition and comparison operations are in the low single digit like 3,5,8 gas. But a store to a new data slot can cost up to a whopping 20,000 gas. 

So simply put the iteration is cheaper (it's also cheaper to store the uint256 rather than strings, because, strings have a length value that takes up a storage slot of it's own, as well as unicode encoding being less efficient)

2) the mapping is just more data to handle when doing deletion.

ptulugu: Thats pretty cool, thanks for the clarificaiton

solidity/contracts/AltheaDB.sol
```diff
@@ -1,54 +1,417 @@
...
+/// Identity struct for an exit, including a list of allowed region codes
+/// This is to protect a user from connecting to an exit that does not allow
+/// their region despite providing the best connection metrics otherwise
+/// The payment types specification prevents clients from moving to exits
+/// that do not accept their tokens as payment
+struct ExitIdentity {
+    uint128 mesh_ip;
+    uint256 wg_key;
+    address eth_addr;
+    uint256[] allowed_regions;
```
ChristianBorst, line 20: Unless you're going to use strings for region values, uint256 seems a bit excessive

solidity/contracts/AltheaDB.sol
```diff
@@ -1,54 +1,417 @@
...
+    uint256[] payment_types;
```
ChristianBorst, line 21: Shouldn't payment types be of type `address[]`? Or is this going to be using strings too?

jkilpatr: I'm thinking of these as just codes that are externally agreed upon to reference different payment setups. Rather than trying to embed all possible settings.

solidity/contracts/AltheaDB.sol
jkilpatr: it is excessive, we could probably do u32 but I don't see how extra options hurts us here.

solidity/test/altheaDBTests.ts
```diff
@@ -0,0 +1,280 @@
...
+  let user2 = {
+    mesh_ip: "0xfd001329",
+    wg_key: "0xAFEDD",
+    eth_addr: "0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD",
+  };
+  let partialDup = {
+    mesh_ip: "0xfd001338",
+    wg_key: "0xAFEDB",
+    eth_addr: "0x154CB202089D58efB56a2B11ce812Ae3882fE1f3",
+  };
```
ChristianBorst, line 41: It would be nice to have partial dup detection on more than just the wg_key

ChristianBorst (approved):
This looks good to me, I just think a couple of the Identity fields are oversized + one test suggestion

PR #840, opened 2023-10-10
Pranay/refactor

ptulugu:
- Update abi parsing for updated identity
- update exit_list endpoint to account for region and payment codes when returning an exit list for backward
  compatibility and add a new endpoint for future clients to use
- add abi parsing for ExitIdentity
- Add 5x Gas limit mulitplier for registration server loop
- Update db client migration to run with a state admin contract key
- Fix CI

Client side updates: 
 This contains updates to the client exit manager to work with the new exit format
 
i.e. A vector of individual exits as potential exits to an exit can connect to, rather than
subnetting or clustering of exits that was done in the past. This allows exits to have only one
identity that the client cares about, which updates gateway billing to use only one exit id in
debt keeper

The Exit list now returns a list of exit id with their respective allowed regions to connect to and
payments accepts, the client (which has these fields set up in its config) will now filter through all
exits it receives, put the ones it could potetially connect to in the ExitServer list. This allows us to
query for general details of these exits. If we are already registered to one exit, we are registered to all
exits, so these new exits move from New -> Registered, which then allows us to switch to them easily when an exit goes down


althea_types/src/interop.rs
```diff
@@ -119,6 +119,52 @@ pub struct Denom {
...
+            1 => Regions::US,
+            2 => Regions::Canada,
+            3 => Regions::PuertoRico,
+            4 => Regions::Mexico,
+            5 => Regions::UK,
+            6 => Regions::SouthAmerica,
+            7 => Regions::Africa,
+            8 => Regions::Australia,
+            // Undefined region, return US by default
+            _ => Regions::US,
```
jkilpatr, line 163: this should be an 'unknown region'

althea_types/src/interop.rs
```diff
@@ -119,6 +119,52 @@ pub struct Denom {
     pub decimal: u64,
 }
 
+/// An enum representation of the Regions supported by althea exits
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub enum Regions {
+    US,
+    Canada,
+    PuertoRico,
```
jkilpatr, line 127: PuertoRico can be lumped in with the US for streaming purposes as far as I know. Include Ghana, Nigeria, Columbia (we have or have had networks there)  Mexico etc then drop the list. Don't assign ID's to very large vague continents as we'll just end up having to redefine them to be country specific for the sake of streaming. All of this is about streaming licensing regions

althea_types/src/interop.rs
```diff
@@ -128,6 +174,31 @@ pub enum SystemChain {
...
+    }
+}
+
+impl From<u8> for SystemChain {
+    fn from(value: u8) -> Self {
+        match value {
+            1 => SystemChain::Althea,
+            2 => SystemChain::Ethereum,
+            3 => SystemChain::Rinkeby,
+            4 => SystemChain::Xdai,
```
jkilpatr, line 195: 👍🏻

althea_types/src/wg_key.rs
```diff
@@ -148,4 +161,12 @@ mod tests {
     fn test_wgkey_panic_from_short_string() {
         WgKey::from_str("ABCDE").unwrap();
     }
+
+    #[test]
+    fn test_wgkey_to_from_uint256() {
+        let key = WgKey::from_str("8BeCExnthLe5ou0EYec5jNqJ/PduZ1x2o7lpXJOpgXk=").unwrap();
+        let uint: Uint256 = key.into();
```
jkilpatr, line 168: 👍🏻

integration_tests/src/utils.rs
```diff
@@ -1104,3 +1098,68 @@ pub async fn populate_routers_eth(rita_identities: InstanceData) {
...
+    web3.wait_for_transaction(res, TX_TIMEOUT, None)
+        .await
+        .unwrap();
+
+    let cluster = TEST_EXIT_DETAILS.get("test").unwrap();
+    let nonce = web3.eth_get_transaction_count(miner_pub_key).await.unwrap();
+
+    for (i, id) in rita_identities.exit_identities.iter().enumerate() {
+        let mut mesh_v2 = id.mesh_ip;
+        for exit in cluster.instances.values() {
```
jkilpatr, line 1127: Is this trying to reconcile the exit id's in the contract with the configured exit cluster data? We should remove exit cluster data from the exit config and have the exit look at the contract for that.

rita_bin/src/database_migration.rs
```diff
@@ -38,6 +38,13 @@ fn main() {
 
     let system = actix_async::System::new();
 
+    system.block_on(db_migration_user_admin(
+        web3_url.clone(),
+        private_key,
+        private_key.to_address(),
```
jkilpatr, line 44: using actix async main (examples in the other bins) you can avoid the block on.

rita_client_registration/src/client_db.rs
```diff
@@ -516,86 +690,143 @@ fn test_parse_abi_array() {
...
+    let bytes = "\
+    0000000000000000000000000000000000000000000000000000000000000020\
+    0000000000000000000000000000000000000000000000000000000000000002\
+    0000000000000000000000000000000000000000000000000000000000000040\
+    0000000000000000000000000000000000000000000000000000000000000180\
+    000000000000000000000000000000009de79d5506c2d55aba50edaee3b9579f\
+    9de79d5506c2d55aba50edaee3b9579f75acd14a63c551c5595bbf66d074a379\
+    000000000000000000000000dcc1137e069cab580fd4199ac682a81978e39bc5\
+    00000000000000000000000000000000000000000000000000000000000000a0\
+    0000000000000000000000000000000000000000000000000000000000000100\
```
jkilpatr, line 813: 💯  nice tests and good job in general with the parsing

rita_db_migration/src/lib.rs
```diff
@@ -94,3 +100,25 @@ pub fn get_database_connection(
...
+    state_admin_key: PrivateKey,
+    user_to_add: Address,
+    db_addr: Address,
+) {
+    let contact = Web3::new(&web3_url, WEB3_TIMEOUT);
+
+    // We need to be a user admin to add users
+    let res = add_user_admin(
+        &contact,
+        db_addr,
```
jkilpatr, line 115: we should check if we are already a user admin before adding ourselves. As this will error if we're already a user admin.

rita_exit/src/rita_loop/mod.rs
```diff
@@ -358,6 +358,7 @@ pub fn start_rita_exit_endpoints(workers: usize) {
                     .route("/client_debt", web::post().to(get_client_debt))
                     .route("/time", web::get().to(get_exit_timestamp_http))
                     .route("/exit_list", web::post().to(get_exit_list))
+                    .route("/exit_list", web::get().to(get_exit_list_v2))
```
jkilpatr, line 361: please use exit_list_v2 as the endpoint string not a post vs get endpoint.

jkilpatr (changes requested):
Nice! Just a few refinements and this will be ready to merge. The parsing changes are nice and well tested.

rita_exit/src/rita_loop/mod.rs
ptulugu: done

rita_bin/src/database_migration.rs
ptulugu: done

integration_tests/src/utils.rs
ptulugu: This setups the correct exit identities in the contract with mesh_ip_v2 instead of mesh_ip (which is returned by spawn_rita). I removed the exit cluster list from exit settings. There are some naming issues in the test since I hardcoded orginal exits with a cluster structure, i was planning to move it to just a vector of exits in my next PR

althea_types/src/interop.rs
ptulugu: done

althea_types/src/interop.rs
ptulugu: Done

rita_db_migration/src/lib.rs
ptulugu: Fixed

integration_tests/src/utils.rs
jkilpatr: @ptulugu  why not update spawn_rita to return mesh_ip_v2? or both ip's go ahead and move to the vector of exits in this pr.

althea_types/src/interop.rs
```diff
@@ -344,6 +364,14 @@ pub struct ExitList {
     pub wg_exit_listen_port: u16,
 }
 
+/// Struct returned when hitting exit_list_V2 endpoint
+#[derive(Default, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub struct ExitListV2 {
+    pub exit_list: Vec<ExitIdentity>,
+    // All exits in a cluster listen on same port
+    pub wg_exit_listen_port: u16,
```
jkilpatr, line 372: maybe make this per exit identity? Just to avoid forcing this assumption.

ptulugu: added it to exitidentity

integration_tests/src/contract_test.rs
```diff
@@ -106,63 +112,53 @@ pub async fn validate_contract_exit_functionality(db_addr: Address) {
...
         db_addr,
         miner_private_key,
-        None,
+        Some(TX_TIMEOUT),
         vec![],
     )
     .await
     .unwrap();
-    contact
-        .wait_for_transaction(res, TX_TIMEOUT, None)
```
jkilpatr: why remove this?

settings/src/client.rs
```diff
@@ -40,6 +40,11 @@ pub struct ExitServer {
     #[serde(default)]
     pub registration_port: u16,
 
+    /// The power we reach out to to hit the register endpoint
+    /// also used for all other exit lifecycle management api calls
+    #[serde(default)]
+    pub wg_exit_listen_port: u16,
```
jkilpatr, line 46: this is going to default to zero, please use the default function feature to specify the correct default value.

ptulugu: added 4875 default registration port and 59998 default wg exit listen port

integration_tests/src/contract_test.rs
ptulugu: The function above this has a parameter for timeout, that I added Some(TX_TIMEOUT) which does the waiting before returning. I added that parameter and removed the wait_for_transaction to remove redundant code

PR #844, opened 2023-10-27
Fixes to db migration binary

ptulugu:
- Limit registration loop batch size to 100 and implement prepare_transaction manually to speed up batch registration
- Modify duplicate user check in solidity contract to use mapping instead of iteration to reduce gas usage when there are >1000+ clients

### DIFF
--- a/integration_tests/src/db_migration_test.rs
+++ b/integration_tests/src/db_migration_test.rs
@@ -61,7 +61,14 @@ pub async fn run_db_migration_test() {
     register_client_batch_loop(get_eth_node(), althea_db_addr, miner_private_key);
 
     info!("Running user migration");
-    match start_db_migration(DB_URI.to_string()) {
+    match start_db_migration(
+        DB_URI.to_string(),
+        get_eth_node(),
+        miner_private_key.to_address(),
+        althea_db_addr,
+    )
+    .await
+    {
         Ok(_) => println!("Successfully migrated all clients!"),
         Err(e) => println!("Failed to migrate clients with {}", e),
     }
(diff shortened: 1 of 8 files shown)


integration_tests/src/db_migration_test.rs
```diff
@@ -41,7 +41,7 @@ pub async fn run_db_migration_test() {
     // Add a bunch of dummy clients to the db to migrate
     add_dummy_clients_to_db(num_clients, &conn);
 
-    thread::sleep(Duration::from_secs(10));
+    thread::sleep(Duration::from_secs(100000));
```
jkilpatr, line 44: is this intentional? that's a long wait?

solidity/contracts/AltheaDB.sol
```diff
@@ -299,6 +269,11 @@ contract AltheaDB {
         if (is_user_admin(msg.sender)) {
             uint256 index = get_index_of_id(entry, state_registeredUsers);
             delete_array_entry(index, state_registeredUsers);
+
+            state_registeredAddr[entry.eth_addr] = false;
+            state_registeredIps[entry.mesh_ip] = false;
+            state_registeredKey[entry.wg_key] = false;
```
jkilpatr, line 275: Ok so there are 3 states now, true, false, and null? So when we try to look up a user who's not in this list is successfully resolves as false? Clever strategy. Couldn't we free the storage here when we delete the user?

integration_tests/src/db_migration_test.rs
ptulugu: whoops, forgot to change this

jkilpatr (approved):

PR #848, opened 2023-11-27
Add bulk registration to AltheaDB

jkilpatr:
This pr adds bulk registration as an endpoint to the AltheaDB.sol contract. This is much easier to manage than sending thousands of transactions to register and migrate users.


ChristianBorst (approved):
The solidity looks good to me, good to know that the camel case is so important.

PR #852, opened 2024-01-12
Babeld config at startup

jkilpatr:
This commit provides functionality to configure Babeld at Rita startup with default interface settings. This allows us to control via Rita how exactly babel is behaving and provides the functions for potnetially tweaking these functions at runtime in the future.


babel_monitor/src/lib.rs
```diff
@@ -196,31 +196,96 @@ pub fn parse_interfaces(stream: &mut TcpStream) -> Result<Vec<Interface>, BabelM
...
+pub fn set_interface_defaults(
+    stream: &mut TcpStream,
+    defaults: BabeldInterfaceConfig,
+) -> Result<(), BabelMonitorError> {
+    let mut command = "default ".to_string();
+    command.push_str(&build_interface_config_string(defaults));
+    let result = run_command(stream, &command)?;
+
+    let _out = result;
+    Ok(())
```
ch-iara, line 253: Do we not want to log these results (L239 as well)

rita_bin/src/exit.rs
```diff
@@ -90,6 +91,10 @@ fn main() {
         println!("Look the exit settings! {settings:?}");
         settings
     };
+    apply_babeld_settings_defaults(
```
ch-iara, line 94: Do the exit settings files not also need to be migrated the way you are doing in client.rs?

rita_client/src/operator_update/mod.rs
```diff
@@ -481,13 +487,12 @@ fn update_payment_settings(
...
         if payment.system_chain != new_chain {
-            set_system_blockchain(new_chain, &mut payment);
+            set_system_blockchain(new_chain, payment);
         }
     }
     if let Some(new_chain) = new_settings.withdraw_chain {
         payment.withdraw_chain = new_chain;
     }
-    payment
 }
```
ch-iara, line 496: Since this is no longer returning anything in this fn does this check for withdraw_chain need to be handled and set somewhere or can this be removed

babel_monitor/src/lib.rs
jkilpatr: probably a good call will do.

rita_bin/src/exit.rs
jkilpatr: exits are at this point all hand configured, so not really as I'll update the exit installer, once we have appliance exits and an update procedure for those then yes we would need migration logic, so good catch.

rita_client/src/operator_update/mod.rs
jkilpatr: payment is a mutable reference so the check is in fact changing the variable passed into the function

this is why mutable references are generally a bad idea, people lose track of when they are being modified. So maybe I should change this back to a return based structure.

babel_monitor/src/lib.rs
jkilpatr: oh actually the response in this case is 'ok' or 'bad', which us handling is probably a good idea...

PR #897, opened 2024-05-07
CI test stability improvements

jkilpatr:
The goal of this pr is to cleanup and improve the performance and reliability of the test environment which is proving to be a barrier to faster development.

Before
image

After 
image


althea_kernel_interface/src/netns.rs
```diff
@@ -7,7 +8,13 @@ impl dyn KernelInterface {
     /// that the lazy static for cross thread comms arch if a bit questionable by nature
     pub fn check_integration_test_netns(&self) -> u32 {
         if cfg!(feature = "integration_test") {
-            let ns = self.run_command("ip", &["netns", "identify"]).unwrap();
+            let mut ns = self.run_command("ip", &["netns", "identify"]);
+            while let Err(e) = ns {
+                warn!("Could not get netns name, retrying: {:?}", e);
+                sleep(std::time::Duration::from_secs(1));
```
ch-iara, line 14: any chance this never exits the loop and stalls forever? are we keeping a time limit somewhere still

ch-iara: found the timeout change nevermind

ch-iara (approved):

PR #919, opened 2024-05-27
Include Beta 21 rc13 fixes

jkilpatr:
See the individual commit messages for more details. This is catching up on all the features from the beta21 branches in beta21rc9 -> beta21rc13


ch-iara (approved):
Looks cohesive to me, self rescue into its own module is helpful to have for ease of understanding. I'm not super familiar with what's going on with the payment controller and debt keeper mods but the updates to validation error handling look useful and the test scenario changes will be good to have

PR #942, opened 2024-08-06
Chiara/websockets to ops

ch-iara:


rita_client/src/heartbeat/mod.rs
```diff
@@ -128,7 +128,7 @@ fn send_udp_heartbeat() {
         heartbeat_url = "7.7.7.7:33333";
         info!("We are using dev env heartbeart url");
     } else if cfg!(feature = "operator_debug") {
-        heartbeat_url = "192.168.10.2:33333";
+        heartbeat_url = "192.168.1.214:33333";
```
jkilpatr, line 131: just flagging that this shouldn't be checked in in the final pr.

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -0,0 +1,118 @@
...
+        let runner = System::new();
+        runner.block_on(async move {
+            let res = awc::Client::new().ws(url).connect().await;
+            match res {
+                Ok((res, mut ws)) => {
+                    info!("Websocket actor is connected {:?}", res);
+                    let mut ops_last_seen_usage_hour: Option<u64> = None;
+                    loop {
+                            // check if there is anything to read first
+                            let res = timeout(Duration::from_millis(10), ws.next()).await;
```
jkilpatr, line 42: this loop should be split out into its own function so it's easier to separate the logic form all the threads/runners

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -0,0 +1,118 @@
...
+                            if let Ok(Some(res)) = res {
+                                let msg = res.unwrap();
+                                ops_last_seen_usage_hour = handle_received_operator_message(msg);
+                            }
+
+                            // then send over new checkin data
+                            let update_data = get_operator_checkin_data(ops_last_seen_usage_hour);
+                            if update_data.is_err() {
+                                error!("Failed to get operator checkin data, cannot check in with operator server");
+                                break;
```
jkilpatr, line 52: why exactly do we want to break in this case?

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -0,0 +1,118 @@
...
+                            }
+                            let serialized = serde_json::to_vec(&update_data.unwrap()).unwrap();
+                            info!("Sending checkin through websocket!");
+                            let did_send = ws.send(ws::Message::Binary(serialized.into())).await;
+                            match did_send {
+                            Ok(_) => {
+                                // check again for any responses to read
+                                let res = timeout(Duration::from_millis(10), ws.next()).await;
```
jkilpatr, line 60: this timeout should be a constant somewhere and should be much lower like 100ms, we're not really waiting for a response from upstream we're checking if one has already arrived.

jkilpatr (changes requested):
This is a great structure overall and only needs some minor cleanup before it's ready to merge.

rita_client/src/operator_update/mod.rs
```diff
@@ -52,38 +61,61 @@ lazy_static! {
...
     }
-
+    neighbor_info
+}
+pub fn get_contact_info() -> Option<ContactType> {
+    let rita_client = settings::get_rita_client();
+    let exit_client = rita_client.exit_client;
+    option_convert(exit_client.contact_info)
+}
+pub fn get_install_details() -> Option<InstallationDetails> {
```
jkilpatr, line 106: these accessory functions would do better in the settings module, also atomic setters would be nice.

ch-iara: I've put these ones that are just from settings in the settings module, but some of these getters are not easily transferable (see get_hardware_info_update) or aren't really settings (the usage/throughput getters and rita uptime) so I've left them here

althea_types/src/websockets.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Fully reboots the router, this includes a power cycle not just a restart of the
+    /// routing processes. For x86 machines this action comes with some risk as devices may
+    /// get stuck in the BIOS if not configured properly.
+    Reboot,
+    /// Restart babeld and rita on the router
+    SoftReboot,
+    /// Sends instructions from op tools about the type of update to perform, either a sysupgrade
+    /// or an opkg update, to be removed after all routers >= beta 19 rc9
+    Update {
+        instruction: UpdateTypeLegacy,
```
jkilpatr, line 109: shouldn't we use the latest update type not the legacy here?

jkilpatr (approved):
I really like the comments and architecture here, very clean and everything I hoped for when speccing this out. 

Just one comment on what type we use. But we can possibly skip that.

althea_types/src/websockets.rs
ch-iara: we do, with UpdateV2- this is left in to maintain ops compatibility since I've just moved this struct over from the other file

jkilpatr:
This is really great work, excited to merge this into the beta22 branch for a firmware build.

PR #988, opened 2024-09-09
Update exit list signing infra for contract based exits

jkilpatr:
This pr can be split into modernization/update work and then exit list functionality changes.

The goal is that routers have a list of keys which can be used as the root of trust for the exit list. An exit list server exists online, publicly queryable, and returns a signed list of exits as they exit in the exit registration smart contract.

This allows exits to simply query and pass along this list to clients at will, the clients can then verify the list against their stored root of trust, and bootstrap using these exits. This resolves the need to trust a local exit to bootstrap from it and simplifies the previous structures in which the exits needed other exits configured in which to serve their own lists.


ch-iara:
Some questions for the signing loop: 
Is the idea to sign all cached values, then store them in a second cache to load? Otherwise I'm not seeing the point there when to_encrypted_server_list can be run when a client requests a signed list (and from my understanding these are one and the same caches we're accessing between the signing loop and the endpoint- but perhaps that's not true since as it stands there is a cache sent in to the endpoint as an argument?)

And then, how do I retrieve data for a contract that is not already in the cache?

exit_trust_root/src/bin.rs
```diff
@@ -25,58 +33,132 @@ pub const DOMAIN: &str = if cfg!(test) || DEVELOPMENT {
 /// The backend RPC port for the info server fucntions implemented in this repo
 const SERVER_PORT: u16 = 9000;
 
+lazy_static! {
+    static ref EXIT_CONTRACT_CACHE: Arc<RwLock<HashMap<Address, ExitContractSignatureCacheValue>>> =
+        Arc::new(RwLock::new(HashMap::new()));
```
jkilpatr, line 38: you can use actix_web db data for this (which won't run into scoping problems in the test environment) https://docs.rs/actix-web/latest/actix_web/web/struct.Data.html

exit_trust_root/src/bin.rs
```diff
@@ -25,58 +33,132 @@ pub const DOMAIN: &str = if cfg!(test) || DEVELOPMENT {
...
+        // we already have a signed list- now to encrypt it given the nonce & our... keys...
+        let plaintext = serde_json::to_string(&self.exit_list.data)
+            .expect("Failed to serialize ExitServerList")
+            .into_bytes();
+        // using the clarity private key as the Crypto_box SecretKey
+        let our_secretkey = SecretKey::from(CONFIG.clarity_private_key.to_bytes());
+        let b = SalsaBox::new(&their_pubkey, &our_secretkey);
+        let ciphertext = b.encrypt((&self.nonce).into(), plaintext.as_ref()).unwrap();
+        EncryptedExitServerList {
+            pubkey: CONFIG.wg_private_key,
```
ch-iara, line 123: Flagging this should be pubkey not priv

althea_types/src/exits/mod.rs
```diff
@@ -101,33 +102,11 @@ pub struct EncryptedExitState {
...
-
-impl ExitListV2 {
-    pub fn into_identities(self) -> Vec<Identity> {
-        self.exit_list.into_iter().map(|exit| exit.into()).collect()
-    }
-}
+// impl ExitList {
+//     pub fn into_identities(self) -> Vec<Identity> {
+//         self.exit_list.into_iter().map(|exit| exit.into()).collect()
+//     }
```
jkilpatr, line 108: this can just be removed I think.

exit_trust_root/config.toml
```diff
@@ -0,0 +1,5 @@
+[clarity_private_key]
+'0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f1e'
```
jkilpatr, line 2: can we remove checked in config files if possible?

exit_trust_root/src/bin.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+async fn retrieve_exit_server_list(
+    exit_contract: Address,
+    cache: Arc<RwLock<HashMap<Address, SignedExitServerList>>>,
+) -> Result<SignedExitServerList, Web3Error> {
+    const WEB3_TIMEOUT: Duration = Duration::from_secs(10);
+    let exits = get_exits_list(
+        &Web3::new("https://dai.althea.net", WEB3_TIMEOUT),
```
jkilpatr, line 69: this url should be a const at the top of the program main

rita_client/src/exit_manager/requests.rs
```diff
@@ -271,10 +268,23 @@ pub async fn get_exit_list(exit: IpAddr) -> Result<ExitListV2, RitaClientError>
...
 
-    match decrypt_exit_list(&our_privkey.into(), value, &exit_pubkey.into()) {
-        Err(e) => Err(e.into()),
-        Ok(a) => Ok(a),
+    //verify the signature on the list
+    let key = list.data.contract;
+    let sig = list.clone().signature;
+    match list.data.verify(key, sig) {
+        true => Ok(list),
+        false => Err(RitaClientError::MiscStringError(
```
jkilpatr, line 286: where it's feasible try to make a new error on RitaClientError rather than just using miscstring error it's a catch all that ideally we use sparingly and eventually eliminate. Having a specific error type will make it easier for callers to match on and handle the signature failed error if they ever want to.

rita_exit/src/network_endpoints/mod.rs
```diff
@@ -191,119 +188,49 @@ pub async fn get_exit_timestamp_http(_req: HttpRequest) -> HttpResponse {
...
-pub async fn get_exit_list(request: Json<EncryptedExitClientIdentity>) -> HttpResponse {
-    let exit_settings = get_rita_exit();
-    let our_secretkey: WgKey = exit_settings.network.wg_private_key.unwrap();
-    let our_secretkey = our_secretkey.into();
-
-    let their_nacl_pubkey = request.pubkey.into();
-
-    let contact = Web3::new(&get_web3_server(), CLIENT_STATUS_TIMEOUT);
+/// This exit may not be included in the list returned by the smart contract! If this is the case it must
+/// first be added to the root of trust server, and clients will know to choose a different exit.
```
jkilpatr, line 192: This comment needs some more detail to avoid confusion for future readers IMO.

Because exits are running their get_exit_list endpoint on a webserver with a multihomed ip within the babel mesh any exit that is plugged in and online will be getting exit server list requests, regardless of if that exit is on the list or otherwise. This is secure due to the signatures of course, but it's important to realize that clients could talk to any one of an arbitrary number of exits on the network for this list and then find only a very small subset, or even none of those exits are permitted roaming targets once it verifies the list results.

rita_exit/src/network_endpoints/mod.rs
```diff
@@ -191,119 +188,49 @@ pub async fn get_exit_timestamp_http(_req: HttpRequest) -> HttpResponse {
...
-                        && !exit.payment_types.is_disjoint(&accepted_payments)
-                    {
-                        ret.push(exit_identity_to_id(exit))
-                    }
-                }
-                ret.push(our_id); // add ourselves to the list
-                ret
+    // we are receiving a SignedExitServerList from the root server.
+    let signed_list: SignedExitServerList =
+        match get_exit_list_from_root(contract_addr).await {
```
jkilpatr, line 199: this should have a cache just like the root of trust server does. That way if the exit experiences a momentary blip in connectivity, or say the exit root of trust server goes down for a few seconds, or even hours, while maint is happening it won't disrupt normal operation.

PR #994, opened 2024-09-18
Add Telnyx data object to expected return type

ch-iara:


jkilpatr:
I amended this into my commit in order to keep the fix clean. It is merged.

PR #997, opened 2024-09-22
Refactor kernel interface to remove mock and global var

jkilpatr:
When a unit testing scheme was originally designed for the kernel interface module in 2017 it was turned into a stateful struct to accomodate a mock test environment that allowed the tests to call the real functions and specify what exact commands they expected to be run.

This structure has not been maintained, new tests are not written in that style and instead use separate testing functions. But the original tests have not been modified much since that time. Leaving Rita with a KI global variable that in production doesn't actually do anything all to support mocking during tests.

This commit removes the mock structure and turns kernel interface into just a grab bag of utility functions where tests functions are all split out internally rather than relying on mock.

While we are touching all of these files it will also be renamed to system_interface rather than kernel interface as we now interface with a lot more than just linux kernel functions from this module.


ch-iara (approved):
Looks good to me, this will make it much less of a pain to run commands without having to do a little dance to grab KI

PR #1000, opened 2024-09-25
Encryption for websocket messages with crypto_box

ch-iara:


althea_types/src/websockets/encryption.rs
```diff
@@ -0,0 +1,221 @@
...
+//! sent to ops tools from a router)
+
+use std::fmt::Display;
+
+use crypto_box::{
+    aead::{Aead, AeadCore},
+    PublicKey, SalsaBox, SecretKey,
+};
+use rand::rngs::OsRng;
+
```
jkilpatr, line 11: if you remove the spaces between imports cargofmt will sort them. very minor nit.

althea_types/src/websockets/encryption.rs
```diff
@@ -0,0 +1,221 @@
...
+    let b = SalsaBox::new(&their_nacl_pubkey, our_secretkey);
+    let decrypted_bytes = match b.decrypt(their_nonce.as_ref().into(), ciphertext.as_ref()) {
+        Ok(value) => value,
+        Err(_) => {
+            return Err(WebsocketEncryptionError::DecryptionError {
+                e: "Could not decrypt websocket message".to_string(),
+            })
+        }
+    };
+    let decrypted_string = match String::from_utf8(decrypted_bytes) {
```
jkilpatr, line 73: I like the error handling here, clean readable and avoids nesting.

althea_types/src/websockets/encryption.rs
```diff
@@ -0,0 +1,221 @@
...
+    EncryptedOpsWebsocketMessage {
+        nonce: nonce.into(),
+        encrypted_ops_websocket_msg: ciphertext,
+        pubkey: our_publickey,
+    }
+}
+
+pub fn decrypt_ops_websocket_msg(
+    our_secretkey: &SecretKey,
+    ops_publickey: &PublicKey,
```
jkilpatr, line 115: implementing these encrypt and decrypt as member functions of the struct might be useful so you can go

encryptedopsmsg.decrypt(args)

opsmsg.encrypt(args)

You don't need to move the functions around to do so, just wrap them in an impl Optoolsmessage or impl EncryptedOpsMessage for example.

althea_types/src/websockets/mod.rs
```diff
@@ -76,6 +78,85 @@ pub enum RouterWebsocketMessage {
     },
 }
 
+// just ignore the hardware info field on the timeseries data message for now
```
jkilpatr, line 81: was this required just because hardware_info doesn't implement partialeq? What issue did you run into trying to add that over in kernel interface?

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -229,14 +289,19 @@ fn get_five_minute_update_data(
         client_mbps,
         relay_mbps,
     };
-    let serialized = serde_json::to_vec(&data).unwrap();
+    let encrypted = encrypt_router_websocket_msg(id.wg_public_key, our_secretkey, ops_pubkey, data);
```
jkilpatr, line 292: you can have the encrypted messages implement a .json() call that does the serde_json::to_vec then you could just call encrypt_router_websocket_msg(args).json()

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -79,8 +113,13 @@ pub fn send_websocket_update() {
...
                                         // and reconnecting the socket
                                         let msg = msg.unwrap();
-                                        if let Some(hour) = handle_received_operator_message(msg) {
-                                            ops_last_seen_usage_hour = Some(hour);
+                                        if let Some(hour) = handle_received_operator_message(msg, &our_secretkey, Some(&ops_pubkey)) {
+                                            match hour {
+                                                ReceivedOpsData::UsageHour(hour) => {
+                                                    ops_last_seen_usage_hour = Some(hour);
+                                                },
+                                                _ => panic!("Why are we receiving a public key again from ops? restarting!"),
```
jkilpatr, line 121: we should probably handle a repeat sending gracefully, if ops gets restarted it will send it again right?

rita_client/src/operator_update/ops_websocket.rs
```diff
@@ -67,6 +73,34 @@ pub fn send_websocket_update() {
...
+                                // we must receive the ops pubkey before we can proceed with encryption
+                                loop {
+                                    // check if we have received the ops pubkey
+                                    if let Ok(Some(msg)) =
+                                        timeout(SOCKET_CHECKER_TIMEOUT, ws.next()).await
+                                    {
+                                        let msg = msg.unwrap();
+                                        if let Some(hour) = handle_received_operator_message(msg, &our_secretkey, None) {
+                                            match hour {
+                                                ReceivedOpsData::WgKey(public_key) => {
```
jkilpatr, line 93: this is a good structure, only two comments

1) this key should also be used over in the heartbeat code, lazy static or setting it / getting it from the config would be appropriate
2) this should happen any time we get the message not just startup, just in case odd things happen. I do see the appeal of just crashing and trying again but I'm worried if we have some odd ordering bug we'll have a crash loop here go on forever.

jkilpatr (changes requested):
This looks like it has all the functionality we need.

I've left a few comments on how to streamline the syntax a little bit and also two pieces of behavior I'd like to see changed. One the key ops sends should also be used for heartbeats and two that we can handle updated keys being sent at odd times. It's just good practice to always handle any message ops throws at us gracefully rather than panicing.

althea_types/src/websockets/mod.rs
ch-iara: yes, exactly- and when adding partialeq to hardware info, a bunch more nested struct members needed partialeq impl as well and I wasn't sure we really needed all that. I can add it in if we think it's important, but the manual impl here was only added because I'm trying to use an assert_eq in the unit test. maybe instead I can just manually compare values in the unit test instead....

rita_client/src/operator_update/ops_websocket.rs
ch-iara: Yes, the idea here is that if ops restarts (which will cut off the websocket connection regardless) we'll get a new key, but the only time we send keys on the ops side are when a connection is first opened. on the router side if this thread panics we restart that loop and get a new connection

rita_client/src/operator_update/ops_websocket.rs
ch-iara: HEARTBEAT_SERVER_KEY now comes from ops over websocket- so if the router cannot connect via socket first it cannot use the dummy key to heartbeat

rita_client/src/operator_update/ops_websocket.rs
ch-iara: Edited so now we gracefully change the ops server key

althea_types/src/websockets/mod.rs
jkilpatr: chasing down all the nested structs with partialeq is probably worth the trouble, but this is a good solution, manually comparing values wouldn't be better because that's only useful in this test versus being useful in any future case where we may want to compare these structs.

jkilpatr (approved):

PR #1032, opened 2024-10-31
Fix missing wifi device data

ch-iara:
Getting wifi iface names from 'uci show wireless' is no longer an option since these do not always get saved to config- and routers that were missing hard coded ifname config were not populating wifi devices to report to ops. Instead, 'iw dev' lists all interfaces for wifi and those ifnames can be more reliably accessed from there.


jkilpatr:
Can you add a unit test for the parsing logic? Just break out the parsing from the calling of the command.

PR #1079, opened 2025-01-08
Fix wifi data gathering

ch-iara:
This fix adds on to 58c0626 which fixed the missing wifi device data with the unintended side effect of breaking the ability to set wifi settings from ops. The original problem was that for routers missing the ifname setting in /etc/config/wireless, the wifi interface names differ between uci and iw, the two places we gather different wifi info to be sent over in the WifiDevice field of HardwareInfo. The first fix used iw instead to gather interface names, but since ifnames may differ between the two systems if routers are missing the config line, this broke wifi settings when routers were searching uci for an interface name only used by iw.

This change calls both uci and iw so no matter whether the router has its ifname nicknamed it can gather the required data to be sent to ops for each interface.

---------
Tested on LR-24 which immediately sent through both wifi device info and the previously broken wifi settings


ch-iara:
refactored in the KI lazy static removal commit- but why now is debt keeper testing failing? And why do rust tests fail on this branch and not beta22 even without the KI change? Not sure what exactly changed between the two that is causing this but for now the test failures on this branch only are what's stopping merge of beta21

PR #1083, opened 2025-01-15
SNAT routing 

ch-iara:
Integration test is still cooking, iptables rules must be translated into nftables as should parsing for the teardown code when nftables in use

### DIFF
--- a/althea_kernel_interface/src/exit_server_tunnel.rs
+++ b/althea_kernel_interface/src/exit_server_tunnel.rs
@@ -1,9 +1,9 @@
 use super::KernelInterfaceError;
-use crate::ip_addr::add_ipv4;
+use crate::ip_addr::{add_ipv4, add_ipv4_mask, delete_ipv4};
 use crate::iptables::{add_iptables_rule, delete_iptables_matching_rules};
 use crate::netfilter::{
-    add_preroute_chain_ipv4, does_nftables_exist, init_filter_chain, init_nat_chain,
-    insert_nft_exit_forward_rules,
+    delete_forward_rule, delete_postrouting_rule, does_nftables_exist, init_filter_chain,
+    init_nat_chain, insert_nft_exit_forward_rules,
 };
 use crate::open_tunnel::to_wg_local;
 use crate::run_command;
@@ -180,9 +180,6 @@ pub fn setup_nat(
 ) -> Result<(), Error> {
     // nat masquerade on exit
     if !does_nftables_exist() {
-        error!("nftables does not exist, setting up iptables masquerade nat rules on exit");
-        // TODO: can we assume any exit upgraded to use exit routing modes will have nftables?
-        // otherwise I must duplicate the entire nat setup in iptables for all exit modes
         add_iptables_rule(
             "iptables",
             &[
@@ -203,7 +200,6 @@ pub fn setup_nat(
 
     // Add v4 and v6 forward rules wg_exit <-> ex_nic
     if !does_nftables_exist() {
-        error!("Setting up iptables nat forward rules on exit");
         // v4 wg_exit -> ex_nic
         add_iptables_rule(
             "iptables",
@@ -277,7 +273,6 @@ pub fn setup_nat(
             )?;
         }
     } else {
-        error!("Setting up nftables nat forward rules on exit");
         insert_nft_exit_forward_rules(interface, external_interface, external_v6)?;
     }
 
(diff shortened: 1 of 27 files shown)


rita_exit/src/database/mod.rs
```diff
@@ -517,3 +526,88 @@ pub fn enforce_exit_clients(client_data: &mut RitaExitData) -> Result<(), Box<Ri
...
+        .filter(|(_k, v)| v.contains(&client))
+        .map(|(&k, _v)| k)
+        .collect();
+    let internal_ip: Vec<Ipv4Addr> = int_assignments
+        .iter()
+        .filter(|(_k, v)| **v == client)
+        .map(|(&k, _v)| k)
+        .collect();
+    match (external_ip.len().cmp(&1), internal_ip.len().cmp(&1)) {
+        (std::cmp::Ordering::Greater, _) | (_, std::cmp::Ordering::Greater) => {
```
jkilpatr, line 589: this function is a little dense with all the functional operations, maybe break it out into a few little utility functions so that it's easier to see how exactly it works. It's not clear to me what happens in the error cases.

Also if the teardown fails we shouldn't release the ip back into the pool to be used, potentially resulting in a not working client connecting, better to properly return an error and keep a hold of the ip until the teardown succeeds.

settings/src/exit.rs
```diff
@@ -245,7 +252,7 @@ fn enable_enforcement_default() -> bool {
 }
 
 fn default_ipv4_routing() -> ExitIpv4RoutingSettings {
-    ExitIpv4RoutingSettings::NAT
+    ExitIpv4RoutingSettings::MASQUERADENAT
```
jkilpatr, line 255: good idea to do the rename here, more clear.

rita_exit/src/database/mod.rs
```diff
@@ -517,3 +526,88 @@ pub fn enforce_exit_clients(client_data: &mut RitaExitData) -> Result<(), Box<Ri
...
+    let int_assignments = client_data.get_internal_ip_assignments();
+    // for each client in inactive, find their wg key from clientandipmap, then
+    // remove their iptables rules if their last seen was over threshold
+    for (client, last_seen) in inactive_list.iter() {
+        if last_seen.elapsed().as_secs() > INACTIVE_CLIENT_THRESHOLD {
+            try_teardown_client_snat(
+                ext_assignments.clone(),
+                int_assignments.clone(),
+                *client,
+                ext_nic,
```
jkilpatr, line 550: why not merge the lists and have one teardown loop, I'm also not seeing how exactly the client's ip's get removed from the active list so that once torn down they can be used again but maybe I'm not reading it right.

rita_exit/src/database/mod.rs
```diff
@@ -517,3 +526,88 @@ pub fn enforce_exit_clients(client_data: &mut RitaExitData) -> Result<(), Box<Ri
     client_data.set_debt_actions(new_debt_actions);
     Ok(())
 }
+
+/// Removes client ip assignments for clients that have not checked in for 1 hour
+pub fn teardown_inactive_clients(client_data: &mut RitaExitData) {
+    let inactive_list = client_data.get_inactive_list();
+    // check for snat mode
+    let ipv4_mode = client_data.get_ipv4_nat_mode();
+    if let ExitIpv4RoutingSettings::SNAT { .. } = ipv4_mode {
```
jkilpatr, line 535: you can check for the mode before getting the inactive list and save some commands/operations

rita_exit/src/database/ipddr_assignment.rs
```diff
@@ -539,7 +592,7 @@ mod tests {
         // this way we ensure we always exhaust our test subnet, so that by the end of the test we can
         // check that we can't assign any more ips
         let num_clients =
-            get_ipv4_external_test_subnet().size() - static_assignments.len() as u32 - 2;
+            get_ipv4_external_test_subnet().size() - static_assignments.len() as u32 - 4;
```
jkilpatr, line 595: in at least one of these tests add a check where you iterate through all the assigned ip's and make sure none are equal to the gateway, external, or broadcast.

jkilpatr (changes requested):
Awesome work getting this working. It's an in depth change with a lot of moving parts. 

I've left some comments on logic cleanup in a few places, and one edgecase that needs to be handled. Overall I like the structure you've used and the implementation is very clean. 

My only flag is the significant use of functional chains to get a given value out of an array for example, these can be hard to read and confusing sometimes. You can usually just ask copilot to generate a utility function or an iterative version of the same check and get a great result.

althea_kernel_interface/src/exit_server_tunnel.rs
```diff
@@ -275,6 +279,131 @@ pub fn setup_nat(
...
+        /*
+        # SNAT for outgoing traffic
+        iptables -A POSTROUTING -t nat -o $EXT_IF -s $INTERNAL_CLIENT -j SNAT --to-source $PUBLIC_CLIENT
+        # Allow forwarded traffic
+        iptables -A FORWARD -s $INTERNAL_CLIENT -j ACCEPT
+        */
+        info!("Setting up iptables SNAT rules on exit for client: {client_external_ip} to {client_internal_ip}");
+        add_iptables_rule(
+            "iptables",
+            &[
```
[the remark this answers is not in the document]
ch-iara: removed iptables rules since if we won't be using them I don't think we need to include them untested

rita_exit/src/database/mod.rs
ch-iara: you're right, this was never being handled. added in removal from the external_ip_assignments list to the teardown function.

ch-iara:
Multi Exit test currently spitting about a bunch of "Too many open files" error from all namespaces and panics when trying to spawn the thread registering to exit?

PR #1104, opened 2025-02-13
CGNAT Exit and integration test

ch-iara:
right now this is manually adding all possible external ips on the exit subnet, which we're hoping to avoid since that gets clunky real fast.


jkilpatr (approved):
This looks good. I only have some very minor comment nits, overall using the build in cgnat is a big improvement in code simplicity!

jkilpatr:
looks like the multi exit test needs a fix.

PR #1110, opened 2025-03-01
Exit Dashboard update for exit mode

ch-iara:
Users can now change their exit mode from the dashboard, including configuring network ips and static assignments.


ch-iara:
Goes with https://github.com/althea-net/rita-exit-dash/pull/1

jkilpatr (approved):

jkilpatr (approved):

PR #1111, opened 2025-03-04
Fix multi exit test, refactor registration state data flow

jkilpatr:
This patch fixes the multi-exit test which has been broken since the patch "Simplify internal and ipv6 assignment". This is becuase that pr stopped using the fixed internal ip assignment code, where the same client would be deterministically assigned the same ip across multiple exit instances.

This sort of id hash based assignment brought in more complexity than it was worth. But removing it revealed that the exit details struct shouldn't be left as a single item. After all our exit details including the default route and our own ip can change between exits.

Including in this patch is an extensive re-write of the dataflow for the exit registration state, this is to move that state out of the config and avoid using a lazy-static to share it across threads. This modified state is more contained, and cleaner, although I would hesitate to call it clean since we sitll have some duplicated data in the exit manager state.

Moving the exit state out of settings (and in fact any thing that may change frequently) is very important becuase the settings lazy static has a copy -> modify -> write structure without a lock, data inconsistency is possible, and unavoidable for anything that changes quickly like exit states now will.

### DIFF
--- a/integration_tests/src/utils.rs
+++ b/integration_tests/src/utils.rs
@@ -336,17 +336,19 @@ pub fn test_all_internet_connectivity(namespaces: NamespaceInfo) {
                 ],
             )
             .unwrap();
-            if String::from_utf8(out.stdout)
-                .unwrap()
-                .contains("1 received")
-            {
+            let output_string = from_utf8(&out.stdout).unwrap();
+            if output_string.contains("1 received") {
                 info!("Ping test passed for {}!", ns.get_name());
                 break;
             } else {
                 if Instant::now() - start > Duration::from_secs(60) {
                     panic!("{} does not have internet connectivity", ns.get_name());
                 }
-                error!("Ping failed for {}, trying again", ns.get_name());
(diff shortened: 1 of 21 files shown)


integration_tests/src/utils.rs
```diff
@@ -336,17 +336,19 @@ pub fn test_all_internet_connectivity(namespaces: NamespaceInfo) {
...
-                .unwrap()
-                .contains("1 received")
-            {
+            let output_string = from_utf8(&out.stdout).unwrap();
+            if output_string.contains("1 received") {
                 info!("Ping test passed for {}!", ns.get_name());
                 break;
             } else {
-                if Instant::now() - start > Duration::from_secs(60) {
+                if Instant::now() - start > Duration::from_secs(600) {
```
ch-iara, line 344: 10min wait intentionally left in? Usually the 1min was fine for github actions suite

jkilpatr: yup you're right here. Will fix

